PR #5, opened 2019-08-07
ci: adding commitlint checking with travis

blacha:
### Change Description:

Adding basic travis ci support


IReese:
I need to review what Travis is before I can make any comment

PR #10, opened 2019-08-19
feat: adding aws cdk for deployment management

blacha:
### Change Description:
Adding scripts to automatically deploy to test environments

- [x] README updated (where applicable)
- [x] CHANGELOG (Unreleased section) updated
- [x] Docstrings / comments included to help explain code

- [x] Confluence user guide updated (where applicable)

#### Testing Tasks:
- [x] Added tests that fail without this change
- [x] All tests are passing in development environment
- [x] Reviewers assigned
- [x] Linked to main issue for ZenHub board


packages/_infra/src/edge/index.ts
```diff
@@ -0,0 +1,50 @@
...
+                    isDefaultBehavior: true,
+                    lambdaFunctionAssociations: [lambdaViewerRequest],
+                },
+            ],
+        };
+
+        const cloudFront = new cf.CloudFrontWebDistribution(this, 'Distribution', {
+            priceClass: cf.PriceClass.PRICE_CLASS_ALL, // Allow Sydney as a origin
+            httpVersion: cf.HttpVersion.HTTP2,
+            viewerProtocolPolicy: cf.ViewerProtocolPolicy.REDIRECT_TO_HTTPS, // Do not allow http connections
```
palmerj: I'm wondering if we still want to allow HTTP, in the context of WMTS. Some GIS client still have issue with HTTPS...

blacha: Do you know which GIS clients they are? 

There are a lot of configuration options for TLS with Cloudfront, I generally use the TLS2018 but they will cause issues with older devices. So having a list would be super helpful

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/secure-connections-supported-viewer-protocols-ciphers.html

palmerj: Generally there are low usage client such as Mapinfo and others. In the past even QGIS has been an issue. Mostly due to people being still on older versions of software and not upgrading (either due to cost or corporate reasons).

Root causes have been client library lack of support for newer TLS standards or even that CA root certs were not correctly installed in desktop environments.

blacha: Just did a quick survey, both google and OSM support tile servers not using `https` so I will take this out.

Openstreetmaps is interesting though, while it will respond over http, it has HSTS set so your browser will force you into https if you try and access their tiles

palmerj: Great.

PR #16, opened 2019-09-13
feat: lambda xyz tile server

blacha:
### Change Description:
Adds XYZ Tile server

- [x] README updated (where applicable)
- [x] CHANGELOG (Unreleased section) updated
- [x] Docstrings / comments included to help explain code

#### Testing Tasks:
- [x] All tests are passing in development environment
- [x] Reviewers assigned
- [x] Linked to main issue for ZenHub board

### DIFF
--- a/packages/lambda-xyz/src/path.ts
+++ b/packages/lambda-xyz/src/path.ts
@@ -1,0 +1,20 @@
+import * as Path from 'path-to-regexp';
+
+const XyzPath = Path('/:group/:z/:x/:y\\.:ext');
+
+export interface PathData {
+    group: string;
+    x: number;
+    y: number;
+    z: number;
+    ext: string;
+}
+export function getXyzFromPath(path: string): null | PathData {
+    const output = XyzPath.exec(path);
+    if (output == null) {
+        return null;
+    }
+
+    const [_, group, z, x, y, ext] = output;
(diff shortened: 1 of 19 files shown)


packages/lambda-xyz/src/path.ts
```diff
@@ -0,0 +1,20 @@
+import * as Path from 'path-to-regexp';
+
+const XyzPath = Path('/:group/:z/:x/:y\\.:ext');
```
palmerj: Sorry for my late drive by comment.. but have we got a design of public target API? Would be good to confirm what group is? Also it's possible we will support multiple projections in the future and maybe the path should design for this? I only raise this as the public API is very important and you don't want to be changing it too much. 

For some sort of example, LDS uses this schema for tile services: https://help.koordinates.com/query-api-and-web-services/using-koordinates-xyz-and-wmts-tiles-services-javascript-clients/. Not saying this is what we should use, but a good reference

blacha: I am happy to change the URL,  It is not designed to be public, was just trying to get something going.

In the future we will likely need a large number of options like projection & scale so using something like the coordinates system would be a good idea.

Once we learn more about our requirements we can change the url to be better suited, this is an internal testing api :)

palmerj: Just wanted to raise it now. Cool if it's internal only

PR #197, opened 2019-12-10
Feat/allow 3857

blacha:
### Change Description:

Adds support for 3857 projections, while improving the cog generation process

### Notes for Testing:

...

#### Testing Tasks:
- [ ] Reviewers assigned


packages/cog/src/gdal.ts
```diff
@@ -55,8 +58,9 @@ export class GdalCogBuilder {
         if (this.config.bbox == null) {
             return [];
         }
-        const bbox = this.config.bbox.map(s => String(s));
-        return ['-projwin', bbox[0], bbox[1], bbox[2], bbox[3], '-projwin_srs', 'EPSG:900913'];
+
+        const [ulX, ulY, lrX, llY] = this.config.bbox;
+        return ['-projwin', ulX, ulY, lrX * TiffBuffer, llY * TiffBuffer, '-projwin_srs', 'EPSG:900913'].map(String);
```
palmerj: Any reason to use EPSG:90013? I think it would be better to use the official EPSG:3857

blacha: None other than its easier for me to remember, couldn't think of the right number off the top of my head.

blacha: I have created a enum of common epsg codes so I don't have to remember the numbers :D

PR #237, opened 2020-01-14
feat: better cogify command

blacha:
this introduces a concept of a job and cog tasks

Given a colleciton of tiff files, the cli will generate a list of
quadkeys that need to be generated.

the CLI can then use the job config to generate the individual tiffs


packages/cog/src/cli/actions/action.job.ts
```diff
@@ -0,0 +1,148 @@
...
+            parameterLongName: '--min-zoom',
+            description: 'min zoom level to use',
+            defaultValue: this.MinZoomDefault,
+            required: false,
+        });
+
+        this.maxCogs = this.defineIntegerParameter({
+            argumentName: 'MAX_TILES',
+            parameterLongName: '--max-tiles',
+            description: 'Maximum number of COGS to create',
```
palmerj: Typo

palmerj:
@blacha I'm interested for the large urban auckland dataset how many COGs are generated?

blacha:
> @blacha I'm interested for the large urban auckland dataset how many COGs are generated?

It depends, on what settings you use to create the processing job.

In general we have been seeing between 3-20 COGs created per imagery set.

palmerj:
> In general we have been seeing between 3-20 COGs created per imagery set.
I guess it does't matter too much, esp once caching kicks in, but high zoom requests that need to mosaic the COG requests together will be much slower with large numbers of COGs. Zoomed in will be ok.

blacha:
> > In general we have been seeing between 3-20 COGs created per imagery set.
> I guess it does't matter too much, esp once caching kicks in, but high zoom requests that need to mosaic the COG requests together will be much slower with large numbers of COGs. Zoomed in will be ok.

The COGS are aligned to WebMercator quadkeys which means at high zoom levels you will only be accessing one COG at a time to fill the tile request.

They have been designed to aligned to the approximate zoom level they will be used at, for urban they turn on at zoom level 14, so the COGS need to be at smallest the size of a z14 tile. 

I believe the smallest we have currently made is a z13 tile for gisborne urban.

We also have in the logging the number of COGs used to generate a tile aswell as the number of internal tiles of the COG that was used to make the tile.

palmerj:
> The COGS are aligned to WebMercator quadkeys which means at high zoom levels you will only be accessing one COG at a time to fill the tile request.

Ah of course. Silly me. No problem then.

> We also have in the logging the number of COGs used to generate a tile aswell as the number of internal tiles of the COG that was used to make the tile.

This is good news. 

I only ask these questions cause I think the future data lake should use this service for previewing datasets.

blacha:
We also oversize the internal COG tiles, so they have been either generated at 1024 or 512 tile size, 

This means we will generally do less requests for when we need two 256x256 tiles out

> > The COGS are aligned to WebMercator quadkeys which means at high zoom levels you will only be accessing one COG at a time to fill the tile request.
> 
> Ah of course. Silly me. No problem then.
> 
> > We also have in the logging the number of COGs used to generate a tile aswell as the number of internal tiles of the COG that was used to make the tile.
> 
> This is good news.
> 
> I only ask these questions cause I think the future data lake should use this service for previewing datasets, or even generating thumbnails

Here is a example COG covering, this is a rural imagery set for wellington, in the basemap it turns on at zoom level 13.

The smallest COG we are generating (Blue outlines) is a z12

2020-01-14_wellington_cogs

The COGs on the east and south side will also be restricted to the size of the source imagery as it does not need to generate the full empty space as long as the top left corner is aligned to a QuadKey everything just works.

PR #257, opened 2020-01-28
Refactor/imagery locations

blacha:
Add more imagery sets


packages/lambda-xyz/src/imagery/gebco.ts
```diff
@@ -0,0 +1,9 @@
+import { MosaicCog } from '../tiff';
+
+export const MosaicGeboBathy = new MosaicCog(
```
josephramsay: MosaicGeb**c**oBathy

PR #292, opened 2020-02-18
build: allow publishing to npm of @basemaps/tiler & @basemaps/metrics

blacha:


packages/tiler-sharp/package.json
```diff
@@ -20,5 +20,11 @@
     "@types/sharp": "^0.22.3",
     "pixelmatch": "^5.1.0",
     "pngjs": "^3.4.0"
-  }
+  },
+  "publishConfig": {
```
blacha: missing private: false

blacha: and readme

packages/tiler/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@basemaps/tiler",
```
blacha: missing publish config

PR #359, opened 2020-03-17
refactor: add lambda-shared/src/router

jacott:
Router allows url handlers to register themselves. This work is the
precusor before chaning the url from which tiles are served.

#### Testing Tasks:
- [ ] Reviewers assigned


packages/lambda-api-tracker/src/__test__/xyz-request.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                            clientIp: '1.2.3.4',
+                        },
+                    },
+                },
+            ],
+        } as any;
+    }
+
+    o.beforeEach(() => {
+        mockLogger = LogConfig.get();
```
blacha: you can disable the logger with `LogConfig.disable()`

by overwriting with the spy you are destroying the logger forever.

packages/lambda-api-tracker/src/__test__/xyz-request.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        const res = await handleRequest(req('/v1/foo'), session, mockLogger);
+
+        o(res.status).equals(404);
+    });
+
+    o('should catch missing api key', async () => {
+        const session = new LambdaSession();
+        const res = await handleRequest(req('/v1/tiles/1/2/3.png', ''), session, mockLogger);
+
+        o(res.status).equals(400);
```
blacha: do these give more information we can test for eg `res.body`? to validate that it is actually failing due to apikey and not something else?

packages/lambda-api-tracker/src/__test__/xyz-request.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    o('should return LambdaHttpResponseCloudFrontRequest on success', async () => {
+        ValidateRequest.validate = async (): Promise<LambdaHttpResponseCloudFront | null> => {
```
blacha: nit: dont need the brackets here can just `=> null`

packages/lambda-api-tracker/src/tile-request.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const projection = new Projection(256);
+
+/**
+ * Validate a CloudFront request has a valid API key and is not abusing the system
+ */
+export default async (info: ReqInfo): Promise<LambdaHttpResponse> => {
+    // Expose some debugging metrics so we can plot tile requests onto a map
+    const xyzData = tileFromPath(info.rest);
+    if (xyzData == null) return new LambdaHttpResponseCloudFront(404, 'Not Found');
```
blacha: Other URLS need to be supported get `/WMTSCapabilites.xml` if you 404 here they wont be able to be accessed.

you need to return the `request` to allow the request to go forward

packages/lambda-shared/src/api-path.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export interface TileData {
+    x: number;
+    y: number;
+    z: number;
+    ext: string;
+}
+
+export const extractAction = (info: ActionData): void => {
```
blacha: this function should not modify the input object, we should keep the input object clean and return the updated data.

being named `extract` implies that it would extract data not modify it.

packages/lambda-xyz/src/__test__/health.test.ts
```diff
@@ -0,0 +1,14 @@
+import { LambdaHttpResponseAlb } from '@basemaps/lambda-shared';
+import * as o from 'ospec';
+import health from '../health';
+
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
blacha: Thoughts on just disabling this eslint rule, seems to be ignored more than used.

packages/lambda-xyz/src/__test__/index.test.ts
```diff
@@ -0,0 +1,34 @@
+// process.env['COG_BUCKET'] = 'fake-bucket';
```
blacha: remove

packages/lambda-xyz/src/__test__/ping.test.ts
```diff
@@ -0,0 +1,14 @@
+import { LambdaHttpResponseAlb } from '@basemaps/lambda-shared';
+import * as o from 'ospec';
+import ping from '../ping';
+
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+o.spec('ping', () => {
```
blacha: this is tested in the file above?

packages/lambda-xyz/src/__test__/version.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+o.spec('version', () => {
+    const origVersion: any = process.env[Env.Version];
+    const origHash: any = process.env[Env.Hash];
+    o.after(() => {
+        process.env[Env.Version] = origVersion;
+        process.env[Env.Hash] = origHash;
+    });
+    process.env[Env.Version] = '1.2.3';
```
blacha: this is duplicated logic with lines 17/18

packages/lambda-api-tracker/src/__test__/xyz-request.test.ts
jacott: Just the `res.statusDescription`; now added to test for statusDescription. This is just refactored code from index.ts; I can add a body if you like or add a ticket to do so?

packages/lambda-api-tracker/src/tile-request.ts
jacott: Yep will do that as part of the WMTS ticket

packages/lambda-shared/src/api-path.ts
jacott: Are you say in general you rather copy instead of update or are you just pointing out the function name/action mismatch?

packages/lambda-xyz/src/__test__/health.test.ts
jacott: this is everywhere in the tests. Should I remove it from all tests or just the ones I've touched?

packages/lambda-xyz/src/__test__/ping.test.ts
jacott: yep I wanted your opinion if you liked tests per file; seems you don't (like me) so I merge these tests

packages/lambda-xyz/src/__test__/health.test.ts
jacott: Okay, I see some tests need it

packages/lambda-api-tracker/src/tile-request.ts
blacha: Ok lets review this after the WMTS work is done.

PR #364, opened 2020-03-18
fix: remove resample arg from batch

josephramsay:
Fixes: #  
resample arg left behind in batch command incompatible with cog job args after move to job.json


blacha:
@josephramsay do you want to rebase out the merge commit then we can merge this in.

PR #374, opened 2020-03-22
fix: update landing page and cli/serve to include aerial/3857

jacott:
### Fixes:

1. landing page tiles url
2. `cli/serve.ts` to forward correct url

#### Testing Tasks:
- [ ] Reviewers assigned

### DIFF
--- a/packages/lambda-xyz/src/cli/serve.ts
+++ b/packages/lambda-xyz/src/cli/serve.ts
@@ -50,7 +50,7 @@ async function main(): Promise<void> {
             const data = await lambda.handleRequest(
                 {
                     httpMethod: 'get',
-                    path: `/v1/tiles/${z}/${x}/${y}.png`,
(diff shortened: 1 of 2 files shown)


packages/lambda-xyz/src/cli/serve.ts
```diff
@@ -50,7 +50,7 @@ async function main(): Promise<void> {
             const data = await lambda.handleRequest(
                 {
                     httpMethod: 'get',
-                    path: `/v1/tiles/${z}/${x}/${y}.png`,
+                    path: `/v1/tiles/aerial/3857/${z}/${x}/${y}.png`,
```
palmerj: Would it not make sense to detect if webp is available in the browser (just about all now I think), then use that? It will really reduce the bandwidth.

jacott: This is just a test server

jacott: But yes for the html landing page we plans to serve webp

PR #384, opened 2020-03-24
feat(tile): serve png, webp and jpeg

jacott:
### Change Description:

Allow requests for jpeg and webp image formats in addition to PNG

### Notes for Testing:

Check that server responds with image of requested type

#### Testing Tasks:
- [ ] Reviewers assigned


packages/lambda-shared/src/__test__/api.path.test.ts
```diff
@@ -2,6 +2,7 @@ import { EPSG } from '@basemaps/geo';
 import * as o from 'ospec';
 import { tileFromPath, TileSetType, TileType } from '../api.path';
 import { LambdaContext } from '../lambda.context';
+import { ImageFormat } from '@basemaps/tiler';
```
blacha: doesnt look like lambda-shared has a dep on `@basemaps/tiler` I wonder if this is going to cause issues down the line.

PR #406, opened 2020-03-30
Feat/wmts improvements

jacott:
### Change Description:

1. Add the layer abstract, this should include information about the imagery set or at least a URL to more detailed information
2. The TileMatrixSet would be better labeled as GoogleMapsCompatible to be consistent with other services, including the LDS. “aerial” is not a good label.
3. Add <WellKnownScaleSet>urn:ogc:def:wkss:OGC:1.0:GoogleMapsCompatible</WellKnownScaleSet> to the tilematrixset definition. Will this add better interoperability with some clients. 
4. You might need to consider adding “crs=“urn:ogc:def:crs:OGC:2:84” to the layer ows:WGS84BoundingBox element to explicitly define the WGS84 coordinate axis order. I remember  ArcMap 10.3/4 needed it and ArcGIS (as least when I last tested). It might be fixed now, but worth another look. I don’t see an harm in adding the attribute in any case for further axis order compatability. 
5. Increase zoom levels to 25

### Notes for Testing:

...

#### Testing Tasks:
- [ ] Reviewers assigned

### DIFF
--- a/packages/geo/src/projection.ts
+++ b/packages/geo/src/projection.ts
@@ -114,6 +114,11 @@ export class Projection {
         return `EPSG:${epsg}`;
     }
 
+    /** convert a EPSG code into a URN */
+    public static toUrn(epsg: EPSG): string {
+        return `urn:ogc:def:crs:EPSG:6.18:3:${epsg}`;
+    }
+
     /** parse a string returning the `EPSG` code **/
     public static parseEpsgString(text: string): EPSG | null {
         return EPSGTextMap[text.replace(/[\W_]/g, '').toLowerCase()] || null;
--- a/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
+++ b/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
@@ -26,7 +26,7 @@ o.spec('wmts', () => {
         ]);
 
         o(listTag(raw, 'ows:WGS84BoundingBox')).deepEquals([
-            '<ows:WGS84BoundingBox>\n' +
+            '<ows:WGS84BoundingBox crs="urn:ogc:def:crs:OGC:2:84">\n' +
                 '  <ows:LowerCorner>-180 -85.0511287798066</ows:LowerCorner>\n' +
                 '  <ows:UpperCorner>180 85.0511287798066</ows:UpperCorner>\n' +
                 '</ows:WGS84BoundingBox>',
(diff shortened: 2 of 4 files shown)


packages/geo/src/projection.ts
```diff
@@ -114,6 +114,11 @@ export class Projection {
         return `EPSG:${epsg}`;
     }
 
+    /** convert a EPSG code into a URN */
+    public static toURN(epsg: EPSG): string {
```
blacha: `toUrn`

packages/lambda-xyz/src/__test__/wmts.capability.test.ts
```diff
@@ -44,11 +44,17 @@ o.spec('wmts', () => {
         o(tileMatrixSet.length).equals(2);
 
         o(listTag(tileMatrixSet[1], 'ows:Identifier')[0]).equals('<ows:Identifier>aerial</ows:Identifier>');
-        o(listTag(tileMatrixSet[1], 'ows:SupportedCRS')).deepEquals(['<ows:SupportedCRS>EPSG:3857</ows:SupportedCRS>']);
+        o(listTag(tileMatrixSet[1], 'ows:SupportedCRS')).deepEquals([
+            '<ows:SupportedCRS>urn:ogc:def:crs:EPSG:6.18:3:3857</ows:SupportedCRS>',
```
blacha: `toUrn`?

packages/lambda-xyz/src/wmts.capability.ts
```diff
@@ -86,7 +114,7 @@ const MatrixSets = new Map<TileSetType, EPSGToGenerator>();
             const matrices = [];
             let scale = 559082264.029,
                 size = 1;
-            for (let i = 0; i < 20; ++i, scale *= 0.5) {
+            for (let i = 0; i < 25; ++i, scale *= 0.5) {
```
blacha: could you extract `25` out into a const? `const MaxZoomLevel = 25`

packages/lambda-xyz/src/wmts.capability.ts
```diff
@@ -1,9 +1,11 @@
-import { V, VNode, TileSetType } from '@basemaps/lambda-shared';
+import { V, VNode, TileSetType, VNodeElement } from '@basemaps/lambda-shared';
 import { EPSG, Projection } from '@basemaps/geo';
 import { ImageFormat } from '@basemaps/tiler';
 
 const { lat, lon } = Projection.Wgs84Bound;
 
+const MaxZoomLevel = 25;
```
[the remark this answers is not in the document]
JBLINZ: @palmerj  We have hear from several users that they do not want be restricted in terms of over-zooming. Often users are displaying much more accurate/detailed data over the top of our basemap, and dont want the contextual raster data to disappear just because its visual appearance is degraded. Is there another way to be transparent about the original resolution on the imagery?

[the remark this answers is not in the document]
palmerj: Hi @JBLINZ, the over-zooming is still possible in the GIS clients (QGIS, ArcGIS), it will just resample the last fetched zoom level of tiles. This change will just stop the fetching of new tiles. So users will be ok. I still think advertising these tile extra zoom levels is not recommended given the resolution. In the future should do this for vector tiles either, but again the client would over-zoom the vector if they want. Note: for web clients users will not be using WMTS endpoint, they will use the XYZ URL manually configuring the rest.

PR #415, opened 2020-03-31
feat: make fetchImagery work with > 100 keys

jacott:
### Change Description:

make fetchImagery work with > 100 keys

If not all keys are fetched list missing ids in error message

#### Testing Tasks:
- [ ] Reviewers assigned


packages/lambda-shared/src/aws/tile.metadata.table.ts
```diff
@@ -111,29 +120,30 @@ export class TileMetadataTable {
...
+                    RequestItems: {
+                        [Const.TileMetadata.TableName]: { Keys },
+                    },
+                })
+                .promise();
+
+            const metadataItems = items.Responses?.[Const.TileMetadata.TableName];
+            if (metadataItems == null) throw new Error('Failed to fetch tile metadata');
+
+            const res1 = metadataItems.map((c) => DynamoDB.Converter.unmarshall(c)) as TileMetadataImageryRecord[];
```
blacha: we are now looping over this about three times, maybe we should return a map of imagery

then we can just 
```
for (const item of metadataItems) {
   const i = Convert(item);
   output.set(i.id, i)
   this.imagery.set(i.id, i)
}
```

packages/lambda-shared/src/aws/__test__/tile.metadata.table.test.ts
```diff
@@ -62,4 +70,76 @@ o.spec('tile.metadata.table', () => {
             }),
         });
     });
+
+    o('should report missing imagery', async () => {
+        const tmtable = new TileMetadataTable();
+        (tmtable as any).fetchImagery = async (): Promise<any[]> => {
```
blacha, line 76: I think you can drop this any casts, if you make the Promise<any[]> actually the same as the `fetchImagery` return.

packages/lambda-shared/src/aws/tile.metadata.table.ts
```diff
@@ -111,29 +120,30 @@ export class TileMetadataTable {
...
-        // TODO diff the arrays see what is missing?
-        if (metadataItems.length < Keys.length) throw new Error('Missing fetched items');
-
-        const results = metadataItems.map((c) => DynamoDB.Converter.unmarshall(c)) as TileMetadataImageryRecord[];
-
-        for (const result of results) this.imagery.set(result.id, result);
+        let mappedKeys = keys.map(toId);
+        let results: TileMetadataImageryRecord[] | null = null;
+
+        while (mappedKeys.length != 0) {
```
blacha, line 126: call it paranoia, but I never like direct equals comparisions on while loops, no idea how .length could get below zero but if it does it will go forever!

packages/lambda-shared/src/aws/__test__/tile.metadata.table.test.ts
jacott: I need to cast to any because `fetchImagery` is private

packages/lambda-shared/src/aws/tile.metadata.table.ts
jacott: I call it paranoia ;)

packages/lambda-shared/src/aws/tile.metadata.table.ts
```diff
@@ -99,8 +99,17 @@ export class TileMetadataTable {
         }
 
         if (imageList.size > 0) {
-            const items = await this.fetchImagery(Array.from(imageList.values()));
+            const keys = Array.from(imageList.values());
+            const items = await this.fetchImagery(keys);
             for (const item of items) output.set(item.id, item);
+
+            if (items.length < keys.length) {
+                let ids = '';
```
blacha, line 107: could simplify to something like 

`const ids = keys.filter(f => !output.has(f)).join(', ')`

packages/lambda-shared/src/aws/tile.metadata.table.ts
jacott: I'll do some more work on this

packages/lambda-shared/src/aws/tile.metadata.table.ts
```diff
@@ -99,41 +99,48 @@ export class TileMetadataTable {
         }
 
         if (imageList.size > 0) {
-            const items = await this.fetchImagery(Array.from(imageList.values()));
-            for (const item of items) output.set(item.id, item);
+            const keys = Array.from(imageList.values());
+            await this.fetchImagery(keys, output);
```
jacott, line 103: I've go a more efficient solution to this.

PR #428, opened 2020-04-05
feat: store metadata imagery on cog creation

jacott:
#### Changes

When `cogify batch` is started store job imagery information in the metadata db

#### Testing Tasks:
- [ ] Reviewers assigned

### DIFF
--- a/packages/cog/src/cli/cogify/__test__/action.batch.test.ts
+++ b/packages/cog/src/cli/cogify/__test__/action.batch.test.ts
@@ -1,0 +1,56 @@
+import { TileMetadataTable } from '@basemaps/lambda-shared';
+import * as o from 'ospec';
+import { CogJob } from '../../../cog/cog';
+import { extractYearFromName, storeImage } from '../action.batch';
+
+o.spec('action.cog', () => {
+    o('extractYearFromName', () => {
+        o(extractYearFromName('2013')).equals(2013);
+        o(extractYearFromName('abc2017def')).equals(2017);
+        o(extractYearFromName('2019_abc')).equals(2019);
+        o(extractYearFromName('12019_abc')).equals(-1);
+        o(extractYearFromName('2019_abc2020')).equals(2020);
+        o(extractYearFromName('2020_abc2019')).equals(2020);
+    });
+
+    o('storeImage', () => {
+        const origNow = Date.now;
+        const origCreate = TileMetadataTable.prototype.create;
+
+        const job: CogJob = {
+            id: 'abc123',
+            name: '2019-new-zealand-sentinel',
+            projection: 3857,
+            source: {
+                resolution: 13,
+            },
+            output: {},
+            quadkeys: ['311333222331', '311333223200', '311333223202', '3113332223131'],
+        } as CogJob;
+
+        const mockNow = Date.now();
+        try {
+            const create = o.spy();
+            Date.now = (): number => mockNow;
+            TileMetadataTable.prototype.create = create as any;
+
+            storeImage(job);
+
+            o(create.args).deepEquals([
+                {
+                    id: 'im_abc123',
+                    name: '2019-new-zealand-sentinel',
+                    createdAt: mockNow,
+                    updatedAt: mockNow,
+                    projection: 3857,
+                    year: 2019,
+                    resolution: 13,
+                    quadKeys: ['311333222331', '311333223200', '311333223202', '3113332223131'],
+                },
+            ]);
+        } finally {
+            TileMetadataTable.prototype.create = origCreate;
+            Date.now = origNow;
+        }
+    });
+});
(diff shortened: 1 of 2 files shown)


packages/cog/src/cli/actions/action.cog.ts
```diff
@@ -11,6 +11,33 @@ import { buildCogForQuadKey, CogJob } from '../../cog/cog';
...
+
+export function storeImage(job: CogJob): void {
+    const now = Date.now();
+    new TileMetadataTable().create({
+        id: job.id, // FIXME - should this be calculated by name?
+        name: job.name,
+        createdAt: now,
+        updatedAt: now,
+        projection: job.projection,
+        year: extractYearFromName(job.name),
```
jacott, line 35: Is the job.name the right place to extract the year from?

packages/cog/src/cli/actions/action.cog.ts
```diff
@@ -11,6 +11,33 @@ import { buildCogForQuadKey, CogJob } from '../../cog/cog';
...
+        if (t > year) year = t;
+    }
+
+    return year;
+}
+
+export function storeImage(job: CogJob): void {
+    const now = Date.now();
+    new TileMetadataTable().create({
+        id: job.id, // FIXME - should this be calculated by name?
```
jacott, line 30: is job.id right?

packages/cog/src/cli/actions/action.cog.ts
```diff
@@ -11,6 +11,33 @@ import { buildCogForQuadKey, CogJob } from '../../cog/cog';
 import { buildWarpedVrt } from '../../cog/cog.vrt';
 import { getJobPath, makeTempFolder } from '../folder';
 
+export function extractYearFromName(name: string): number {
+    const re = /(?:^|\D)(\d{4})(?:$|\D)/g;
+
+    let year = 2000;
+
+    for (let m = re.exec(name); m != null; m = re.exec(name)) {
+        const t = +m[1];
```
blacha, line 20: maybe be specific `parseInt(m[1])`

packages/cog/src/cli/actions/action.cog.ts
blacha: job.name is as good as it gets.

packages/cog/src/cli/actions/action.cog.ts
```diff
@@ -11,6 +11,33 @@ import { buildCogForQuadKey, CogJob } from '../../cog/cog';
...
+        if (t > year) year = t;
+    }
+
+    return year;
```
blacha, line 24: I am not sure we want to default to year `2000`

some imagery sets will not have the year inside of them.

packages/cog/src/cli/actions/action.cog.ts
blacha: This `storeImage` looks to be getting called every time we process a cog, so for some imagery sets this will be called 100+ times.

I wonder if we want to run a quadkey `addToSet` `removeFromSet` action rather than create

packages/cog/src/cli/actions/action.cog.ts
jacott: What do you suggest if year is missing? 1900?

packages/cog/src/cli/actions/action.cog.ts
jacott: Should this be done in action.batch rather than action.cog?

packages/cog/src/cli/actions/action.cog.ts
blacha: maybe leave it empty, its upto the user to configure it. or even -1?

packages/cog/src/cli/actions/action.cog.ts
blacha: We could insert a record inside of `batch` then update it inside of `cog`.

Insert a record with `quadKeys: []` and then inside of cog we could do a `addToSet: quadkey :quadKeys` (not sure the exact syntax)

packages/cog/src/cli/actions/action.cog.ts
jacott: -1 since the field is not optional

packages/cog/src/cli/actions/action.cog.ts
jacott: What is `addToSet`?  wouldn't updating in action.cog still result in 100+ writes?

blacha: 100 writes are fine if you are actually modifying something.

In this PR, you are writing the same data 100s of time.

if we were doing a `addToArray` we would be updating what "quadKeys" have been processed

Examples on how to adjust an array with dynamodb updates.

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.UpdateExpressions.html#Expressions.UpdateExpressions.SET.AddingListElements

jacott: The problem with updating the `quadKeys` attribute is that the order in which the `gdal_translate` finishes is indeterminate and I can not therefore easily update the array in proper order. What is wrong with just creating the entire record in `action.batch` prior to job submission and not performing any updates in `action.cog`?

blacha: The order doesnt actually matter, it will check each quadkey for intersections.

it was only ordered so that I could look at them nicely.

but either way is fine.

PR #482, opened 2020-04-16
feat: parse vrt files so we can modify them

blacha:
Parse VRT files into VDom so we can modify the file locations.

### DIFF
--- a/packages/lambda-shared/src/vdom.parse.ts
+++ b/packages/lambda-shared/src/vdom.parse.ts
@@ -1,0 +1,49 @@
+import * as sax from 'sax';
+import { VNodeElement, VNodeText } from './vdom';
+
+export class VNodeParser {
+    root: VNodeElement;
+    parser: sax.SAXParser;
+    tree: VNodeElement[];
+
+    static parse(xml: string): Promise<VNodeElement> {
+        return new VNodeParser().parse(xml);
+    }
+
+    constructor() {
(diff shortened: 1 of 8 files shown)


packages/lambda-shared/src/vdom.parse.ts
```diff
@@ -0,0 +1,49 @@
+import * as sax from 'sax';
+import { VNodeElement, VNodeText } from './vdom';
+
+export class VdomParser {
+    root: VNodeElement;
+    parser: sax.SAXParser;
+    tree: VNodeElement[];
+
+    static parse(xml: string): Promise<VNodeElement> {
```
jacott, line 9: suggest `parse` takes `Buffer` arg in addition to `string`

jacott: Ignore suggestion

packages/lambda-shared/src/vdom.ts
```diff
@@ -69,13 +73,13 @@ export class VNodeElement extends VNode {
...
-        let out = '';
+        const out = [];
         for (const key of keys) {
             const val = this.attrs[key];
             if (val == null) continue;
-            out += ` ${key}="${val}"`;
+            out.push(` ${key}="${val}"`);
         }
-        return out;
+        return out.join('');
```
jacott, line 82: This is much slower than string concat

PR #487, opened 2020-04-17
feat: adding more utility functions for quad keys

blacha:


packages/geo/src/quad.key.ts
```diff
@@ -9,4 +25,90 @@ export const QuadKey = {
...
+        return qk.substr(0, qk.length - 1);
+    },
+
+    /**
+     * Find duplicate quadkeys and remove them while simplifying groupings
+     *
+     * TODO is there a faster way to do this
+     *
+     * @param quadKeys
+     */
```
jacott, line 57: Faster would be to use a trie instead of a Set. It would automatically remove duplicates

blacha: Just done some benchmarking it can process 500,000 quadkeys of length 10-15 in < 100ms so it seems fast enough for my usecase.

PR #514, opened 2020-04-22
chore(cog/builder): log more info for multi-projections

jacott:

#### Testing Tasks:
- [ ] Reviewers assigned

### DIFF
--- a/packages/cog/src/cog/builder.ts
+++ b/packages/cog/src/cog/builder.ts
@@ -81,7 +81,18 @@ export class CogBuilder {
 
                 const imageProjection = this.findProjection(tiff, logger);
                 if (imageProjection != null && projection != imageProjection) {
-                    if (projection != null) throw new Error('Multiple projections');
+                    if (projection != null) {
+                        logger.error(
+                            {
+                                firstImage: sources[0].name,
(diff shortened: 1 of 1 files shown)


packages/cog/src/cog/builder.ts
```diff
@@ -81,7 +81,16 @@ export class CogBuilder {
 
                 const imageProjection = this.findProjection(tiff, logger);
                 if (imageProjection != null && projection != imageProjection) {
-                    if (projection != null) throw new Error('Multiple projections');
+                    if (projection != null) {
+                        logger.error(
+                            {
+                                first: sources[0].name + ': ' + projection,
```
blacha, line 87: leaving these as keys makes it easier to pull out in the logging

```
{
firstImage: sources[0].name
projection,
currentImage: source.name,
currentProjection: imageProjection
}
```

blacha: Also our elastic search does not like it if we mix and match types on keys, so if we use the key "current" and its a number somewhere in the code... it has to be a number everytime we use that key.

Just a example, not sure if we use the key "current" anywhere already.

PR #517, opened 2020-04-23
Adding Sentinel imagery clipped to coast

josephramsay:
Fixes: #  Ragged boundary and inconsistent colouring between sentinel, gebco and aerial imagery 

### Change Description:

Sentinel imagery clipped to coast line shape file


packages/lambda-xyz/src/cli/imagery/sentinel.ts
```diff
@@ -1,10 +1,74 @@
...
+import { MosaicCog } from './mosaic.cog';
 
 MosaicCog.create({
-    id: '01E4PN40AEH10EH128AG28YSDM',
-    name: 'new_zealand_sentinel_2018-19_10m',
-    quadKeys: ['31'],
+    id: '01E6JDD1C6RMH29Q8Z97AEE79S',
+    name: 'sentinel',
+    quadKeys: [
+        '3113332',
```
blacha, line 8: This is way too many quad keys for this imagery.

It should be at most about 5?

packages/lambda-xyz/src/cli/imagery/sentinel.ts
```diff
@@ -1,10 +1,74 @@
-import { MosaicCog } from './mosaic.cog';
 import { EPSG } from '@basemaps/geo';
+import { MosaicCog } from './mosaic.cog';
 
 MosaicCog.create({
-    id: '01E4PN40AEH10EH128AG28YSDM',
-    name: 'new_zealand_sentinel_2018-19_10m',
-    quadKeys: ['31'],
+    id: '01E6JDD1C6RMH29Q8Z97AEE79S',
+    name: 'sentinel',
```
blacha, line 6: We should be keeping the name from what it is in the datalake `new_zealand_sentinel_2018-19_10m`

palmerj: Note the new data lake structure has changed: https://confluence.linz.govt.nz/display/GEOD/File+and+Folder+Structure+in+the+Raster+Data+Store

Differences what you have now is the year ranges will include the full year string and the data type will be appended to the end (e.g RGBA)

josephramsay: Changed back to new_zealand_sentinel_2018-19_10m

packages/lambda-xyz/src/cli/imagery/sentinel.ts
josephramsay: Set max-cogs to 5, returned 3 quadkeys

josephramsay:
Reprocessed with 3 quadkeys, missed cutline. Will redo

PR #539, opened 2020-04-30
Feat/cog/optimize covering

jacott:
### Change Description:

Auto optimize the number and size of the quadkeys to cover the source imagery 

#### Testing Tasks:
- [ ] Reviewers assigned


packages/cog/package.json
```diff
@@ -27,7 +27,8 @@
     "chalk": "^4.0.0",
     "p-limit": "^2.2.1",
     "pretty-json-log": "^0.2.2",
-    "proj4": "^2.6.0"
+    "proj4": "^2.6.0",
+    "quadtree-lib": "^1.0.9"
```
jacott, line 31: Too slow

packages/cog/src/cog/covering.ts
```diff
@@ -1,43 +1,42 @@
...
+                }
+            }
+        } else {
+            // Use source for covering
+            for (const poly of sourceMetadata.bounds.features) {
+                for (const q of TileCover.cover(poly.geometry, 1, sourceZ)) {
+                    covering.add(q);
+                }
             }
         }
```
jacott, line 36: Naive  logic assumes cutline is within source imagery which is usually not true

packages/cog/src/cog/cog.ts
```diff
@@ -22,11 +22,11 @@ export function onProgress(keys: Record<string, any>, logger: LogType): (p: numb
...
  * Return the width/height of the quadkey in pixels at the target resolution
  * @param quadKey
- * @param targetResolution
+ * @param targetResolutionZoom
  * @param tileSize (Optional) size of each tile
  */
-export function getTileSize(quadKey: string, targetResolution: number, tileSize = 256): number {
-    return tileSize * Math.pow(2, targetResolution - quadKey.length + 1);
+export function getTileSize(quadKey: string, targetResolutionZoom: number, tileSize = 256): number {
+    return tileSize * Math.pow(2, targetResolutionZoom - quadKey.length + 1) * 1.5;
```
blacha, line 29: Why 1.5?

packages/cog/src/cog/cog.ts
```diff
@@ -22,11 +22,11 @@ export function onProgress(keys: Record<string, any>, logger: LogType): (p: numb
 /**
  * Return the width/height of the quadkey in pixels at the target resolution
  * @param quadKey
- * @param targetResolution
+ * @param targetResolutionZoom
```
blacha, line 25: What is a resolution zoom?

is this pixels / meters? or a `zoom`?

packages/cog/src/cog/cog.ts
```diff
@@ -51,15 +51,15 @@ export async function buildCogForQuadKey(
...
+    const [west, north] = forward(bbox.slice(0, 2));
+    const [east, south] = forward(bbox.slice(2));
 
     const [x, y, z] = QuadKey.toXYZ(quadKey);
 
     const alignmentLevels = job.source.resolution - z;
 
     const cogBuild = new GdalCogBuilder(vrtLocation, outputTiffPath, {
-        bbox: [east, south, west, north],
+        bbox: [west, south, east, north],
```
blacha, line 62: is a bbox meant to be WSEN or ENWS?

I think we both? possibly should just pick one.

packages/cog/src/cog/covering.ts
```diff
@@ -1,43 +1,25 @@
...
-     * @param maxTiles Max number of tiles to be a "valid" covering
+     * @param featureCollection Source TIff Polygons in GeoJson WGS84
+
+     * @param cutline A cutline to apply. Can be empty in which case the source Tiff bounds will be
+     * used
      */
-    optimize(sourceMetadata: SourceMetadata, cutline: JobCutline, maxZoom = 13, maxTiles = 25): QuadKeyTrie {
-        const featureCollection = sourceMetadata.bounds;
+    optimize(sourceMetadata: SourceMetadata, cutline: JobCutline): QuadKeyTrie {
+        const sourceZ = sourceMetadata.resolution;
```
blacha, line 14: I think we are not using the word resolution correctly here, this should be something like `zoom`

I think we should use 
`Resolution` for pixels/meter
`Zoom` Tile Zoom level

packages/cog/src/cog/covering.ts
```diff
@@ -1,43 +1,25 @@
...
-                    for (let i = 0; i < index; i++) {
-                        if (QuadKey.intersects(f, ary[i])) {
-                            return false;
-                        }
-                    }
-                    return true;
-                })
-                // make the output go from 0 -> 3 and 001 -> 001111
-                .sort(QuadKey.compareKeys);
+        TileCover.mergeQuadKeys(covering.cursor(), 0.6, corseZ, fineZ);
```
blacha, line 21: can we extract 0.6 to a const, and explain why we chose it.

packages/cog/src/cog/job.cutline.ts
```diff
@@ -1,85 +1,76 @@
...
 
-function addPolygonToQuadKey(self: JobCutline, quadKey: string, coords: Position[]): void {
-    let coordMap = self.quadKeyMap.get(quadKey);
-    if (coordMap == null) {
-        coordMap = new Map<Position[], number>();
-        self.quadKeyMap.set(quadKey, coordMap!);
-    }
-    coordMap!.set(coords, -1);
+function mergeCovering(self: QuadKeyTrie, other: string[] | QuadKeyTrie): void {
+    for (const qk of other) self.add(qk);
```
blacha, line 10: do we really need a function for this?

packages/geo/src/tile.cover.ts
```diff
@@ -26,4 +27,30 @@ export const TileCover = {
 
         return GeoJson.toFeatureCollection(polygons);
     },
+
+    /**
+     * Merge quadKeys if covering is > MaxPc and zoomLevel is > minZ
+     * @param c the Covering cursor - set to node to merge from
```
blacha, line 33: avoid single letter var names.

packages/geo/src/tile.cover.ts
```diff
@@ -26,4 +27,30 @@ export const TileCover = {
...
+     * @param maxPc the fraction of coverage allowd to not merge
```
blacha, line 34: `maxPc`  doesnt really explain what this is?

`coveragePercent`?

also spelling of `allowd`

packages/cog/src/cog/cog.ts
jacott: I don't know; just is

packages/cog/src/cog/cog.ts
jacott: what about just `targetZoom`?

packages/cog/src/cog/cog.ts
jacott: I don't know if we have two points (lon,lat) `[-5,-6]` and `[20, 40]` then the bbox should be `[-5,-6,20,40]`

jacott: In other words the lesser lon, the lesser lat, the greater lon, the grater lat, so `[west, south, east, north]` 

from the rfc: https://tools.ietf.org/html/rfc7946#section-5
 
   A GeoJSON object MAY have a member named "bbox" to include
   information on the coordinate range for its Geometries, Features, or
   FeatureCollections.  The value of the bbox member MUST be an array of
   length 2*n where n is the number of dimensions represented in the
   contained geometries, with all axes of the most southwesterly point
   followed by all axes of the more northeasterly point.  The axes order
   of a bbox follows the axes order of geometries.

packages/cog/src/cog/covering.ts
jacott: Yep we talked about renaming it. Shall I go ahead and call it `zoom`?

packages/cog/src/cog/job.cutline.ts
```diff
@@ -1,85 +1,76 @@
...
+        const keepLines = new Set<GeoJSON.Position[]>();
+
+        const coords: Position[][] = [[]];
+
+        const geom = GeoJson.toFeaturePolygon(coords).geometry as GeoJSON.Polygon;
+
+        const clCovering = new QuadKeyTrie();
+
+        for (const poly of this.polygons) {
+            coords[0] = poly;
```
blacha, line 63: Just make a new geojson, screwing with the internal coordinate array makes this code block difficult to understand.

readability is much better than performance gain you would get by doing this.

packages/cog/src/cog/covering.ts
blacha: Yup just rename it

packages/cog/src/cog/cog.ts
blacha: so GeoJSON has a standard of west,south,east,north. I believe the codebase has used east north west south in places.

so we should possilbly look to change to WSEN.

packages/geo/src/quad.key.trie.ts
```diff
@@ -1,80 +1,295 @@
...
+    Finished,
 }
 
-function collectForPoint(quadKeys: string[], point: number[], node: QuadKeyTrieNode, currentStr = ''): void {
-    if (QuadKey.containsPoint(currentStr, point)) {
-        if (node[QkIndexKey]) quadKeys.push(currentStr);
-        for (const key of QuadKey.Keys) {
-            const newCurrent = node[key];
-            if (newCurrent != null) collectForPoint(quadKeys, point, newCurrent, currentStr + key);
+const symMemo = Symbol('memo');
```
blacha, line 9: dont prefix vars by their types.

you also dont need symbols for this.

packages/geo/src/quad.key.trie.ts
```diff
@@ -1,80 +1,295 @@
...
+            m.index = index;
+            m.parent = parent;
+            m.quadKey = quadKey;
+        }
+    }
+}
+
+export class QuadKeyTrieCursor {
+    readonly trie: QuadKeyTrie;
+    [symMemo]: CursorMemo;
```
blacha, line 71: this doesnt ened to be a symbol

packages/cog/src/cog/constants.ts
```diff
@@ -0,0 +1,14 @@
+/** The image with of a single tile in pixels */
+export const singleTileWidth = 256 * 1.5;
```
blacha, line 2: lets make constants start with a capital letter

PR #546, opened 2020-04-30
refactor: improve guessing logic to be more correct

blacha:
look for the full NZTM name rather than the NZGD which may or may not actually be NZTM


palmerj:
LGTM

PR #553, opened 2020-05-04
refactor: rename resample to resampling to be consistent with gdal

blacha:
Remove webpack as its huge, use ncc as its small and just "works"


packages/cog/src/gdal/gdal.ts
```diff
@@ -54,12 +54,11 @@ export class GdalCogBuilder {
 
         this.config = {
             bbox: config.bbox,
-            alignmentLevels: config.alignmentLevels ?? 1,
-            compression: config.compression ?? 'webp',
-            resampling: config.resampling ?? 'bilinear',
-            blockSize: config.blockSize ?? 512,
+            alignmentLevels: config.alignmentLevels ?? GdalCogBuilderDefaults.alignmentLevels,
```
blacha, line 57: it would be nice if we could just `{...GdalCogBuilderDefaults, ..config}` but config has empty keys so it doesnt work :'(

PR #555, opened 2020-05-05
feat(cli): switch to priority numbers rather than array position

blacha:


blacha:
This should be merged after #554

packages/lambda-xyz/src/tile.set.ts
```diff
@@ -17,8 +17,8 @@ export class TileSet {
     name: string;
     tag: TileSetTag;
     projection: EPSG;
-    tileSet: TileMetadataSetRecord;
-    imagery: Map<string, TileMetadataImageryRecord>;
+    private tileSet: TileMetadataSetRecord;
```
blacha, line 20: forcing these private so people will be forced to use `allImagery()` which provides a sorted rule set

PR #576, opened 2020-05-06
fix(cli): root quadkey causes issues with dynamodb so never use it

blacha:


packages/cog/src/cog/cutline.ts
```diff
@@ -180,6 +180,9 @@ export class Cutline {
 
         covering.mergeQuadKeys(CoveringPercentage, minZ, minZ + 2);
 
+        /** We should never return a full cover, using '' as a index causes problems */
+        if (covering.has('')) return QuadKeyTrie.fromList(QuadKey.children(''));
+
```
jacott, line 185: How about a test for this?

PR #583, opened 2020-05-07
Feat/auto batch

blacha:


packages/cog/src/cli/cogify/action.job.ts
```diff
@@ -236,6 +238,10 @@ export class ActionJobCreate extends CommandLineAction {
             const geoJsonCoveringOutput = getJobPath(job, `covering.geojson`);
             await outputFs.writeJson(geoJsonCoveringOutput, TileCover.toGeoJson(quadkeys), logger);
 
+            if (this.submitBatch.value) {
+                await ActionBatchJob.batchJob(jobFile, true, logger);
```
blacha, line 242: --batch will auto assume --commit as there is no --commit here.

Will need to also check for --vrt

PR #584, opened 2020-05-07
fix: cutline defaults added

josephramsay:
Fixes: #
Sets a default cutline file when no arg is provided based on imagery set type

### DIFF
--- a/packages/cog/src/cog/cutline.ts
+++ b/packages/cog/src/cog/cutline.ts
@@ -8,6 +8,12 @@ import { CoveringPercentage, CutlineZoom, ZoomDifferenceForMaxImage } from './co
 import { CogJob, SourceMetadata } from './types';
 
 const PaddingFactor = 1.125;
+type CutlineItem = { path: string; blend: number };
+export const CutlineMap: { [key: string]: CutlineItem } = {
+    sentinel: { path: 's3://basemaps-cog-test/NZCoastCutline.sentinel.geojson', blend: 5 },
+    urban: { path: 's3://basemaps-cog-test/NZCoastCutline.urban.geojson', blend: 20 },
+    rural: { path: 's3://basemaps-cog-test/NZCoastCutline.rural.geojson', blend: 20 },
+};
 
 function findGeoJsonProjection(geojson: any | null): EPSG {
     return Projection.parseEpsgString(geojson?.crs?.properties?.name ?? '') ?? EPSG.Wgs84;
@@ -206,5 +212,20 @@ export class Cutline {
     static async loadCutline(path: string): Promise<Cutline> {
         const geojson = (await FileOperator.create(path).readJson(path)) as FeatureCollection;
         return new Cutline(geojson);
+    }
+
+    /**
+     * Assume a cutline default based on a provided imagery name
+     *
+     * @param imageryName the name of the dataset being cut, should contain a substring denoting dataset type; sentinel, urban, rural.
+     */
+    static defaultCutline(imageryName: string): CutlineItem {
+        const foundNames = Object.keys(CutlineMap)
+            .map((cutName) => (imageryName.toLowerCase().search(cutName) > -1 ? cutName : ''))
+            .filter((matchedName) => !!matchedName);
+        if (foundNames.length != 1) {
+            throw new Error('Matched ' + foundNames.length + ' cutline names');
+        }
+        return CutlineMap[foundNames[0]];
     }
 }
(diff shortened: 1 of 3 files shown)


packages/cog/src/cog/cutline.ts
```diff
@@ -8,6 +8,12 @@ import { CoveringPercentage, CutlineZoom, ZoomDifferenceForMaxImage } from './co
 import { CogJob, SourceMetadata } from './types';
 
 const PaddingFactor = 1.125;
+type CutlineItem = { path: string; blend: number };
+export const CutlineMap: { [key: string]: CutlineItem } = {
+    sentinel: { path: 's3://basemaps-cog-test/NZCoastCutline.sentinel.geojson', blend: 5 },
```
blacha, line 13: We cannot have the full path to these stored here as they vary based on prod vs dev.

maybe just leave it as the key `urban` `sentinel` `rural` then we can use the key in the `output` location possibly `s3://${output}/cutline/${cutline}.json`

packages/cog/src/cog/cutline.ts
```diff
@@ -8,6 +8,12 @@ import { CoveringPercentage, CutlineZoom, ZoomDifferenceForMaxImage } from './co
...
+    urban: { path: 's3://basemaps-cog-test/NZCoastCutline.urban.geojson', blend: 20 },
+    rural: { path: 's3://basemaps-cog-test/NZCoastCutline.rural.geojson', blend: 20 },
```
jacott, line 15: I don't think we should have hard coded buckets in here. We could derive the bucket from the input (or output) dir.

packages/cog/src/cog/cutline.ts
```diff
@@ -207,4 +213,19 @@ export class Cutline {
...
         return new Cutline(geojson);
     }
+
+    /**
+     * Assume a cutline default based on a provided imagery name
+     *
+     * @param imageryName the name of the dataset being cut, should contain a substring denoting dataset type; sentinel, urban, rural.
+     */
+    static defaultCutline(imageryName: string): CutlineItem {
+        const foundNames = Object.keys(CutlineMap)
```
blacha, line 223: this could be simplified with `const found = Object.keys(CutlineMap).find(f => imageryName.toLowerCase().includes(f))`

if we really want to check how many we could change the `find` to a `filter`

packages/cog/src/cog/cutline.ts
jacott: We may only be able to have one default cutline if we want to avoid making it specific to NZ. Its name should be generic like `default-cutline.geojson.gz`.

jacott:
Ultimately we may just decide it is better not to set a default for the cutline but error if a cutline is not supplied (or `--nocutline`) is not stated. And instead decide on naming and location for the cutlines which we could wrap up in a specific script for NZ imagery building.

packages/cog/src/cog/cutline.ts
josephramsay: I wondered about that. Wasn't sure where to get the bucket name. Will do

packages/cog/src/cog/cutline.ts
josephramsay: I did actually try .find() using ts-node but got this `error TS2339: Property 'find' does not exist on type 'string[]'`. Need to upgrade I assume

packages/cog/src/cog/cutline.ts
josephramsay: 1. Thanks. Was mentioned above so that's done now. 
2.The sentinel and urban/rural cutlines are distinct. Do you mean a "use this one here if its available" kinda default?

packages/cog/src/cog/cutline.ts
josephramsay: Filter version of above works as before and more concise.

packages/cog/src/cog/cutline.ts
josephramsay: As above. Used outputConfig.path

packages/cog/src/cog/cutline.ts
jacott: Maybe just have one default and for sentinel it has to be specified. That way nothing in the logic is specific to NZ. Of course where to find the default is tricky and may mean it is better to just complain if no cutline is given rather than look for it relative to source or output directories. @blacha what do you think?

josephramsay: This seems like something with a wider scope, i.e. building with a deliberate country neutral focus. In the meantime shall we just park the issue?

josephramsay:
Killing as function migrated to basemaps-teams

PR #599, opened 2020-05-11
feat(lambda-xyz): Support tags and imagery sets for WMTSCapabilities.xml

jacott:


packages/lambda-xyz/src/cli/serve.ts
```diff
@@ -95,9 +100,46 @@ async function main(): Promise<void> {
...
+            const { imageryName, projection } = req.params;
+            const ctx = new LambdaContext(
+                {
+                    httpMethod: 'get',
+                    path: `/v1/tiles/${imageryName}/${projection}/WMTSCapabilities.xml`,
+                } as any,
+                logger,
+            );
+            try {
+                const data = await lambda.handleRequest(ctx);
```
blacha, line 118: IS this the same as above? could we make a function `await runLambda(path, req, res)` which does all the try catch/res.end stuff

packages/lambda-xyz/src/wmts.capability.ts
```diff
@@ -22,12 +24,12 @@ for (const k of ImageFormatOrder) {
...
-    [TileSetType.aerial]: 'GoogleMapsCompatible',
+const tileMatrixSetId: Partial<Record<string, string>> = {
+    aerial: 'GoogleMapsCompatible',
 };
 
-const LayerPreamble: Partial<Record<TileSetType, VNodeElement[]>> = {
-    [TileSetType.aerial]: [
+const LayerPreamble: Partial<Record<string, VNodeElement[]>> = {
+    aerial: [
         V('ows:Title', 'NZ Aerial Imagery Basemap'),
```
blacha, line 33: I think we need to move this into the database.

packages/lambda-xyz/src/wmts.capability.ts
```diff
@@ -22,12 +24,12 @@ for (const k of ImageFormatOrder) {
     formats.push(V('Format', 'image/' + k));
 }
 
-const tileMatrixSetId: Partial<Record<TileSetType, string>> = {
-    [TileSetType.aerial]: 'GoogleMapsCompatible',
+const tileMatrixSetId: Partial<Record<string, string>> = {
+    aerial: 'GoogleMapsCompatible',
```
blacha, line 28: Can we do this based on EPSG code rather than name?

packages/lambda-xyz/src/routes/tile.ts
```diff
@@ -141,7 +133,10 @@ export async function Wmts(req: LambdaContext, wmtsData: TileDataWmts): Promise<
 
     const host = Env.get(Env.PublicUrlBase);
 
-    const xml = buildWmtsCapability(host, req.apiKey || '', wmtsData.tileSet, wmtsData.projection);
+    // TODO get all projections if wmtsData.projection is null
+    await loadTileSet(req, wmtsData.name, wmtsData.projection ?? EPSG.Google);
```
blacha, line 137: could we make this return a tileset[] and push that into buildWmtsCapability?

packages/lambda-xyz/src/wmts.capability.ts
jacott: Do you want to do it now?  That is a bit of planning etc. We can make another ticket for that work?

packages/lambda-xyz/src/wmts.capability.ts
jacott: I had that originally but then I have to convert back and forth and parseInt etc. This is more straight forward

packages/lambda-xyz/src/routes/tile.ts
jacott: Eventually but not needed now so don't want to future proof it until we support another projection; we may end up changing the way we do things

blacha: Ahh was more thinking there are a number of `.get(name, projection)` that assume that a `.load()` has been run inside of non async functions.

this could lead to confusion down the line if the function is called before a `.load()` is called.

if this `loadTileSet()` returns a `TileSet` it then can be passed into the function and it is already initialized.

It doesnt have to reutrn a TileSet[] but could just return TileSet then pass that into `buildWmtsCapability`

packages/lambda-xyz/src/wmts.capability.ts
blacha: Not sure I follow?
```
if( projection == EPSG.Google) return 'GooleMapsCompatible;
throw new Error(`Invalid projection: ${projection}`
```

packages/lambda-xyz/src/wmts.capability.ts
blacha: Ok lets make the database changes for the next PR

packages/lambda-xyz/src/routes/tile.ts
jacott: Now just does 3857 if missing and passes TileSet to wmts. In Future will pass an array of TileSets

packages/lambda-shared/src/aws/tile.metadata.ts
```diff
@@ -55,6 +55,10 @@ export interface TileMetadataSetRecord extends BaseDynamoTable {
     /** TileSet set name */
     name: string;
 
+    title?: string;
```
blacha, line 58: Want to add comments for what these are used for?

packages/lambda-xyz/src/__test__/xyz.test.ts
```diff
@@ -48,15 +36,15 @@ o.spec('LambdaXyz', () => {
...
+            tileSet.load = () => Promise.resolve(true);
             tileSet.getTiffsForQuadKey = async (): Promise<[]> => [];
         }
     });
 
     o.afterEach(() => {
         TileSets.clear();
-        Tilers.tile256 = new Tiler(256);
-        Tilers.compose256 = new TileMakerSharp(256);
+        Tilers.tile256 = origTile256;
```
blacha, line 46: These shouldn't have internal state, so should just be able to recreate them.

jacott: Yeah but then if you change them to something else you have to remember to change it here too. This way we are confident that it is always restored to what it was.

PR #622, opened 2020-05-14
fix(cli): git hash cannot be fetched inside the docker cli

blacha:
- inside of batch use the environment variables
- otherwise look it up from the package.json
- fall back to git-rev-sync


blacha:
This PR may kill all running batch instances

PR #624, opened 2020-05-14
feat(lambda-shared): add TileMetadataProvider

jacott:
### Changes

Add `provider` tool to configure the basemaps provider information. Provider information is used when a user requests a WMTSCapabilities.xml file.

The entire history of the provider is stored inside the metadata database and version tags are used to configure what is supplied to the user.


packages/cli/src/cli/basemaps/action.tileset.info.ts
```diff
@@ -61,7 +61,11 @@ export class TileSetInfoAction extends TileSetBaseAction {
             console.log(this.renderHelpText());
             return;
         }
-        const tsData = await Aws.tileMetadata.TileSet.get(tileSet, projection, this.version.value! ?? TileSetTag.Head);
+        const tsData = await Aws.tileMetadata.TileSet.get(
+            tileSet,
+            projection,
+            this.version.value! ?? TileMetadataTag.Head,
```
blacha, line 67: do you need a `!` if you are `??` it

jacott: That was your code; not mine; I'll correct it

jacott: Apparently we do need it?

jacott: It is because of the overload of the get method and typescript not handling it properly

packages/cli/src/cli/provider/action.provider.history.ts
```diff
@@ -0,0 +1,73 @@
...
+    protected onDefineParameters(): void {}
+
+    async getAllTags(): Promise<Map<TileMetadataTag, TileMetadataProviderRecord>> {
+        const allTags: Map<TileMetadataTag, TileMetadataProviderRecord> = new Map();
+        await Promise.all(
+            Object.values(TileMetadataTag).map(async (tag) => {
+                try {
+                    const value = await Aws.tileMetadata.Provider.get(tag);
+                    allTags.set(tag, value);
+                } catch (e) {}
```
blacha, line 27: Do we really want to swallow _all_ errors?

packages/cli/src/cli/provider/action.provider.tag.ts
```diff
@@ -0,0 +1,93 @@
...
+                return;
+        }
+
+        const before = await Aws.tileMetadata.Provider.get(tag);
+
+        LogConfig.get().info({ version, tag }, 'Tagging');
+
+        if (this.commit.value) {
+            await Aws.tileMetadata.Provider.tag(tag, version);
+            if (tag === TileMetadataTag.Production) {
```
blacha, line 74: this is a pretty huge invalidation.

also why just prod?

packages/cli/src/cli/provider/provider.util.ts
```diff
@@ -0,0 +1,103 @@
...
+            individualName: '',
+            phone: '',
+            position: '',
+        },
+        name: '',
+        site: '',
+    },
+};
+
+export function validateProvider(after: Record<string, any>, before: Record<string, any>): any {
```
blacha, line 54: Needs some comments to describe what this actually does?

also why does it return a any?

packages/cli/src/cli/provider/action.provider.tag.ts
```diff
@@ -0,0 +1,93 @@
...
+            required: false,
+        });
+    }
+
+    protected async onExecute(): Promise<void> {
+        const tagInput = this.tag.value!;
+        const version = this.version.value!;
+
+        let tag: TileMetadataTag;
+        switch (tagInput) {
```
blacha, line 56: We now do this in multiple places, possibly should export a function to do this, so we dont forget to update it in both places.

packages/lambda-xyz/src/routes/tile.ts
```diff
@@ -138,7 +140,9 @@ export async function Wmts(req: LambdaContext, wmtsData: TileDataWmts): Promise<
     // null
     const tileSet = await loadTileSet(req, wmtsData.name, wmtsData.projection ?? EPSG.Google);
 
-    const xml = tileSet == null ? null : buildWmtsCapability(host, req, tileSet);
+    const provider = await Aws.tileMetadata.Provider.get(TileMetadataTag.Production);
```
blacha, line 143: I think we will likely need more than one provider in the future, how about adding a name to this?

maybe just hard code it to `linz` for now.

packages/cli/src/cli/provider/action.provider.tag.ts
jacott: Just prod because it is a huge invalidation and we currently only use the production tag to serve wmts files

packages/lambda-xyz/src/routes/tile.ts
jacott: How about `"default"` (or `main`) rather than `linz` to be generic.

jacott: I think we should hold off supporting more than one provider until such as situation arises rather than future proofing now.

blacha: The problem we will have is since there is no provision for more than one it would require a data migration to fix it. I am not a huge fan of migrating data in production environments if i can avoid it :)  hard coding something as the name means we dont have to migrate data in the future. I could see us serving other imagery sets that we have access to that may not be `linz` as the provider.

looking at the code the provider name was ` 'Land Information New Zealand'` so what is wrong with `linz` as being the provider name?

jacott: If you are strongly in favour of adding name I will but I'm not happy with linz as the value to lookup. `main` or `default` make much more sense to me.

> looking at the code the provider name was 'Land Information New Zealand' so what is wrong with linz as being the provider name?

Because linz is nz specific and the main point of this PR is to remove linz specific code.

jacott: Another option is to add provider to the tileset record and use that as the name to lookup

blacha: > Another option is to add provider to the tileset record and use that as the name to lookup

Sounds good to me, lets just do that.

jacott: Of course that also requires a migration so I would suggest holding off until we need it

jacott: Now stores and retrieves record as `pv_main_*`

PR #631, opened 2020-05-18
feat(lambda-xyz): optimise imagery loading

jacott:
Only reload imagery if tileSet has been updated


blacha:
Im not sure we need to do this, as we cache the imagery.. all a `.load()` does is fetch keys out of a map if they have already been fetched before.

jacott:
Okay, misread the code

PR #632, opened 2020-05-18
feat(geo): support chatham projection 3793

blacha:


packages/geo/src/projection.ts
```diff
@@ -11,23 +11,31 @@ export enum EPSG {
...
 
     wgs84: EPSG.Wgs84,
     epsg4326: EPSG.Wgs84,
-    '4326': EPSG.Wgs84,
+    [EPSG.Wgs84]: EPSG.Wgs84,
 
-    nztm: EPSG.Nztm,
-    epsg2193: EPSG.Nztm,
-    '2193': EPSG.Nztm,
-    nzgd2000: EPSG.Nztm,
```
blacha, line 30: nzgd2000 was wrong, renamed it to nztm2000

palmerj: Oh yes, this is a good fix!

PR #640, opened 2020-05-18
fix(lambda-shared): fix order equal priority images are sorted

jacott:
Changed to compare year and resolution if priority the same.

Resolves linz/basemaps-team#159

### DIFF
--- a/packages/cli/src/cli/basemaps/action.tileset.history.ts
+++ b/packages/cli/src/cli/basemaps/action.tileset.history.ts
@@ -1,9 +1,8 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { Aws, LogConfig, TileMetadataImageryRecord, TileMetadataSetRecord, TileSetTag } from '@basemaps/lambda-shared';
-import * as chalk from 'chalk';
 import { CliTable } from '../cli.table';
 import { TileSetBaseAction } from './tileset.action';
-import { printTileSet, TileSetTable } from './tileset.util';
+import { printTileSet, showDiff } from './tileset.util';
 
 const MaxHistory = 199;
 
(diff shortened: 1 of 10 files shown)


packages/lambda-xyz/src/tile.set.ts
```diff
@@ -70,14 +70,8 @@ export class TileSet {
...
 
-    *allImagery(): Generator<{ rule: TileMetadataImageRule; imagery: TileMetadataImageryRecord }> {
-        const rules = Aws.tileMetadata.TileSet.rules(this.tileSet);
-        for (const rule of rules) {
-            const imagery = this.imagery.get(rule.id);
-            if (imagery == null) throw new Error(`Unable to find imagery for key: ${rule.id}`);
-
-            yield { rule, imagery };
-        }
+    *allImagery(): Generator<TileSetRuleImagery> {
```
blacha, line 73: I think we can just remove this, and expose the `imagery` again, the reason it was made private was because it was no longer sorted.

packages/cli/src/cli/basemaps/action.tileset.history.ts
```diff
@@ -94,19 +94,19 @@ export class TileSetHistoryAction extends TileSetBaseAction {
     showDiff(
```
blacha, line 94: somewhat related,  could you make this static, or possibly move it into the utils function, would be helpful to reuse in other places :)

packages/lambda-shared/src/aws/tile.metadata.imagery.ts
```diff
@@ -1,4 +1,23 @@
-import { TileMetadataTable, TileMetadataImageryRecord, TileMetadataSetRecord } from './tile.metadata';
+import {
+    TileMetadataImageryRecord,
+    TileMetadataSetRecord,
+    TileMetadataTable,
+    TileSetRuleImagery,
+} from './tile.metadata';
+
+function compareImageSets(a: TileSetRuleImagery, b: TileSetRuleImagery): number {
+    let diff = a.rule.priority - b.rule.priority;
```
blacha, line 9: I find this quite hard to reason about, I know some people like the single return but without reading this entire function you have no idea what diff is going to be.

To me this is much more understandable
https://github.com/linz/basemaps/commit/c7ba7993e1544f7d120f7612d17b6f427549d716#diff-08397e95c4ab9c39c3653125000cfb3eR13

And has the helpful addition of comments explaining why the comparisons are being made.

PR #664, opened 2020-05-21
feat: Allow composite imagery from different COG buckets

jacott:
This also requires a basemaps-team dynamoDB migration script See PR https://github.com/linz/basemaps-team/pull/186

The bucket location is now stored in each imagery record


packages/lambda-shared/src/aws/tile.metadata.base.ts
```diff
@@ -39,6 +39,8 @@ export interface TileMetadataImageryRecord extends BaseDynamoTable {
 
     projection: EPSG.Google;
 
+    uriBase: string;
```
blacha, line 42: could you add a comment explaining what this is used for?

packages/lambda-xyz/src/tile.set.ts
```diff
@@ -20,25 +17,24 @@ export class TileSet {
...
+     */
     static BasePath(record: TileMetadataImageryRecord, quadKey?: string): string {
-        const id = TileMetadataTable.unprefix(RecordPrefix.Imagery, record.id);
-        const basePath = [record.projection, record.name, id].join('/');
         if (quadKey == null) {
-            return basePath;
+            return record.uri;
         }
-        return path.join(basePath, `${quadKey}.tiff`);
+        return path.join(record.uri, `${quadKey}.tiff`);
```
blacha, line 30: path.join doesnt work with uris :(

```
path.join('s3://foo/bar','baz')
's3:/foo/bar/baz'
```

jacott: No test for this!

jacott: I've also down cased the method name

packages/lambda-xyz/src/tile.set.ts
```diff
@@ -20,25 +16,27 @@ export class TileSet {
...
+    static basePath(record: TileMetadataImageryRecord, quadKey?: string): string {
         if (quadKey == null) {
-            return basePath;
+            return record.uri;
         }
-        return path.join(basePath, `${quadKey}.tiff`);
+        if (record.uri.endsWith('/')) {
+            throw new Error("Invalid uri ending with '/' " + record.uri);
+        }
+        return record.uri + `/${quadKey}.tiff`;
```
blacha, line 32: Mixing + and template strings?

could just `${record.uri}/${quadKey.tiff}`

PR #671, opened 2020-05-21
refactor: remove buffer function from compositions

blacha:
let other classes use tiff.getTile()

not everything that consumes this API needs a `Buffer` internally this is all done as a  `UInt8Array` so we shouldnt need to force `Buffers` on everyone


packages/lambda-xyz/src/routes/tile.ts
```diff
@@ -107,7 +107,8 @@ export async function Tile(req: LambdaContext, xyzData: TileDataXyz): Promise<La
 
     if (!Env.isProduction()) {
```
jacott, line 108: Would it be nicer here to either just test if debug is turned on or test both?

blacha: Eventually i think we should possibly have a `DEBUG=:area` and let that configure the debugging logging for specific areas of the code base.

eg `DEBUG=tiler` or something

PR #672, opened 2020-05-22
fix: do not use full tiff files for generating etags

blacha:
as the tiff is often not fully loaded this could cause werid etag generation
instead use just the name of the tiff


packages/lambda-xyz/src/__test__/xyz.test.ts
```diff
@@ -85,7 +89,7 @@ o.spec('LambdaXyz', () => {
         const res = await handleRequest(request);
         o(res.status).equals(200);
         o(res.header('content-type')).equals('image/webp');
-        o(res.header('eTaG')).equals('w8RW7/QQDHDMiiNDlbACrigBcCEveyedjHn7QMX6kDg=');
+        o(res.header('eTaG')).equals('foo');
```
jacott, line 92: You do not appear to be asserting that `generate` is being called with the correct arguments

packages/lambda-xyz/src/__test__/xyz.test.ts
```diff
@@ -18,6 +19,7 @@ o.spec('LambdaXyz', () => {
     const rasterMockBuffer = Buffer.from([1]);
     const origTile256 = Tilers.tile256;
     const origCompose256 = Tilers.compose256;
+    const origTileEtag = TileEtag.generate;
```
jacott, line 23: You do not appear to be restoring this

blacha: its restored on line 53

packages/lambda-xyz/src/__test__/xyz.test.ts
blacha: I am unsure how much we need to test, considering typescript is already checking that the correct values are passed in. Also the values looked to be completely mocked anyway.

PR #676, opened 2020-05-26
Feat/tile matrix sets

blacha:
Fixes: #


packages/geo/src/__tests__/tile.matrix.set.test.ts
```diff
@@ -0,0 +1,71 @@
+import * as o from 'ospec';
+import { GoogleTms } from '../tms/google';
+import { Projection } from '../projection';
+import { approxEqual } from './bounds.tile.test';
+import { Epsg } from '../epsg';
+
+o.spec('TileMatrixSet', () => {
```
palmerj, line 7: Going to add some NZTM2000 tests?

blacha: This is very much a work in progress (hence draft) mostly made the PR to see if it builds :)

PR #688, opened 2020-05-28
feat(geo): convert quadkey to/from tile index

blacha:


packages/geo/src/quad.key.ts
```diff
@@ -56,6 +61,45 @@ export const QuadKey = {
...
+        let x = 0;
+        let y = 0;
+        const z = quadKey.length;
+
+        for (let i = z; i > 0; i--) {
+            const mask = 1 << (i - 1);
+            const q = quadKey.charCodeAt(z - i);
+            if (q === CHAR_1) x |= mask;
+            if (q === CHAR_2) y |= mask;
+            if (q === CHAR_3) {
```
jacott, line 78: Why using char code? I think this is actually slower than `quadkey[z-1]`

packages/geo/src/quad.key.ts
```diff
@@ -56,6 +61,45 @@ export const QuadKey = {
...
+        let quadKey = '';
+        for (let zI = z; zI > 0; zI--) {
+            let b = 0;
+            const mask = 1 << (zI - 1);
+            if ((x & mask) !== 0) b++;
+            if ((y & mask) !== 0) b += 2;
+            quadKey += b.toString();
+        }
+        return quadKey;
+    },
```
jacott, line 101: I do this slightly differently

```typescript
function xyzToQuadKey(x: number, y: number, z: number): string {
    let level = 1 << z;
    if (x < 0 || x > level || y < 0 || y > level) return '';

    let qk = '';
    let hl = 0;
    let mask = 0;
    while (level > 1) {
        hl = level >> 1;
        mask = hl - 1;

        if (x < hl) {
            qk += y < hl ? '2' : '0';
        } else {
            qk += y < hl ? '3' : '1';
        }
        level = hl;
        x &= mask;
        y &= mask;
    }
    return qk;
}
```

packages/geo/src/quad.key.ts
blacha: Im pretty sure using array indexes on strings actually creates a new string, which is kinda slow, I'll benchmark it

blacha: node 12

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `node tests/char.code.js` | 398.1 ± 5.8 | 391.1 | 407.4 | 1.0 |
| `node tests/array.index.js` | 447.0 ± 7.6 | 440.7 | 466.0 | 1.1 |

node 14

| Command | Mean [ms] | Min [ms] | Max [ms] | Relative |
|:---|---:|---:|---:|---:|
| `node tests/char.code.js` | 393.3 ± 9.4 | 383.4 | 427.2 | 1.0 |
| `node tests/array.index.js` | 560.6 ± 16.0 | 541.0 | 617.3 | 1.4 |

packages/geo/src/quad.key.ts
blacha: This logic is flawed, when running x:0,y:0,z:30, I get `2222222222...` when you should be getting `000000`

interestingly it is slightly faster though, however switching from a `b.toSTring()` to a `String.fromCharCode()` makes them about the same speed but the PR one is giving the correct answer

PR #689, opened 2020-05-28
feat(wmts): support multiple layers and multiple projections

blacha:
Fixes: #

### DIFF
--- a/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
+++ b/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
@@ -1,142 +1,120 @@
 import { Epsg } from '@basemaps/geo';
-import { HttpHeader, VNodeElement } from '@basemaps/lambda-shared';
+import { V, VNodeElement } from '@basemaps/lambda-shared';
 import { createHash } from 'crypto';
 import * as o from 'ospec';
-import { TileSet } from '../tile.set';
-import { TileSets } from '../tile.set.cache';
-import { buildWmtsCapability, buildWmtsCapabilityToVNode } from '../wmts.capability';
-import { addTitleAndDesc, mockRequest, Provider } from './xyz.testhelper';
+import { WmtsCapabilities } from '../wmts.capability';
+import { Provider, FakeTileSet } from './xyz.util';
 
-const listTag = (node: VNodeElement, tag: string): string[] => Array.from(node.tags(tag)).map((n) => n.toString());
+function tags(node: VNodeElement | null | undefined, tag: string): VNodeElement[] {
+    if (node == null) return [];
+    return [...node.tags(tag)];
+}
+function listTag(node: VNodeElement | null | undefined, tag: string): string[] {
+    return tags(node, tag).map((n) => n.toString());
+}
 
-const TileSetNames = ['aerial', 'aerial@beta', '01E7PJFR9AMQFJ05X9G7FQ3XMW'];
+o.spec('WmtsCapabilities', () => {
+    const apiKey = 'secret1234';
+    const tileSet = new FakeTileSet('aerial', Epsg.Google);
+    const tileSetImagery = new FakeTileSet('01E7PJFR9AMQFJ05X9G7FQ3XMW', Epsg.Google);
 
-o.spec('wmts', () => {
-    o.beforeEach(() => {
-        for (const name of TileSetNames) {
-            const tileSet = new TileSet(name, Epsg.Google);
-            addTitleAndDesc(tileSet);
-            TileSets.set(tileSet.id, tileSet);
-        }
-    });
+    o('should build capability xml for tileset and projection', () => {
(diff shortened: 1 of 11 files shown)


packages/lambda-xyz/src/__test__/wmts.capability.test.ts
```diff
@@ -1,142 +1,120 @@
...
-            Provider,
-            TileSets.get('01E7PJFR9AMQFJ05X9G7FQ3XMW_3857')!,
-        )!;
+        const raw = new WmtsCapabilities('https://basemaps.test', Provider, [tileSetImagery]).toVNode();
 
-        const tms = raw?.find('TileMatrixSet', 'ows:Identifier')!;
+        const tms = raw?.find('TileMatrixSet', 'ows:Identifier');
 
-        o(tms.textContent).equals('GoogleMapsCompatible');
+        o(tms?.textContent).equals('WebMercatorQuad');
```
jacott, line 118: Why the name change

packages/lambda-xyz/src/__test__/xyz.test.ts
```diff
@@ -175,8 +171,8 @@ o.spec('LambdaXyz', () => {
             );
 
             const vdom = await VNodeParser.parse(body);
-            const url = vdom.tags('ResourceURL').next().value!;
-            o(url.toString()).equals(
+            const url = vdom.tags('ResourceURL').next().value;
+            o(url?.toString()).equals(
                 '<ResourceURL format="image/png" resourceType="tile" ' +
```
jacott, line 176: Why do you prefer `url?.` to `url!` in tests?

blacha: given `x = { a: 1} | null`

```
x = null
y = x.a!  // Runtime exception 
y = x?.a // y = undefined
```

inside of tests
```
o(x.a!).equals('foobar') // TypeError: Cannot read property 'a' of null
o(x?.a).equals('foobar') // expected undefined to equal "foobar"
```

and generated code

```
var x = null;
console.log(x.a);
console.log(x === null || x === void 0 ? void 0 : x.a);
```

So in general never use `!` unless you know for a fact that the thing is never going to be null, eslint also warns me about `!` usage as its not really good to use.

packages/lambda-xyz/src/__test__/wmts.capability.test.ts
blacha: I used http://docs.opengeospatial.org/is/17-083r2/17-083r2.html#73 for the source of our tile matrix definition. it uses `WebMercatorQuad`

packages/lambda-xyz/src/__test__/wmts.capability.test.ts
```diff
@@ -1,142 +1,120 @@
...
+            V('Format', 'image/webp').toString(),
+            V('Format', 'image/jpeg').toString(),
         ]);
 
-        o(listTag(layer, 'ows:WGS84BoundingBox')).deepEquals([
-            '<ows:WGS84BoundingBox crs="urn:ogc:def:crs:OGC:2:84">\n' +
-                '  <ows:LowerCorner>-180 -85.0511287798066</ows:LowerCorner>\n' +
-                '  <ows:UpperCorner>180 85.0511287798066</ows:UpperCorner>\n' +
-                '</ows:WGS84BoundingBox>',
+        // FIXME I dont think this this really needed? we will need to reproject to get these values
```
palmerj, line 42: From memory this is important in some clients such as ArcGIS Desktop (not pro). We should test this out once in dev.

packages/lambda-xyz/src/__test__/wmts.capability.test.ts
```diff
@@ -145,4 +123,60 @@ o.spec('wmts', () => {
...
+        o(layers.length).equals(2);
+
+        o(layers[0].find('ows:Title')?.textContent).equals('aerial-title');
+        o(layers[0].find('TileMatrixSet')?.textContent).equals('NZTM2000');
+
+        o(layers[1].find('ows:Title')?.textContent).equals('imagery-title');
+        o(layers[1].find('TileMatrixSet')?.textContent).equals('NZTM2000');
+    });
+
+    o('should support multiple different projections on differnt tiles sets', () => {
```
palmerj, line 167: I don't see any tests for the NZTM2000 matrixset (zoom level ID, tile dimension, scale etc)? I think we should check that as well

blacha: The NZTM2000 matrix set tests are coming in the next PR.

packages/lambda-xyz/src/__test__/wmts.capability.test.ts
blacha: Ahh good to know, we are looking to add it back in once we add `proj` to this area of the codebase.

PR #699, opened 2020-05-29
feat: render tiles using tile matrix sets

blacha:
Fixes: #


packages/lambda-xyz/src/cli/dump.ts
```diff
@@ -0,0 +1,48 @@
...
+            httpMethod: 'get',
+            path: `/v1/tiles/${tileSet.name}/${projection.code}/${xyz.z}/${xyz.x}/${xyz.y}.${ext}`,
+        } as any,
+        logger,
+    );
+
+    const tileData = await Tile(ctx, { ...xyz, projection, name: tileSet.name, ext, type: TileType.Image });
+
+    await writeFileSync(`output_${xyz.x}_${xyz.y}_z${xyz.z}.${ext}`, tileData.body);
+
```
jacott, line 41: spurious await

packages/tiler/src/tiler.ts
```diff
@@ -55,20 +51,25 @@ export class Tiler {
...
-        const screenBoundsPx = this.projection.getPixelsFromTile(x, y);
+        const screenPx = this.tms.tileToPixels(x, y);
+        const screenBoundsPx = new Bounds(screenPx.x, screenPx.y, this.tms.tileSize, this.tms.tileSize);
+
         /** Raster pixels of the input geotiff */
-        const tiffBoundsPx = this.projection.getPixelsBoundsFromMeters(extentMeters, zoom);
+        const bbox = tiff.images[0].bbox;
+        const ul = this.tms.sourceToPixels(bbox[0], bbox[3], zoom);
+        const lr = this.tms.sourceToPixels(bbox[2], bbox[1], zoom);
+        const tiffBoundsPx = Bounds.fromUpperLeftLowerRight(ul, lr);
```
jacott, line 63: This is something I can use in COG create. Can we make this its own method?

blacha: It was a method, I am not sure the bbox's are standard though, so dont really want to add it to the `TileMatrixSet` class..

I am then unsure where else to put it.

PR #723, opened 2020-06-03
refactor: split lambda-shared into seperate packages

blacha:


packages/lambda/src/lambda.ts
```diff
@@ -19,14 +17,13 @@ export class LambdaFunction {
      */
     public static wrap(
         fn: (req: LambdaContext) => Promise<LambdaHttpResponse>,
+        logger: any, // TODO
```
jacott, line 20: Is this still TODO?

blacha: Good catch, will fix

PR #728, opened 2020-06-03
Refactor geo: Remove Google Mercator dependency

jacott:
In preparation for generating COGs for EPSG NZTM2000 and simplifying the generation logic.

Removes all methods from `geo` package that assume a Google projection.


packages/cli/src/proj/index.ts
```diff
@@ -6,16 +6,10 @@ import { Citm2000 } from './citm2000';
...
 proj4.defs(Epsg.Citm2000.toEpsgString(), Citm2000);
 
-export function getProjection(fromProjection: Epsg, toProjection?: Epsg): proj4.Converter | null {
-    try {
-        return proj4(fromProjection.toEpsgString(), toProjection?.toEpsgString());
-    } catch (e) {
-        return null;
-    }
+export function getProjection(fromProjection: Epsg, toProjection?: Epsg): proj4.Converter {
+    return proj4(fromProjection.toEpsgString(), toProjection?.toEpsgString());
```
blacha, line 10: This throws a error, do we want this to throw ? it returns a bad error

packages/geo/src/__tests__/bounds.tile.test.ts
```diff
@@ -1,139 +0,0 @@
-import { Bounds } from '../bounds';
-import { Projection } from '../projection';
-import * as o from 'ospec';
-import { Approx } from '@basemaps/test';
-
-o.spec('TilingBounds', () => {
```
blacha, line 6: I think these tests are useful, we should maybe move them somewhere

PR #736, opened 2020-06-04
fix(cli): fix regression in calculating image zoom resolution

jacott:
The refactor of removing Google Mercator code from `geo` broke the logic in `getTiffResolution`

### DIFF
--- a/packages/cli/src/cog/builder.ts
+++ b/packages/cli/src/cog/builder.ts
@@ -14,6 +14,8 @@ export const InvalidProjectionCode = 32767;
 export const CacheFolder = './.cache';
 export const TileSize = 256;
 
+const InitialResolution = (2 * Math.PI * 6378137.0) / TileSize;
+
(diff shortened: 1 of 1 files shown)


packages/cli/src/cog/builder.ts
```diff
@@ -119,7 +121,7 @@ export class CogBuilder {
         const [resX] = image.resolution;
         let z = 30;
         while (z > 0) {
-            const currentZoom = TileSize / (1 << z);
+            const currentZoom = InitialResolution / (1 << z);
```
blacha, line 124: might be a idea to leave this as a function and just call said function

jacott: This logic is all going to change once we incorporate NZTM2000

blacha: its more to prevent bugs like this :)

PR #738, opened 2020-06-05
Refactor/cli/remove using cover and trie

jacott:
Reduce the complexity of cog creation code.


packages/cli/src/cog/cutline.ts
```diff
@@ -178,42 +115,89 @@ export class Cutline {
...
+     * @param minZ Don't produce any quadKeys of this length or less and creater than this length + 2
+     * @param cf merge tiles that cover at least this fraction
+     * @return the quadKeys and percentage covered of the world by this srcArea set
+     */
+    private makeQuadKeys(
+        quadKey: string,
+        srcArea: MultiPolygon,
+        minZ: number,
+        cf: number,
+    ): { qks: string[]; af: number } {
```
blacha, line 153: could we be more descriptive? `quadKeys` `coveringFactor` 

what is a `af` ?

packages/cli/src/cog/tms.util.ts
```diff
@@ -8,22 +8,14 @@ export interface LatLon {
...
+        const ul = tms.tileToSource(tile);
+        const lr = tms.tileToSource({ x: tile.x + 1, y: tile.y + 1, z: tile.z });
+        return new Bounds(ul.x, lr.y, Math.abs(lr.x - ul.x), Math.abs(lr.y - ul.y));
     },
 
     /**
-     * Convert a tile to a GeoJson Polygon in Wsg84 units
+     * Convert a tile to a GeoJson Polygon in Source units
      */
     tileToPolygon(tms: TileMatrixSet, tile: Tile): [number, number][][] {
```
blacha, line 20: is this `[number, number][][]` a geojson type ?

packages/cli/src/cog/cutline.ts
jacott: af => actualFraction

packages/cli/src/cog/tms.util.ts
jacott: `Position[][]`

PR #739, opened 2020-06-05
refactor(cli): Remove more Google Mercator only code

jacott:
Move projection logic from cli to shared

Add `--target-projection` to `cogify job`


packages/cli/src/cli/cogify/action.job.ts
```diff
@@ -86,6 +87,12 @@ export class ActionJobCreate extends CommandLineAction {
         if (this.cutline?.value) {
             cutline = { source: this.cutline.value, blend: this.cutlineBlend.value ?? 20 };
         }
+
+        const targetProjection = ProjectionTileMatrixSet.tryGet(this.targetProjection?.value);
+        if (targetProjection == null) throw new Error('Invalid target-projection');
+        if (targetProjection.tms.projection !== Epsg.Google)
+            throw new Error('Only currently working for ' + Epsg.Google.toEpsgString());
```
blacha, line 94: brackets it when multi lining.

packages/shared/src/tms/projection.tile.matrix.set.ts
```diff
@@ -0,0 +1,96 @@
...
+        const ptms = CodeMap.get(epsgCode);
+        if (ptms != null) return ptms;
+        throw new Error(`Invalid projection: ${epsgCode}`);
+    }
+
+    static tryGet(epsgCode?: EpsgCode): ProjectionTileMatrixSet | null {
+        return (epsgCode && CodeMap.get(epsgCode)) ?? null;
+    }
+
+    get forward(): (coordinates: Position) => Position {
```
blacha, line 46: do we want to name these more inline what they are doing, since they are always to/from wgs84

blacha: eg `sourceToWgs84()` ` wgs84toSource`

PR #744, opened 2020-06-07
Refactor/cli/replace remaining google specific code

jacott:
Replace the last parts of the code where it assumes projecting to Google Mercator


packages/cli/src/cog/__test__/quadkey.vrt.test.ts
```diff
@@ -14,191 +17,272 @@ o.spec('quadkey.vrt', () => {
...
+    function writePaddingQuadKey(quadKey = '31', proj = googleProj): void {
+        const tile = QuadKey.toTile(quadKey);
+        const qkBounds = proj.tileToSourceBounds(tile);
+        const px = proj.tms.pixelScale(tile.z);
+        const qkPadded = qkBounds.scaleFromCenter((qkBounds.width + px * 10) / qkBounds.width);
+
+        const { toWsg84 } = proj;
+
+        writeFileSync(
+            process.env.HOME + '/tmp/poly.geojson',
```
blacha, line 33: I doubt most people have a `~/tmp`

packages/cli/src/cog/cutline.ts
```diff
@@ -136,47 +143,80 @@ export class Cutline {
...
-                ans.quadKeys.length > 1 &&
-                quadKey.length >= minZ
-            )
-                ans.quadKeys = [quadKey];
-
-            return ans;
         }
+
+        const ans: { quadKeys: string[]; fractionCovered: number } = { quadKeys: [], fractionCovered: 0 };
+        for (let i = 0; i < 4; ++i) {
```
blacha, line 201: `i` is really weird here, we also have a const of all the character names for qks.

```
for (const key of QuadKey.Keys) { ... }
```

packages/cli/src/cog/cutline.ts
```diff
@@ -136,47 +143,80 @@ export class Cutline {
...
+                quadKey + i.toString(),
+                intArea,
+                minZ,
+                coveringFraction,
+            );
+            if (fractionCovered == 0) continue;
+            ans.fractionCovered += fractionCovered * 0.25;
+            ans.quadKeys = ans.quadKeys.concat(quadKeys);
+        }
+        if (
```
blacha, line 212: this is a big if, could this be broken out into smaller ifs, also use brackets.

packages/cli/src/cog/job.ts
```diff
@@ -134,9 +126,17 @@ export const CogJobFactory = {
             logger.info(
                 {
                     // Size of the biggest image
-                    big: getTileSize(firstQk, metadata.resolution),
+                    big: ctx.targetProjection.getImageWidth(
+                        metadata.resZoom,
+                        firstQk.length,
+                        ctx.targetProjection.tms.tileSize * 1.5,
```
blacha, line 132: We generate `BLOCKSIZE` sized tiles not `tileSize`

packages/cli/src/cog/quadkey.vrt.ts
```diff
@@ -1,71 +1,133 @@
...
         const inputTotal = job.source.files.length;
 
         cutline.filterSourcesForQuadKey(quadKey, job, sourceGeo);
 
-        const useTifs = new Set<string>(job.source.files.map((n) => basename(n)));
-
-        if (useTifs.size == 0) {
-            logger.warn({ quadKey }, 'EmptySourceImagery');
+        if (job.source.files.length == 0) {
+            return '';
```
blacha, line 101: be explicit in handling edge cases,  `return null`

packages/cli/src/gdal/gdal.config.ts
```diff
@@ -8,11 +10,16 @@ export type GdalCogBuilderOptionsResampling =
     | 'mode';
 
 export interface GdalCogBuilderOptions {
+    /**
+     * How to align levels
+     */
+    tilingScheme: string;
```
blacha, line 16: enum ?

packages/__tests__/src/rounding.ts
```diff
@@ -0,0 +1,24 @@
+export function round(z = 8): (n: number) => number {
+    const p = 10 ** z;
+    return (n: number): number => Math.round(n * p) / p;
+}
+
+export function roundJson(json: any, z = 8): any {
```
blacha, line 6: why json?  you dont `JSON.parse()` it?

packages/cli/package.json
```diff
@@ -26,8 +26,8 @@
     "@rushstack/ts-command-line": "^4.3.13",
     "ansi-colors": "^4.1.1",
     "git-rev-sync": "^2.0.0",
-    "martinez-polygon-clipping": "^0.7.0",
     "p-limit": "^2.2.1",
+    "polygon-clipping": "^0.14.3",
```
blacha, line 30: Whats the difference?

packages/cli/src/cli/cogify/action.batch.ts
```diff
@@ -99,7 +98,7 @@ export class ActionBatchJob extends CommandLineAction {
         isCommit: boolean,
     ): Promise<{ jobName: string; jobId: string; memory: number }> {
         const jobName = ActionBatchJob.id(job, quadKey);
-        const alignmentLevels = job.source.resolution - quadKey.length;
+        const alignmentLevels = job.source.resZoom - quadKey.length;
```
blacha, line 101: This is the math, I was talking about a few days ago being wrong

GDAL uses the blocksize for this, so bigger block size = less alignment levels

packages/cli/src/cog/builder.ts
```diff
@@ -120,34 +117,31 @@ export class CogBuilder {
...
         // Get best image resolution
-        const [resX] = image.resolution;
-        let z = 30;
-        while (z > 0) {
-            const currentZoom = InitialResolution / (1 << z);
-            if (currentZoom >= resX) {
-                break;
+        let z = -1;
+        for (const zoom of this.targetProj.tms.zooms) {
+            z = parseInt(zoom.identifier);
```
blacha, line 141: the index is the zoom, Im not sure we want `zoom.identifier` knowledge here, 

```
for (let z  =0; z < zooms.length...)

```

should this also throw if it cannot find one?

packages/cli/src/cog/__test__/quadkey.vrt.test.ts
jacott: Only used for debugging not invoked in the code. If someone wants to use it they can create a tmp dir

packages/cli/src/cog/job.ts
jacott: Do we want the BLOCKSIZE hard coded? Do we want it the same for all Epsg codes?

packages/cli/package.json
jacott: The other broke when zoomed in a lot. This one seems more correct is based off the other one but uses later dependencies but was released longer ago

packages/cli/src/cli/cogify/action.batch.ts
jacott: Do you have the formula we want to use?

packages/cli/src/cog/builder.ts
jacott: zooms is a sparse array so `for(;i < len;++i)` not appropriate with sparse array

packages/__tests__/src/rounding.ts
jacott: Meaning Plain old javascript objects; not classes or functions etc.

packages/cli/src/cog/job.ts
blacha: This is where GDAL math might come in handy, it lets us have a more fine grained choice of target resolution. 

I believe we are likely wanting to have a bigger block size, 512, 1024 or 2048 because it gives a better image quality. but we will want to calculate which one based on image vs tms pixel scales.

packages/__tests__/src/rounding.ts
blacha: If i see `json: any` it normally implies the `any` is a string

could be `roundObject` Also not a huge fan of the any's especially not returning the any

I wonder if you could type it with something like `roundObject<T extends Object>(obj:T, z =8): T { ... }`

packages/cli/src/cog/builder.ts
blacha: it shouldnt be a sparse array? what TMS has it as sparse?

packages/cli/src/cog/__test__/quadkey.vrt.test.ts
blacha: I think every other test like this just jams it in the current directory, you need to change this too as its causing a LGTM error, easiest way is just to export the function.

packages/cli/src/cog/cutline.ts
jacott: Not sure what to do about the if statement. I added comments for each and line

packages/cli/src/cog/builder.ts
jacott: line 36 `tile.matrix.set.ts`

```typescript
        for (const z of def.tileMatrix) {
            const zoomIndex = parseInt(z.identifier);
            if (this.zooms[zoomIndex]) throw new Error(`Duplicate tileMatrix identifier ${z.identifier}`);
            if (z.tileHeight != z.tileWidth) throw new Error('Only square tiles supported');
            if (z.tileHeight != this.tileSize) throw new Error('All tiles must have the same tile size');
            this.zooms[parseInt(z.identifier)] = z;
        }
```

Also not the redundant parseInt instead of using `zoomIndex`

packages/__tests__/src/rounding.ts
jacott: Is can round non objects too. How about `export function roundAnything(thing: any, z = 8): any {`?

blacha: or `round` and make the other one `roundNumber` or `roundFunc`

packages/cli/src/cog/builder.ts
blacha: This actually sets all the values between 0-X, i am not sure it is a vaild TMS unless it has a definition for all zoom levels.

blacha: also unsure if we should really be `parseInt` it in the first place, might be a idea to convert this to a `for( let i =0; ...`

or even a `this.zooms = def.tileMatrix.map(c => ...`

packages/cli/src/cog/__test__/quadkey.vrt.test.ts
jacott: I'll just rip it out for now

packages/__tests__/src/rounding.ts
jacott: Now makeRound and round

packages/cli/src/cog/builder.ts
jacott: If we use `map` we need to ensure that the id matches the index. Shall I change the code to do that?

packages/cli/src/cli/cogify/action.batch.ts
blacha: I do not.

packages/cli/src/cog/builder.ts
blacha: This is my uncertainty, I do not know if that is part of the spec. for the TMS we currently have the id is a string that is the zoom index

blacha: Looking at the spec, this identifier is a `ows:CodeType` which doesnt seem to correlate to a zoom index :shrug:

packages/cli/src/cli/cogify/action.batch.ts
jacott: When adding a datum an `blockFactor` can also now be given which defaults to 2. The `blockSize` is `blockFactor * tms.tileSize`

packages/cli/src/cog/builder.ts
jacott: Now sorts the zooms by `scaleDenominator` and maps to `this.zooms`

PR #764, opened 2020-06-10
feat(landing): track with google analytics if $GOOGLE_ANALYTICS is set

blacha:

### DIFF
--- a/packages/landing/deploy.js
+++ b/packages/landing/deploy.js
@@ -20,12 +20,34 @@ async function getHash(Bucket, Key) {
     return null;
 }
 
+/** If $GOOGLE_ANALYTICS is set, insert the tracking code into the index.html */
+async function insertGoogleAnalytics() {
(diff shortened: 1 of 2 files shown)


packages/landing/deploy.js
```diff
@@ -20,12 +20,34 @@ async function getHash(Bucket, Key) {
...
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', '${trackingId}');
+    </script>
+</html>`;
+
+    // Put the tracking at the bottom of the index.html
+    const outputHtml = rawIndexHtml.toString().replace('</html>', trackingCode);
+    await fs.writeFile('./static/index.html', outputHtml);
```
jacott, line 41: A script tag can't go directly into an html tag

blacha: I mean it "can" and it works in both chrome and firefox, ideally i guess it should live inside the head tag with a "defer"

PR #772, opened 2020-06-15
Feat/cli/improve cog aliged levels

jacott:
1. fix alignment levels to not assume powers of two
1. set explicit targetRes (`-tr`) for cog creation
1. use `-tap` 
    (target aligned pixels) align the coordinates of the extent of the output file to the values of the -tr, such that the aligned extent includes the minimum extent.

### DIFF
--- a/packages/__tests__/src/rounding.ts
+++ b/packages/__tests__/src/rounding.ts
@@ -15,10 +15,15 @@ export function round(thing: any, z = 8): any {
         if (typeof obj === 'number') {
             return r(obj);
         }
+        if (typeof obj === 'string') {
+            const n = +obj;
+            if (isNaN(n)) return obj;
+            return r(n).toString();
+        }
         if (Array.isArray(obj)) {
             return obj.map(recurse);
         }
-        if (obj == null || typeof obj !== 'object') return obj;
+        if (obj == null || obj.constructor !== Object) return obj;
 
         const ans = Object.assign({}, obj);
         for (const key in ans) {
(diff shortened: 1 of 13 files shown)


packages/shared/src/tms/projection.tile.matrix.set.ts
```diff
@@ -77,6 +77,25 @@ export class ProjectionTileMatrixSet {
...
+     */
+    getTiffResZoom(resX: number, blockFactor = 1): number {
+        // Get best image resolution
+        const { tms } = this;
+        let z = 0;
+        for (; z < tms.zooms.length; ++z) {
+            if (tms.pixelScale(z) <= resX * blockFactor) return z;
+        }
+        if (z == tms.zooms.length) return z - 1;
+        return -1;
```
blacha, line 96: do we want to throw here? does anything check for `== -1`?

packages/__tests__/src/rounding.ts
```diff
@@ -15,10 +15,15 @@ export function round(thing: any, z = 8): any {
         if (typeof obj === 'number') {
             return r(obj);
         }
+        if (typeof obj === 'string') {
+            const n = +obj;
```
blacha, line 19: Try and avoid `+` Could we just `Number(obj)` ?

packages/__tests__/src/rounding.ts
```diff
@@ -15,10 +15,15 @@ export function round(thing: any, z = 8): any {
...
+            if (isNaN(n)) return obj;
+            return r(n).toString();
+        }
         if (Array.isArray(obj)) {
             return obj.map(recurse);
         }
-        if (obj == null || typeof obj !== 'object') return obj;
+        if (obj == null || obj.constructor !== Object) return obj;
```
blacha, line 26: What is this protecting against?

jacott: It stops copying things like logger which we then can't deep equal against

PR #774, opened 2020-06-16
feat!: replace quadkeys with tiles and bounding boxes

jacott:
1. Replace quadKey logic with bounding box intersects
2. Name COGS `z-x-y.tiff`

BREAKING CHANGE: This commit need `TileMetadataImageryRecords` to be migrated before deployment otherwise existing imagery will fail to be rendered. Migrated records will still refer to COGs by their quadKey name rather than the updated `z-x-y` name.


packages/cli/src/cli/cogify/action.batch.ts
```diff
@@ -71,7 +73,8 @@ export function createImageryRecordFromJob(job: CogJob): TileMetadataImageryReco
         projection,
         year: extractYearFromName(job.name),
         resolution: extractResolutionFromName(job.name),
-        quadKeys: job.quadkeys,
+        bounds: job.bounds,
+        cogBounds: job.cogBounds,
```
blacha, line 77: I dont like `cogBounds` here, This doesnt actually have to be a COG, it could be any tiff file, it also is not just the `bounds`

could we have maybe `files` ?

packages/cli/src/cog/__test__/cog.test.ts
```diff
@@ -29,16 +30,14 @@ o.spec('cog', () => {
...
             o(convertArgs[0].info).equals(logger.info);
 
-            // -projwin 18472078.003508832 -5948635.289265559 18785164.071364917 -6261721.357121641
-            // -projwin_srs EPSG:3857
             const { config } = gdalCogBuilder!;
             config.bbox = round(config.bbox, 4);
             config.targetRes = round(config.targetRes, 4);
             o(config).deepEquals({
-                bbox: [17532819.7999, -5009377.0857, 20037527.452, -7514084.7378],
+                bbox: [17532819.7999, -5009377.0857, 20037508.3428, -7514065.6285],
```
blacha, line 40: how come this changed?

packages/cli/src/cog/cutline.ts
```diff
@@ -131,21 +129,23 @@ export class Cutline {
...
         }
-        if (quadKeys.length == 1 && quadKeys[0] == '') {
-            // empty quakKey strings cause problems for filenames.
-            quadKeys = Array.from(this.tmsQk.coverTile()).map((t) => this.tmsQk.fromTile(t));
-        }
-        return quadKeys.sort(QuadKey.compareKeys);
+
+        return cogs
+            .map((t) => {
+                return Object.assign({ name: TileMatrixSet.tileToName(t) }, tms.tileToSourceBounds(t).toJson());
```
blacha, line 146: Can you `...` over `Object.assign`?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+     * @param tile
+     */
+    *coverTile(tile?: Tile): Generator<Tile, null, void> {
+        if (tile == null) {
+            yield* this.topLevelTiles();
+            return null;
+        }
+        const z = tile.z + 1;
+        const { matrixWidth, matrixHeight } = this.zooms[tile.z];
+        const { matrixWidth: pw, matrixHeight: ph } = this.zooms[z];
```
blacha, line 166: what is a pw? ph?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+
+        const xfact = pw / matrixWidth;
```
blacha, line 168: `xFact` What is it a factor of ? scale?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+        const yfact = ph / matrixHeight;
+
+        const xs = Math.floor(tile.x * xfact);
```
blacha, line 171: what is a xs?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+        const ys = Math.floor(tile.y * yfact);
+
+        const xe = Math.floor((tile.x + 1) * xfact);
```
blacha, line 174: what is a xe?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+        }
+
+        return null;
+    }
+
+    /**
+     * Make a name for a `tile`
+     */
+    public static tileToName(tile: Tile): string {
+        return `${tile.z}-${tile.x}-${tile.y}`;
```
blacha, line 190: this is different to a lot of places where we have `{x}_{y}_z{z}`

do we want to prefix the keys `z${z}_x${x}_y${y}`

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,89 @@ export class TileMatrixSet {
...
+    }
+
+    /**
+     * Convert `name` to a `tile`
+     */
+    public static nameToTile(name: string): Tile {
+        let i = 0;
+        const [z, x, y] = name.split('-').map((p) => {
+            if (++i > 3) {
```
blacha, line 199: this doesn't need to be inside the loop

```
const parts = name.split('-');
if (parts.length != 3) throw ...
const [z,x,y] = parts.map( ...
```

packages/shared/src/aws/tile.metadata.imagery.ts
```diff
@@ -33,19 +33,20 @@ export function compareImageSets(a: TileSetRuleImagery, b: TileSetRuleImagery):
...
 
-    public async get(imgId: string): Promise<TileMetadataImageryRecord> {
+    public async get(imgId: string): Promise<TileMetadataImageryRecordV1> {
         const existing = this.imagery.get(imgId);
         if (existing) return existing;
 
-        const item = await this.metadata.get<TileMetadataImageryRecord>(imgId);
+        const item = await this.metadata.get<TileMetadataImageryRecordV1>(imgId);
         if (item == null) throw new Error('Unable to find imagery: ' + imgId);
+        if (item.v == null) throw new Error('version 0 imagery no longer suppored');
```
blacha, line 49: We need to support v0 for one release then we can patch it out.

packages/shared/src/tms/__test__/test.util.ts
```diff
@@ -0,0 +1,16 @@
+import { NamedBounds } from '../../aws/tile.metadata.base';
+import { ProjectionTileMatrixSet } from '../projection.tile.matrix.set';
+import { EpsgCode, QuadKey } from '@basemaps/geo';
+
+export function qkToName(qk: string): string {
+    const tile = QuadKey.toTile(qk);
+    return `${tile.z}-${tile.x}-${tile.y}`;
```
blacha, line 7: could we reference the other function here?

packages/tiler/src/__test__/tiler.test.ts
```diff
@@ -44,7 +44,7 @@ o.spec('tiler.test', () => {
             getTileBounds() {
                 return { x: 0, y: 0, width: 512, height: 387 };
             },
-            tif: { source: { name: '313111000120111.tiff' } },
+            tif: { source: { name: '15-32295-20496.tiff' } },
```
blacha, line 47: do we use `tif` or `tiff`

packages/cli/src/cog/__test__/cog.test.ts
jacott: No longer padded

packages/shared/src/aws/tile.metadata.imagery.ts
jacott: @blacha We don't if we migrate db before using

packages/tiler/src/__test__/tiler.test.ts
jacott: Lets leave that question for another PR

packages/shared/src/aws/tile.metadata.imagery.ts
blacha: cant migrate since the old version doesnt support the new version. so downtime!

and if we did migrate we dont need the if.

packages/cli/src/cog/cutline.ts
```diff
@@ -21,11 +21,32 @@ function polysSameShape(a: Polygon, b: Polygon): boolean {
...
+    return a.x <= b.x && a.x + a.width >= b.x + b.width && a.y <= b.y && a.y + a.height >= b.y + b.height;
+}
+
+function addNonDupes(list: Tile[], addList: Tile[]): void {
+    const len = list.length;
+    loop: for (const add of addList) {
+        for (let i = 0; i < len; ++i) {
+            const curr = list[i];
+            if (curr.x == add.x && curr.y == add.y && curr.z == add.z) {
+                continue loop;
```
blacha, line 38: dont use goto's

packages/cli/src/cog/cutline.ts
```diff
@@ -21,11 +21,32 @@ function polysSameShape(a: Polygon, b: Polygon): boolean {
     return true;
 }
 
+function namedBounds(tms: TileMatrixSet, tile: Tile): NamedBounds {
+    return { name: TileMatrixSet.tileToName(tile), ...tms.tileToSourceBounds(tile).toJson() };
+}
+
+function contains(a: BoundingBox, b: BoundingBox): boolean {
+    return a.x <= b.x && a.x + a.width >= b.x + b.width && a.y <= b.y && a.y + a.height >= b.y + b.height;
```
blacha, line 29: is this not already part of Bounds? maybe static method?

packages/cli/src/cog/job.ts
```diff
@@ -1,4 +1,5 @@
 import { Epsg } from '@basemaps/geo';
+import { compareArea } from '@basemaps/geo/build/bounds';
```
blacha, line 2: does this need to be imported directly from the build folder?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,84 @@ export class TileMatrixSet {
...
+     * @param tile
+     */
+    *coverTile(tile?: Tile): Generator<Tile, null, void> {
+        if (tile == null) {
+            yield* this.topLevelTiles();
+            return null;
+        }
+        const z = tile.z + 1;
+        const { matrixWidth, matrixHeight } = this.zooms[tile.z];
+        const { matrixWidth: parentWidth, matrixHeight: parentHeight } = this.zooms[z];
```
blacha, line 166: would this be a child width? since its one more zoom level in?

packages/geo/src/tile.matrix.set.ts
```diff
@@ -131,4 +125,84 @@ export class TileMatrixSet {
...
+    /**
+     * Convert `name` to a `tile`
+     */
+    public static nameToTile(name: string): Tile {
+        const parts = name.split('-');
+        if (parts.length == 3) {
+            const z = Number(parts[0]);
+            const x = Number(parts[1]);
+            const y = Number(parts[2]);
+            if (z === z && y === y && x === x) {
```
blacha, line 202: This is a odd way of `isNaN` what does this track instead?

packages/shared/src/aws/tile.metadata.imagery.ts
jacott: Can now migrate before deployment https://github.com/linz/basemaps-team/pull/198

packages/cli/src/cog/cutline.ts
jacott: And you reasoning is?

blacha: we have 0 gotos in the codebase right now, this would be changing the convention of not using them.

packages/geo/src/tile.matrix.set.ts
blacha: Lets just move our convention to z-x-y

PR #779, opened 2020-06-19
feat: nztm landing page

blacha:


packages/landing/deploy.js
```diff
@@ -78,12 +60,14 @@ async function deploy() {
                 Key: fileName,
                 Body: fileData,
                 Metadata: { [HashKey]: hash },
-                ContentType: mime.lookup(fileName),
```
blacha, line 81: This adds `utf-8` encoding if needed

packages/landing/package.json
```diff
@@ -9,12 +9,21 @@
...
+    "ol": "^6.3.1",
+    "proj4": "^2.6.2"
   },
   "devDependencies": {
-    "mime-types": "^2.1.26"
-  }
+    "parcel": "^1.12.4"
+  },
+  "browserslist": [
+    "last 2 Chrome versions"
```
blacha, line 27: Possibly need to change this at some stage

packages/landing/src/map.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.updateUrlTimer = setTimeout(() => this.updateUrl(), 1000);
+    };
+
+    getLocation(): MapLocation {
+        const view = this.map.getView();
+        const center = view.getCenter();
+        if (center == null) throw new Error('Invalid Map location');
+        const [lon, lat] = proj
+            .transform(center, `EPSG:${this.config.projection}`, UrlProjection)
+            .map((c: number) => Math.floor(c * 10e5) / 10e5); // Limit to 5 decimal places (TODO is 5 enough)
```
jacott, line 95: Google Maps uses 7, I believe

packages/landing/src/map.ts
```diff
@@ -0,0 +1,107 @@
...
+/** Projection to use for the URL bar */
+const UrlProjection = 'EPSG:4326';
+
+/** Default center point if none provided */
+const DefaultCenter: Record<number, MapLocation> = {
+    [Epsg.Google.code]: { lat: -41.277848, lon: 174.7763921, zoom: 8 },
+    [Epsg.Nztm2000.code]: { lat: -41.277848, lon: 174.6763921, zoom: 2 },
+};
+
+const attributions = '<a href="http://data.linz.govt.nz">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
```
palmerj, line 21: Maybe for now this can be a link https://www.linz.govt.nz/data/licensing-and-using-data/attributing-elevation-or-aerial-imagery-data?

packages/landing/src/map.ts
palmerj: If it's just for centring the map 5dp (~1m) should be ok. 7dp will give you ~1cm precision.

packages/landing/src/global.ts
```diff
@@ -0,0 +1,12 @@
+import { Basemaps } from './map';
+
+declare global {
+    interface Window {
+        // Google analytics
+        dataLayer: any[];
+        gtag(...args: any[]): void;
```
jacott, line 7: This is not used

packages/landing/src/config.ts
```diff
@@ -0,0 +1,26 @@
...
+if (Config.GoogleAnalytics != null && typeof window != 'undefined') {
+    window.dataLayer = window.dataLayer || [];
+    function gtag(...args: any[]): void {
+        window.dataLayer.push(args);
+    }
+    gtag('js', new Date());
+    gtag('config', `${Config.GoogleAnalytics}`);
+
+    const script = document.createElement('script');
+    script.setAttribute('async', 'true');
```
jacott, line 23: The standard is to set this to an empty string https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#Syntax

packages/landing/src/map.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.map = new Map({
+            target: this.el,
+            view,
+            layers: [this.getLayer()],
+        });
+
+        this.map.addEventListener('postrender', this.postRender);
+    }
+
+    updateUrlTimer: unknown | null = null;
```
jacott, line 83: It would be better for this to be zero when no timer is running rather than switching data types between null and number

packages/landing/package.json
jacott: How about `['> 5%', 'last 2 versions']`

packages/landing/src/map.ts
blacha: I was meaning to check what google maps did, they use 7 maybe we should use 7 too, simple fix

packages/landing/src/map.ts
blacha: There is a snippet somewhere that vicki has made, just hadnt got around to adding it in yet. I believe it links to somewhere like that.

I took this snippet from a codepen I think you gave me

packages/landing/src/map.ts
blacha: 2 problems.

1. null vs 0, 0 would be a convention, to know that 0 is not a valid timer number (is it?) I would prefer to be explict and say that if its null then there is none.

2. there is a weird typing issue with timers in DOM vs Node, since we have both sets of types imported, hence why its set as a unknown.

packages/landing/src/map.ts
blacha: Switched to 7dp

packages/landing/src/map.ts
palmerj: > I took this snippet from a codepen I think you gave me

yeah that codepend snippet was a copy and paste for general LDS with some random basemaps text!

Should so something like

```html
<a href=" https://www.linz.govt.nz/data/licensing-and-using-data/attributing-elevation-or-aerial-imagery-data">Sourced from LINZ. CC BY 4.0</a>'
```

blacha: Cool, updated.

packages/landing/src/config.ts
blacha: Done,

PR #792, opened 2020-06-23
build(deps): aws-cdk 1.46.0

blacha:


blacha:
@jacott Trying something new here, I have removed `^` so that dependabot hopefully wont try and update these deps!

PR #793, opened 2020-06-23
build: switch to esbuild for bundling backend and frontend

blacha:


packages/shared/src/tms/proj.ts
```diff
@@ -0,0 +1,12 @@
+import * as projImport from 'proj4';
```
blacha, line 1: There is no shared lib that has proj in both frontend and backend code hence the copy pasta

packages/_infra/package.json
```diff
@@ -10,9 +10,7 @@
     "deploy:diff": "cdk diff || true",
     "deploy:deploy": "cdk deploy '*' -y --require-approval never"
   },
-  "dependencies": {
-    "git-rev-sync": "^2.0.0"
```
blacha, line 14: git-rev-sync has a weird `require('src/' + arg)` which makes it hard to bundle

packages/cli/src/cog/cutline.ts
```diff
@@ -2,18 +2,20 @@ import { Bounds, Epsg, GeoJson, Tile, TileMatrixSet } from '@basemaps/geo';
 import { compareName, FileOperator, NamedBounds, ProjectionTileMatrixSet } from '@basemaps/shared';
 import { FeatureCollection, Position } from 'geojson';
 import { basename } from 'path';
-import { intersection, MultiPolygon, Polygon, Ring, union } from 'polygon-clipping';
+import * as pc from 'polygon-clipping';
 import { CoveringFraction, MaxImagePixelWidth } from './constants';
 import { CogJob, SourceMetadata } from './types';
 
+const { intersection, union } = pc;
```
blacha, line 9: this doesnt actually work, it works if bundled but not if running through node.

packages/__tests__/src/assert.ts
```diff
@@ -1,4 +1,4 @@
-import * as o from 'ospec';
```
blacha, line 1: To get `esbuild` building the same as `ncc` I enabled `esModuleInteropt` which removes the need for import * from a lot of libs.

PR #810, opened 2020-06-24
feat(landing): track mouse position in debug mode

blacha:
image

### DIFF
--- a/packages/landing/src/debug.ts
+++ b/packages/landing/src/debug.ts
@@ -1,6 +1,15 @@
 import { Basemaps } from './map';
 import OSM from 'ol/source/OSM';
 import TileLayer from 'ol/layer/Tile';
+import { Epsg } from '@basemaps/geo';
+import * as proj from 'ol/proj.js';
+
+function round(max: number): (c: number) => number {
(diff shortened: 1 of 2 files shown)


packages/landing/src/debug.ts
```diff
@@ -92,9 +101,29 @@ export function addDebugLayer(bm: Basemaps): void {
...
+    bm.map.getViewport().addEventListener('mousemove', (e) => {
+        const px = bm.map.getEventPixel(e);
+        const coord = bm.map.getCoordinateFromPixel(px);
+
+        const wgs84 = proj
+            .transform(coord, bm.config.projection.toEpsgString(), Epsg.Wgs84.toEpsgString())
+            .map(round7)
+            .join(', ');
+        mouseWgs84El.value.innerHTML = wgs84;
+        mouseEl.value.innerHTML = coord.map(round3).join(', ');
```
jacott, line 121: `textContent` would be a lot faster than `innerHTML`

blacha: my simple benchmark shows this not to be the case

image

PR #813, opened 2020-06-25
feat(landing): api key generation and menu information 

blacha:


packages/landing/src/api.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (apiKey == null) return newApiKey();
+    if (!apiKey.startsWith('c')) return newApiKey();
+
+    try {
+        const keyCreatedAt = ulid.decodeTime(apiKey.slice(1).toUpperCase());
+        if (Date.now() - API_EXPIRE_TIME < keyCreatedAt) return apiKey;
+        return newApiKey();
+    } catch (e) {
+        return newApiKey();
+    }
```
jacott, line 26: Why the try catch need here? Please add comment to code

blacha: You dont want this ever to error, and we call into a third party library.

if a error occurs just generate a new key doesnt really matter about making more keys

blacha: also protects against if the apikey is not a ulid

PR #821, opened 2020-06-25
fix(tiler-sharp): handle empty images

jacott:
### Fixes

Some COGs may be created with no content. Do not process these images.


jacott:
Closing because fixed upstream

PR #827, opened 2020-06-26
feat(landing): copy link button

blacha:
image

onclick:

image

### DIFF
--- a/packages/landing/static/index.css
+++ b/packages/landing/static/index.css
@@ -221,3 +221,50 @@ button {
 .info-links li a:hover {
     border: none;
 }
+
+
+.lui-form-icon {
+    position: relative;
+}
+
+.lui-standard-form .lui-form-icon.lui-form-icon--left input:not([type=checkbox]):not([type=radio]),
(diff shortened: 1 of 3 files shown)


packages/landing/static/index.css
```diff
@@ -38,7 +38,7 @@ button {
 }
 
 #map .ol-zoom-in {
-    border-radius: 4px 4px 0 0;
+    border-radius: 3 0 3;
```
jacott, line 41: What's this? where are the units or `%`? why repeat the three but not the zero?

packages/landing/static/index.css
```diff
@@ -151,6 +151,7 @@ button {
     position: absolute;
     z-index: 1;
     top: 0;
+    right: 0;
     right: -348px;
```
jacott, line 155: Why set right twice?

packages/landing/static/index.css
blacha: ahh thats a bad copy pasta

PR #834, opened 2020-06-28
fix(cli): mitigate polygon intersection errors

jacott:
Use clipline to reduce polygon when we don't care about degenerate edges. When we do care call
intersection to remove the degenerate edges.

Store image bounds not polygons in SourceMetadata and also store in CogJob. This reduces
floating-point errors when combining source imagery bounds into composite polygons.

Note:

The Martinez polygon-clipping libraries can suffer from errors when lines align. Other intersection
libraries are intractable for the size of coastline polygons.  A solution is to call intersection
after first calling clipline.

### DIFF
--- a/packages/cli/src/cog/cog.vrt.ts
+++ b/packages/cli/src/cog/cog.vrt.ts
@@ -1,11 +1,10 @@
-import { FileOperator, LogType, isConfigS3Role, Aws } from '@basemaps/shared';
-import { FeatureCollection } from 'geojson';
+import { Epsg } from '@basemaps/geo';
+import { Aws, FileOperator, isConfigS3Role, LogType } from '@basemaps/shared';
+import { GdalCogBuilder } from '../gdal/gdal';
+import { GdalCommand } from '../gdal/gdal.command';
+import { onProgress } from './cog';
 import { Cutline } from './cutline';
 import { CogJob } from './types';
-import { onProgress } from './cog';
-import { GdalCogBuilder } from '../gdal/gdal';
-import { Epsg } from '@basemaps/geo';
-import { GdalCommand } from '../gdal/gdal.command';
 
 /**
  * Build the VRT for the needed source imagery
@@ -28,14 +27,14 @@ async function buildPlainVrt(
         gdalCommand.setCredentials(credentials);
     } else {
         if (gdalCommand.mount) {
-            for (const file of job.source.files) {
-                gdalCommand.mount(file);
+            for (const { name } of job.source.files) {
+                gdalCommand.mount(name);
             }
         }
     }
 
     logger.debug({ buildOpts: buildOpts.join(' ') }, 'gdalbuildvrt');
-    const sourceFiles = job.source.files.map((c) => c.replace('s3://', '/vsis3/'));
+    const sourceFiles = job.source.files.map(({ name }) => name.replace('s3://', '/vsis3/'));
     await gdalCommand.run('gdalbuildvrt', [...buildOpts, vrtPath, ...sourceFiles], logger);
 }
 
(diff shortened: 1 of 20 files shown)


packages/cli/src/cog/cog.vrt.ts
```diff
@@ -102,7 +99,7 @@ export const CogVrt = {
 
         const inputTotal = job.source.files.length;
 
-        cutline.filterSourcesForName(name, job, sourceGeo);
+        cutline.filterSourcesForName(name, job);
```
blacha, line 102: possibly unrelated to this PR
but do we really need to screw with the `job.json` could we maybe `const filesToUse = cutline.filter...`

Im not a huge fan of functions destroying a object that is commonly used in a bunch of places.

packages/cli/src/cog/cutline.ts
```diff
@@ -98,38 +94,36 @@ export class Cutline {
      * @param job
      * @param  sourceGeo
      */
-    filterSourcesForName(name: string, job: CogJob, sourceGeo: FeatureCollection): void {
+    filterSourcesForName(name: string, job: CogJob): void {
         const tile = TileMatrixSet.nameToTile(name);
-        const qkBounds = this.targetProj.tileToSourceBounds(tile);
+        const { targetProj } = this;
+        const qkBounds = targetProj.tileToSourceBounds(tile);
```
blacha, line 100: NIT: these are no longer quadkeys

packages/geo/src/bounds.ts
```diff
@@ -153,6 +168,30 @@ export class Bounds implements BoundingBox {
...
     }
 
+    /**
+     * Find the bounds of a GeoJson MultiPolygon
+
+     * @param multipoly the polygon to measure
+     * @throws Error if `multipoly` is empty
+     */
+    public static fromMultiPolygon(multipoly: number[][][][]): Bounds {
+        if (multipoly.length == 0) return new Bounds(0, 0, 0, 0);
```
blacha, line 178: Comments say we throw here?

packages/geo/src/bounds.ts
```diff
@@ -153,6 +168,30 @@ export class Bounds implements BoundingBox {
...
+        let minX = multipoly[0][0][0][0];
+        let minY = multipoly[0][0][0][1];
+        let maxX = minX;
+        let maxY = minY;
+        for (const [poly] of multipoly) {
```
blacha, line 183: do we have any concerns about polygons wrapping the wrong way if the polygon crosses the antimeridaian?

packages/shared/src/tms/projection.tile.matrix.set.ts
```diff
@@ -64,6 +64,20 @@ export class ProjectionTileMatrixSet {
         return (epsgCode && CodeMap.get(epsgCode)) ?? null;
     }
 
+    /**
+     * Project the points in a MultiPolygon array to the `targetProjection`. Modifies `multipoly`.
+     * Does not support holes in polygons.
+     */
+    projectMultipolygon(multipoly: number[][][][], targetProjection: ProjectionTileMatrixSet): void {
```
blacha, line 71: return a new polygon instead of destroying the current one

packages/cli/src/cog/cog.vrt.ts
jacott: I'll add to my TODO list

packages/geo/src/bounds.ts
jacott: No because we are in source units not WSG84

packages/shared/src/tms/projection.tile.matrix.set.ts
jacott: Yeah, I knew you would say that. It just means I have to write more code because I have to copy one that are in same projection because otherwise I break the contract; I hate wasting CPU cycles also.

packages/cli/src/cog/cog.vrt.ts
blacha: Cool doesnt have to hold this PR up

packages/shared/src/tms/projection.tile.matrix.set.ts
blacha: Hmm, could change it so if projection source == target throw? we shouldnt be projecting to the same thing?

in terms of lines, its pretty silimar with 

```
const output = [];
      for (const poly of multipoly) {
            output.push(poly[0].map((p) => fromWsg84(toWsg84(p))));
        }
return output;
```

and you are probably wasting more cpu cycles by using a `.map()` inside of a loop and creating a  function inside the loop too.

jacott: Now returns same multipoly if same projection otherwise projects to new projection

PR #837, opened 2020-06-29
chore(github): Get git version and hash using git command

jacott:


packages/shared/src/cli/__test__/git.tag.test.ts
```diff
@@ -0,0 +1,9 @@
+import o from 'ospec';
+import { GitTag } from '../git.tag';
+
+o.spec('git.tag', () => {
+    o('format', () => {
+        o(/^v\d+\.\d+\.\d+(-\d+-g[0-9a-f]+)?$/.test(GitTag.version)).equals(true);
+        o(/^[0-9a-f]+$/.test(GitTag.hash)).equals(true);
```
blacha, line 7: when git moves to sha2 hashes i think this is going to fail.

packages/shared/src/cli/git.tag.ts
```diff
@@ -0,0 +1,6 @@
+import { execFileSync } from 'child_process';
+
+export const GitTag = {
```
blacha, line 3: want to make this a function so we dont execute files every time this script is imported, this allows us to export it inside of the main index, and then you can import it from the index rather than via /build

```
export function GitTag() { return ... }
```

packages/shared/src/cli/__test__/git.tag.test.ts
jacott: It's a test so we'll know right away and it would be good to get a "heads up" about it

packages/shared/src/cli/git.tag.ts
```diff
@@ -0,0 +1,6 @@
...
+    version: execFileSync('git', ['describe', '--tags', '--always', '--match', 'v*']).toString().trim(),
+    hash: execFileSync('git', ['rev-parse', 'HEAD']).toString().trim(),
```
blacha, line 5: any idea what this returns inside of CI ?

packages/shared/src/cli/git.tag.ts
jacott: But it is a sync command so executing it every time is not nice and we don't want it in the index because it may not be valid once deployed

packages/shared/src/cli/git.tag.ts
jacott: The build passes so I assume it is a valid version and hash

packages/shared/src/cli/git.tag.ts
blacha: exactly you are waiting on two sub processes to run even if you dont even use these values.

the infra version info already has a caching mechanism, 

The cli in production will never want it never to run, and then adds a dep on git for our container.

packages/shared/src/cli/git.tag.ts
blacha: Is it the expected value?

packages/shared/src/cli/git.tag.ts
jacott: You are only waiting on this if you *do* use this process otherwise the module would not import the version module or this module directly

packages/shared/src/cli/git.tag.ts
jacott: I assume so because the test passes.

PR #842, opened 2020-06-30
fix(cli): ensure fatal errors set process exit code to 1

jacott:
### Fixes

`cogify cog` was detecting `job.json` version mismatch but was exiting the process with code 0 (success); now exits with code 1.

### DIFF
--- a/packages/shared/src/logger.fatal.error.ts
+++ b/packages/shared/src/logger.fatal.error.ts
@@ -1,0 +1,10 @@
+/**
+ * Utility error to throw expecting that Logger will fatal log its contents
+ */
+export class LoggerFatalError extends Error {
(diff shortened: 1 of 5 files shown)


packages/shared/src/logger.fatal.error.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Utility error to throw expecting that Logger will fatal log its contents
+ */
+export class LoggerFatalError extends Error {
+    obj: object;
```
blacha, line 5: For some reason our linter says to use `Record<string, unknown>` rather than `Object` I dont know what a `object` is in terms of typing, I cant ctrl click it to goto its defs

Not too sure the reasoning. but make it a `Record<>`

PR #852, opened 2020-07-02
chore(ci): use npm-audit-resolver

jacott:
When audit issues are not fixed in a timely manner allow change to ignore issues for a time period.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -22,7 +22,7 @@ jobs:
         run: yarn
 
       - name: Audit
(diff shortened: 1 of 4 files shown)


.github/workflows/push.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         run: yarn
 
       - name: Audit
-        run: yarn audit
+        run: ./node_modules/.bin/check-audit --yarn
```
blacha, line 25: `npx check-audit`

PR #853, opened 2020-07-03
feat(landing): report tile loading stats

blacha:
record how long tiles take to load for users

### DIFF
--- a/packages/landing/src/map.ts
+++ b/packages/landing/src/map.ts
@@ -10,7 +10,6 @@ import { NztmOl } from './nztm2000';
 import TileSource from 'ol/source/Tile';
 import BaseEvent from 'ol/events/Event';
 import { ImageTile } from 'ol';
-import { count } from 'console';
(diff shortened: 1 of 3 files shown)


packages/landing/src/map.ts
```diff
@@ -68,15 +86,55 @@ export class Basemaps {
...
+                this.tileTrackTimer = setTimeout(this.reportTileStats, TileReportTimeDurationMs);
+            }
+        }
+    };
+
+    reportTileStats = (): void => {
+        this.tileTrackTimer = null;
+        const tileLoadTimes = this.tileLoadTimes;
+        if (tileLoadTimes.length < TileMinReportCount) return;
+        this.tileLoadTimes = [];
```
jacott, line 127: Invalid reference to `this` from within a function that is called without binding `this` to it.

blacha: this is a `() => {}` defined so `this` is bound at constructor time.

PR #858, opened 2020-07-05
fix(landing): allow firefox to render webps if it supports it

blacha:
Firefox does not support the creation of webp from canvas, but supports rendering it


packages/landing/src/index.ts
```diff
@@ -4,24 +4,12 @@ import { Basemaps } from './map';
...
-    // very old browser like IE 8, canvas not supported
-    return false;
-}
-
-document.addEventListener('DOMContentLoaded', () => {
+document.addEventListener('DOMContentLoaded', async () => {
     const mapEl = document.getElementById('map');
     if (mapEl == null) throw new Error('Cannot find #map element');
-    if (canUseWebP()) WindowUrl.ImageFormat = 'webp';
+    if (await isWebpSupported) WindowUrl.ImageFormat = 'webp';
```
jacott, line 13: Shouldn't this be a function call?

blacha: Its currently just a Promise, as you dont need to wait for `DOMContentLoaded` to start testing, I was wondering if we should wait though...

so when you import the file, it tries to render a webp image, rendering looks to take 20+ms so we could add it to DOMContentLoaded event

jacott: Ignore that

PR #863, opened 2020-07-06
fix(cli): make clipMultipolygon always remove degenerate edges

jacott:
Some unnecessary cogs were being created because intersecting bounds but not intersecting polygons
was resulting in degenerate-edges-only-polygons which were not being interpreted as empty. By always
removing the degenerate edges we loose some performance but get the correct results.

We could restore most of the performance with some extra logic but for simplicity's sake it is not worth it.


jacott:
Fixes https://github.com/linz/basemaps-team/issues/239

PR #894, opened 2020-07-09
fix(wmts): add style tag to wmtscaps

josephramsay:
Fixes: #
Added Style tag to Layer/Content setting title/id as default

### DIFF
--- a/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
+++ b/packages/lambda-xyz/src/__test__/wmts.capability.test.ts
@@ -58,6 +58,13 @@ o.spec('WmtsCapabilities', () => {
         o(layer?.find('ows:Title')?.textContent).equals('aerial:title');
         o(layer?.find('ows:Identifier')?.textContent).equals('aerial-3857');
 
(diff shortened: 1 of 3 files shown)


packages/lambda-xyz/src/__test__/wmts.capability.test.ts
```diff
@@ -58,6 +58,11 @@ o.spec('WmtsCapabilities', () => {
         o(layer?.find('ows:Title')?.textContent).equals('aerial:title');
         o(layer?.find('ows:Identifier')?.textContent).equals('aerial-3857');
 
+        o(layer?.find('Style')?.textContent).equals(
+            V('Style', { isDefault: 'true' }, [V('ows:Title', 'Default Style'), V('ows:Identifier', 'default')])
+                .textContent,
+        );
+
```
jacott, line 65: `textContent` does Not test that the element names are correct. I suggest using `toString`

josephramsay: :+1: Okay. I did consider that but saw the prior test was using textContent

PR #913, opened 2020-07-13
fix(tiler): try to minimize the error when rounding boundaries

jacott:
Overview tiles that do not align with the requested zoom level often leave multiple pixel gaps
between COGs. This fix reduces that size of that gap.

### DIFF
--- a/packages/geo/src/bounds.ts
+++ b/packages/geo/src/bounds.ts
@@ -156,8 +156,13 @@ export class Bounds implements BoundingBox {
         );
     }
 
+    /**
+     * Round dimensions to integers keeping the error a low as possible
+     */
     public round(): Bounds {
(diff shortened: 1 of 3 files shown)


packages/geo/src/bounds.ts
```diff
@@ -156,8 +156,13 @@ export class Bounds implements BoundingBox {
         );
     }
 
+    /**
+     * Round dimensions to integers keeping the error a low as possible
+     */
     public round(): Bounds {
-        return new Bounds(Math.round(this.x), Math.round(this.y), Math.round(this.width), Math.round(this.height));
+        const x = Math.round(this.x),
```
blacha, line 163: prefer 
```
const x =
const y =
```

PR #920, opened 2020-07-13
feat(cli): Allow creation of one cog covering entire extent

jacott:
This is for use with creating the gebco COG because the default process can't handle the
anti-meridian and one cog is optimal for the gebco layer.

### DIFF
--- a/packages/cli/src/cli/cogify/action.job.ts
+++ b/packages/cli/src/cli/cogify/action.job.ts
@@ -7,7 +7,7 @@ import {
     CommandLineStringParameter,
 } from '@rushstack/ts-command-line';
 import { CogJobFactory, JobCreationContext, MaxConcurrencyDefault } from '../../cog/job';
-import { GdalCogBuilderDefaults } from '../../gdal/gdal.config';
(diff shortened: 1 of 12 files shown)


packages/cli/src/cli/cogify/action.job.ts
```diff
@@ -89,6 +91,24 @@ export class ActionJobCreate extends CommandLineAction {
         const targetProjection = ProjectionTileMatrixSet.tryGet(this.targetProjection?.value);
         if (targetProjection == null) throw new Error('Invalid target-projection');
 
+        let resampling:
+            | {
```
blacha, line 95: Extract this into a named type, since we use it in a few places

packages/geo/src/tile.matrix.set.ts
```diff
@@ -29,6 +29,8 @@ export class TileMatrixSet {
     /** Array index of y coordinate */
     indexY = 1;
 
+    extent: Bounds;
```
blacha, line 32: NIT brief comment maybe?

PR #929, opened 2020-07-15
fix(cli): fix creating cutlines with self crossing edges

jacott:
Two problems:

1. Union of source polygons sometimes not removing common edges of adjacent images.
2. Storing resulting polygon in WSG84 sometimes causes close edges to cross.

Solution to 1 is to expand the source imagery bounds by a small fraction.

Solution to 2 is to store geojson cutline in target projection with a CRS.


packages/cli/src/cog/cutline.ts
```diff
@@ -11,6 +11,12 @@ const { intersection, union } = pc;
 /** Padding to always apply to image boundies */
 const PixelPadding = 100;
 
+const SourceSmoothScale = 1 + 1e-8;
+
+function copyPoint(p: number[]): [number, number] {
+    return [p[0], p[1]];
```
blacha, line 17: do you expect points to have more than two values in it ? aka why this over `[].slice()`

jacott: faster.  2ns vs 6ns

packages/cli/src/cog/cutline.ts
```diff
@@ -188,8 +197,12 @@ export class Cutline {
      * Convert JobCutline to geojson FeatureCollection
      */
     toGeoJson(clipPoly = this.clipPoly): FeatureCollection {
-        const { toWsg84 } = this.targetPtms.proj;
-        return GeoJson.toFeatureCollection([GeoJson.toFeatureMultiPolygon(clipPoly.map((p) => [p[0].map(toWsg84)]))]);
+        const feature = GeoJson.toFeatureCollection([GeoJson.toFeatureMultiPolygon(clipPoly)]);
```
blacha, line 200: we should be avoiding `as any`

how about 
```
export type FeatureCollectionCrs = FeatureCollection & { crs: { type: 'name', properties: { name: string } }}

toGeoJson(...) : FeatureCollectionCrs
```

PR #943, opened 2020-07-20
fix(lambda-api): track api key usage

blacha:


packages/lambda-api-tracker/src/__test__/xyz.request.test.ts
```diff
@@ -77,34 +77,4 @@ o.spec('xyz-request', () => {
             'x-linz-request-id': [{ key: 'x-linz-request-id', value: String(res.header(HttpHeader.RequestId)) }],
         });
     });
-
-    o('should not cache WMTSCapabilities', async () => {
```
blacha, line 81: Adding cache control on the inwards request makes no sense, as this is not served to the client.
```
Client -> CloudFront -> XyzLambda -> TileLambda -> CloudFront -> Client
```

packages/lambda-api-tracker/src/index.ts
```diff
@@ -1,60 +1,45 @@
...
+    const res = await ValidateRequest.validate(req);
     req.timer.end('validate');
 
-    if (res != null) {
-        return res;
-    }
+    if (res != null) return res;
 
-    const response = new LambdaHttpResponse(100, 'Continue');
-    // Api key will be trimmed from the forwarded request so pass it via a well known header
```
blacha, line 54: These are now being proxied through to the tiler lambda

packages/lambda-api-tracker/src/__test__/validate.test.ts
```diff
@@ -42,15 +43,16 @@ o.spec('ApiValidate', (): void => {
...
+    //     const result = await ValidateRequest.validate(makeContext(dummyApiKey));
+    //     o(result).notEquals(null);
+    //     if (result == null) throw new Error('Validate returns null result');
 
-        o(result.status).equals(403);
-        o(result.statusDescription).equals('Invalid API Key');
-    });
+    //     o(result.status).equals(403);
+    //     o(result.statusDescription).equals('Invalid API Key');
+    // });
```
jacott, line 55: Any reason why can't re-enable it now?

packages/lambda-api-tracker/src/__test__/xyz.request.test.ts
jacott: Are we setting it on the outwards response?

packages/lambda-api-tracker/src/__test__/validate.test.ts
blacha: this check was to validate it exists in the database, we could add another check that validates the format. Ill make a separate pr for that

PR #953, opened 2020-07-21
feat(lambda-xyz): Serve WMTSCapabilities for all TileSets

jacott:
When the tile set name is not supplied create a WMTSCapabilities xml file with all supported Layers
including individual layers from composite tileSets.

When the projection is not supplied create a WMTSCapabilities xml file with layers for all supported
target projections.

Allow fetching sub imagery tiles by the parent imagery name plus the child imagery set name. For
example `/v1/tiles/aerial@beta:tasman_rural_2018-19_0-3m/2193/8/405/321.png`

Group projections for same imagery.

Set the `ows:BoundingBox` for each layer to the extent of the layer

Set the `ows:WGS84BoundingBox` also (after the `ows:BoundingBox`). Currently if the bounding box
crosses the anti-merdian just make the boundingbox the world extent. It may be possible to instead
create two bounding boxes on each side of the anti-merdian.

### DIFF
--- a/packages/__tests__/src/rounding.ts
+++ b/packages/__tests__/src/rounding.ts
@@ -33,3 +33,8 @@ export function round(thing: any, z = 8): any {
     };
     return recurse(thing);
 }
+
+export function roundString(str: string, z = 8): string {
+    const r = makeRound(z);
+    return str.replace(/\d+\.\d+/g, (n) => r(parseFloat(n)).toString());
+}
--- a/packages/geo/src/wmts/wmts.ts
+++ b/packages/geo/src/wmts/wmts.ts
@@ -1,4 +1,5 @@
 import type { Epsg } from '../epsg';
+import { Bounds } from '../bounds';
 
 /**
  * Cut down interface, with only the information needed to render a tileset
@@ -9,6 +10,7 @@ export interface WmtsLayer {
     description: string;
     projection: Epsg;
     taggedName: string;
+    extent: Bounds;
 }
 
 /** WMTS Provider information */
--- a/packages/lambda-xyz/src/cli/serve.ts
+++ b/packages/lambda-xyz/src/cli/serve.ts
@@ -45,10 +45,27 @@ async function handleRequest(
     }
 }
 
-async function main(): Promise<void> {
-    if (Env.get(Env.PublicUrlBase) == '') {
-        process.env[Env.PublicUrlBase] = `http://localhost:${port}`;
-    }
+function useAws(): void {
+    app.get('/v1/*', async (req: express.Request, res: express.Response) => {
+        const startTime = Date.now();
+        const requestId = ulid.ulid();
+        const logger = LogConfig.get().child({ id: requestId });
+        const ctx = new LambdaContext(
+            {
+                httpMethod: 'get',
+                path: req.path,
+                queryStringParameters: req.query,
+            } as any,
+            logger,
+        );
+
+        await handleRequest(ctx, res, startTime, logger);
+    });
+
+    LogConfig.get().info({ port, base: process.env[Env.PublicUrlBase], aws: process.env['AWS_PROFILE'] }, 'Listen');
+}
+
+async function useLocal(): Promise<void> {
     let projection: number = Epsg.Google.code;
     const filePath = process.argv[2];
 
@@ -93,12 +110,24 @@ async function main(): Promise<void> {
     });
 
     app.use(express.static(__dirname + '/../../../landing/dist/'));
-    await new Promise((resolve) => app.listen(port, resolve));
 
     const url = Env.get(Env.PublicUrlBase) + `/?i=${tileSetName}&p=${projection}`;
     const wmts = Env.get(Env.PublicUrlBase) + `/v1/WMTSCapabilities.xml`;
     const xyz = Env.get(Env.PublicUrlBase) + `/v1/tiles/${tileSetName}/${projection}/{z}/{x}/{y}.png`;
     LogConfig.get().info({ url, wmts, xyz }, 'Listen');
+}
+
+async function main(): Promise<void> {
+    if (Env.get(Env.PublicUrlBase) == '') {
+        process.env[Env.PublicUrlBase] = `http://localhost:${port}`;
+    }
+
+    if (process.argv.length < 3) {
+        useAws();
+    } else {
+        await useLocal();
+    }
+    await new Promise((resolve) => app.listen(port, resolve));
 }
 
 main().catch((e) => console.error(e));
--- a/packages/lambda-xyz/src/routes/tile.ts
+++ b/packages/lambda-xyz/src/routes/tile.ts
@@ -1,14 +1,14 @@
-import { Epsg, TileMatrixSet, Tile } from '@basemaps/geo';
+import { Tile, TileMatrixSet } from '@basemaps/geo';
 import { HttpHeader, LambdaContext, LambdaHttpResponse } from '@basemaps/lambda';
 import {
     Aws,
     Env,
+    ProjectionTileMatrixSet,
     TileDataWmts,
     TileDataXyz,
     tileFromPath,
     TileMetadataTag,
     TileType,
-    ProjectionTileMatrixSet,
 } from '@basemaps/shared';
 import { TileMakerSharp } from '@basemaps/tiler-sharp';
 import { CogTiff } from '@cogeotiff/core';
@@ -16,7 +16,7 @@ import { createHash } from 'crypto';
 import pLimit from 'p-limit';
 import { EmptyPng } from '../png';
 import { TileSet } from '../tile.set';
-import { loadTileSet } from '../tile.set.cache';
+import { loadTileSet, loadTileSets } from '../tile.set.cache';
 import { Tilers } from '../tiler';
 import { WmtsCapabilities } from '../wmts.capability';
 import { TileEtag } from './tile.etag';
@@ -39,6 +39,8 @@ const LoadingQueue = pLimit(Env.getNumber(Env.TiffConcurrency, 5));
 
 /** Background color of tiles where the tileset does not define a color */
 const DefaultBackground = { r: 0, g: 0, b: 0, alpha: 0 };
+
+const NotFound = new LambdaHttpResponse(404, 'Not Found');
 
 /** Initialize the tiffs before reading */
 async function initTiffs(tileSet: TileSet, tms: TileMatrixSet, tile: Tile, ctx: LambdaContext): Promise<CogTiff[]> {
@@ -86,7 +88,9 @@ export async function tile(req: LambdaContext, xyzData: TileDataXyz): Promise<La
     const latLon = ProjectionTileMatrixSet.get(xyzData.projection.code).tileCenterToLatLon(xyzData);
     req.set('location', latLon);
 
-    const tileSet = await loadTileSet(req, xyzData.name, xyzData.projection);
+    req.timer.start('tileset:load');
+    const tileSet = await loadTileSet(xyzData.name, xyzData.projection);
+    req.timer.end('tileset:load');
     if (tileSet == null) return new LambdaHttpResponse(404, 'Tileset Not Found');
 
     const tiffs = await initTiffs(tileSet, tiler.tms, xyzData, req);
@@ -132,13 +136,16 @@ export async function wmts(req: LambdaContext, wmtsData: TileDataWmts): Promise<
 
     const host = Env.get(Env.PublicUrlBase) ?? '';
 
-    // TODO when we support more than one projection
-    const tileSet = await loadTileSet(req, wmtsData.name, wmtsData.projection ?? Epsg.Google);
-    const provider = await Aws.tileMetadata.Provider.get(TileMetadataTag.Production);
-    if (tileSet == null || provider == null) return new LambdaHttpResponse(404, 'Not Found');
+    req.timer.start('tileset:load');
+    const tileSets = await loadTileSets(wmtsData.name, wmtsData.projection);
+    req.timer.end('tileset:load');
+    if (tileSets.length == 0) return NotFound;
 
-    const xml = WmtsCapabilities.toXml(host, provider, [tileSet], req.apiKey);
-    if (xml == null) return new LambdaHttpResponse(404, 'Not Found');
+    const provider = await Aws.tileMetadata.Provider.get(TileMetadataTag.Production);
+    if (provider == null) return NotFound;
+
+    const xml = WmtsCapabilities.toXml(host, provider, tileSets, req.apiKey);
+    if (xml == null) return NotFound;
 
     const data = Buffer.from(xml);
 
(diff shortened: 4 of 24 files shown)


packages/geo/src/epsg.ts
```diff
@@ -32,6 +32,7 @@ const EPSGTextMap: Record<string, EpsgCode> = {
 
 export class Epsg {
     static Codes = new Map<number, Epsg>();
+    static TargetCodes = [EpsgCode.Google, EpsgCode.Nztm2000];
```
blacha, line 35: What is a TargetCodes?

are these the only output codes that we are using? Not sure if this should be placed here as its somewhat specific to our implmentation of cogification.

packages/lambda-xyz/src/routes/tile.ts
```diff
@@ -40,6 +40,10 @@ const LoadingQueue = pLimit(Env.getNumber(Env.TiffConcurrency, 5));
 /** Background color of tiles where the tileset does not define a color */
 const DefaultBackground = { r: 0, g: 0, b: 0, alpha: 0 };
 
+function notFound(): LambdaHttpResponse {
+    return new LambdaHttpResponse(404, 'Not Found');
```
blacha, line 44: is there anything stopping us doing
```
const NotFound = new LambdaHttpResponse(404, 'Not Found');
```

I dont think we need new instance of this every time.

packages/lambda-xyz/src/routes/tile.ts
```diff
@@ -84,7 +88,11 @@ export async function tile(req: LambdaContext, xyzData: TileDataXyz): Promise<La
     const latLon = ProjectionTileMatrixSet.get(xyzData.projection.code).tileCenterToLatLon(xyzData);
     req.set('location', latLon);
 
-    const tileSet = await loadTileSet(req, xyzData.name, xyzData.projection);
+    req.set('tileSet', xyzData.name);
+    req.set('projection', xyzData.projection);
```
blacha, line 92: I believe my previous PR added this info into the log lines here.

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,88 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+        // single tileSet
+        const ts = await loadTileSet(name, projection);
+
+        return ts == null ? [] : [ts];
+    }
+
+    const tileSets: TileSet[] = [];
+
+    if (name === '') {
+        for (const tsn in TileSetName) {
```
blacha, line 88: `for in` is dangerous with enums, as sometimes they will be doubly linked eg here is the generated source for EpsgCode enum

```
    {
  '2193': 'Nztm2000',
  '3793': 'Citm2000',
  '3857': 'Google',
  '4326': 'Wgs84',
  Google: 3857,
  Wgs84: 4326,
  Nztm2000: 2193,
  Citm2000: 3793
}
```

packages/shared/src/api.path.ts
```diff
@@ -29,9 +29,17 @@ export interface TileDataWmts {
     projection: Epsg | null;
 }
 
+function parseTargetEpsg(text: string): Epsg | null {
+    const projection = Epsg.parse(text);
+    if (projection == null || !Epsg.TargetCodes.find((c) => c === projection.code)) {
```
blacha, line 34: could save on a function creation with `Epsg.TargetCodes.includes(projection.code)`

packages/lambda-xyz/src/routes/tile.ts
jacott: The DRYing up no longer applies because WMTS and tile rendering now take different paths

packages/geo/src/epsg.ts
jacott: Moved to `ProjectionTileMatrixSet`

packages/lambda-xyz/src/routes/tile.ts
jacott: Moved to `TileOrWmts`

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,104 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
     }
     return tileSet;
 }
+
+function titleizeName(name: string): string {
+    return name.replace(/_0-/g, ' 0.').replace(/_/g, ' ');
+}
+
+function individualTileSet(parent: TileSet, image: TileSetRuleImagery): TileSet {
+    const id = image.imagery.id.slice(3);
```
blacha, line 31: can we use `unprefix(image.imagery.id)` here? which explains why we are trimming a id by 3 characters.

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,104 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+        id,
+        minZoom: 0,
+        maxZoom: image.rule.maxZoom,
+        priority: 0,
+    };
+    // Override imagery (for completeness)
+    tileSet.imagery = { [id]: rule };
+    copy.tileSet = tileSet;
+
+    copy.imagery = [
```
blacha, line 51: NIT: you should be able to put this on one line?

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,104 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+}
+
+/**
+ * Load a collection of TileSets. Used by WMTSCapabilities
+
+ * @param name if an empty string load all TileSets
+ * @param projection if null load all projections
+ */
+export async function loadTileSets(nameStr: string, projection: Epsg | null): Promise<TileSet[]> {
+    if (nameStr !== '' && nameStr[0] !== '@' && projection != null) {
```
blacha, line 92: why can a nameStr startwith '@' ?

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,104 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+    // use parent data as prototype for child;
+    copy.tileSet = Object.create(parent.tileSet ?? null);
+
+    const { tileSet } = copy;
+
+    copy.titleOverride = `${parent.title} ${titleizeName(image.imagery.name)}`;
+    copy.extentOverride = Bounds.fromJson(image.imagery.bounds).toBbox();
+
+    const rule = {
+        id,
```
blacha, line 42: are these meant to be unprefixed keys?

packages/geo/src/index.ts
```diff
@@ -4,3 +4,4 @@ export { GeoJson } from './geo.json';
 export { Epsg, EpsgCode } from './epsg';
 export { TileMatrixSet, Tile } from './tile.matrix.set';
 export { WmtsLayer, WmtsProvider } from './wmts/wmts';
+export { TileSetName, TileSetNameValues } from './tile.set.name';
```
blacha, line 7: This is also quite specific to our process, is there another place this could live?

maybe something like `shared` ?

packages/geo/src/wmts/wmts.ts
```diff
@@ -9,6 +9,7 @@ export interface WmtsLayer {
     description: string;
     projection: Epsg;
     taggedName: string;
+    extent: [number, number, number, number];
```
blacha, line 12: what format is this extent in? `minX minY maxX maxY`?

packages/lambda-xyz/src/tile.set.cache.ts
jacott: Do that you can request all layers for a tagged tileSet for example: `http://localhost:5050/v1/tiles/@head/WMTSCapabilities.xml`

packages/lambda-xyz/src/cli/serve.ts
```diff
@@ -49,56 +49,82 @@ async function main(): Promise<void> {
...
-        const tileSet = new TileSetLocal(tileSetName, Epsg.Google, filePath);
-        await tileSet.load();
-        const tiffFiles = tileSet.tiffs.map((c) => c.source.name).join(', ');
-        // TODO is there a better name for this
-        tileSet.setTitle(`Local - ${tiffFiles}`);
-        TileSets.set(tileSet.id, tileSet);
-        LogConfig.get().info({ tileSets: [...TileSets.keys()] }, 'LoadedTileSets');
-        projection = tileSet.projection.code;
-    }
+    if (process.argv[2] === '--use-dynamodb') {
```
blacha, line 53: I don't really like the idea of having this large `if` here.

is there a way to use dynamodb by default and just override what is being returned by dynamodb when a folder is provided

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +9,104 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+        {
+            rule,
+            imagery: image.imagery,
+        },
+    ];
+
+    return copy;
+}
+
+async function collectTileSets(tileSets: TileSet[], name: string, projection: Epsg | null): Promise<void> {
```
blacha, line 61: ideally we should not be modifying stuff we pass into functions, is there a reason we need to "collect" rather than just return the list of tile sets?

eg 
```
getTileSets(name: string, projection?: Epsg): Promise<TileSet[]>
```

blacha: also do we ever dedupe this?

jacott: When it is a private function then I see no problem with collecting the in the list; the name says what it's doing.

jacott: dedupe is not required here because a tileSet does not contain duplicates.

packages/__tests__/src/rounding.ts
```diff
@@ -33,3 +33,8 @@ export function round(thing: any, z = 8): any {
     };
     return recurse(thing);
 }
+
+export function roundString(str: string, z = 8): string {
+    const r = makeRound(z);
+    return str.replace(/\d+\.\d+/g, (n) => r(parseFloat(n)).toString());
```
blacha, line 39: regexp for number matching is hard. eg `-` vs `0x10` vs `1e10` `5n`

would be safer to split on a known eg `foo.split(` `).map(r).join(' ')`

packages/shared/src/proj/projection.ts
```diff
@@ -73,38 +73,47 @@ export class Projection {
...
         return this.projection.inverse;
     }
 
+    boundsToWgs84(source: Bounds): Bounds {
+        const sw = this.toWgs84([source.x, source.y]);
+        const ne = this.toWgs84([source.right, source.bottom]);
+
+        const width = ne[0] - sw[0];
+
+        return new Bounds(sw[0], sw[1], width < 0 ? width + 360 : width, ne[1] - sw[1]);
```
blacha, line 102: why normalize only one side?

should also be done inside of a while loop to properly normalize them

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +16,115 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
-    req.timer.end('tileset:load');
     if (!loaded) {
         TileSets.delete(tileSet.id);
         return null;
     }
+    if (subsetName !== '') {
+        const image = tileSet.imagery.find((i) => i.imagery.name === subsetName);
+        if (image == null) return null;
+        const subTileSet = individualTileSet(tileSet, image);
+        subTileSet.tileSet.background = undefined;
```
blacha, line 73: does this need to be done here or can it be done inside the `individualTileSet` ?

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +16,115 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+        return ts == null ? [] : [ts];
+    }
+
+    const { name, tag } = Aws.tileMetadata.TileSet.parse(nameStr);
+
+    const tileSets: TileSet[] = [];
+
+    if (name === '') {
+        for (const tsn of TileSetNameValues()) {
+            await collectTileSets(tileSets, tag == null ? tsn : `${tsn}@${tag}`, projection);
```
blacha, line 123: I think we need to multithread this, otherwise this is doing upto `TileSetNames.length x projection.length` queries in series.

if we add for instance CITM projection + Topo 50 & 250 could take a while

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +16,115 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
...
+        const subTileSet = individualTileSet(tileSet, image);
+        subTileSet.tileSet.background = undefined;
+        return subTileSet;
+    }
     return tileSet;
 }
+
+async function collectTileSets(tileSets: TileSet[], name: string, projection: Epsg | null): Promise<void> {
+    if (projection == null) {
+        for (const code of ProjectionTileMatrixSet.targetCodes()) {
```
blacha, line 81: multithread this too

packages/lambda-xyz/src/tile.set.cache.ts
```diff
@@ -9,20 +16,115 @@ export function getTileSet(name: string, projection: Epsg): TileSet | undefined
     return TileSets.get(tileSetId);
 }
 
-export async function loadTileSet(req: LambdaContext, name: string, projection: Epsg): Promise<TileSet | null> {
-    req.set('tileSet', name);
-    req.set('projection', projection);
+function titleizeName(name: string): string {
```
blacha, line 19: what does this actually do is a titleize ?

can you give some examples

packages/__tests__/src/rounding.ts
jacott: This is just for testing with numbers with an explicit decimal point in it. We won't get false positives so no need to be elaborate.

packages/shared/src/proj/projection.ts
jacott: What else is there to normalize?

Projections will never span more than the entire glob so no need to be in a loop.

palmerj: One think to check with QGIS and ArcGIS is if bounds over the Anti-merdian are acceptable. e.g

<ows:WGS84BoundingBox>
<ows:LowerCorner>165.0 -50.0</ows:LowerCorner>
<ows:UpperCorner>185.0 -25.0</ows:UpperCorner>
</ows:WGS84BoundingBox>

I know in WFS this works ok. But likely the WMTS ArcGIS codebase is completely different.

palmerj:
> It may be possible to instead create two bounding boxes on each side of the anti-merdian.

Is this possible with the current standard? I know with the new OGC REST APIs it's possible to define the extent as a mulit-polygon which is normalised to the -180 to 180 longitude range.

jacott:
@palmerj two WGS84BoundingBoxes did not work when I tried it with QGIS; it ignored the second one.

packages/shared/src/proj/projection.ts
jacott: ATM it will revert to globe extent if bounds crosses the anti-meridian

blacha: > Projections will never span more than the entire glob so no need to be in a loop.

Is this one of those you will never need more than 64k of ram sort of comments :P

PR #991, opened 2020-07-29
build(deps): aws-cdk 1.55.0

blacha:
Try another different way of getting dependabot to update all dependencies in one PR


blacha:
No longer needed as #1014 is newer

PR #995, opened 2020-07-30
fix(cli): folders must be mounted to allow docker to read the source files

blacha:


packages/cli/src/cog/cog.vrt.ts
```diff
@@ -128,8 +116,14 @@ export const CogVrt = {
...
 
         const gdalCommand = Gdal.create();
-        if (gdalCommand.mount) {
+
+        // If required assume role
+        if (isConfigS3Role(job.source)) {
+            const credentials = Aws.credentials.getCredentialsForRole(job.source.roleArn, job.source.externalId);
+            gdalCommand.setCredentials(credentials);
+        } else if (gdalCommand.mount != null) {
             gdalCommand.mount(tmpFolder);
```
jacott, line 125: This should not be `else if` for mounting `tmpFolder`

PR #1013, opened 2020-08-05
refactor(geojson): move geojson builders to @linzjs/geojson package

jacott:


packages/linzjs-geojson/package.json
```diff
@@ -20,7 +20,7 @@
     "polygon-clipping": "^0.15.1"
   },
   "peerDependencies": {},
-    "devDependencies": {
-        "@types/geojson": "^7946.0.7"
-    }
+  "devDependencies": {
+    "@types/geojson": "^7946.0.7"
```
blacha, line 24: I think since we export these types we need to make this a actual dependency

PR #1023, opened 2020-08-07
feat(cli)! Store stac objects along side CogJob

jacott:
`job.json` is reformatted.
Will copy title and description from source imagery stac file to individual tile sets if present.


packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                temporal,
+            },
+
+            license,
+            keywords,
+
+            providers,
+
+            properties: {
+                created: nowStr,
```
blacha, line 214: is this the prefered way of storing timestamps in here?

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                'eo:gsd': ctx.targetProjection.tms.pixelScale(metadata.resZoom),
+                'proj:epsg': ctx.targetProjection.tms.projection.code,
+                'linz:basemaps': {
+                    name: imageryName,
+                    source: {
+                        files: metadata.bounds,
```
blacha, line 220: Are these files not just "links"?

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                        'proj:epsg': metadata.projection,
+                        'eo:gsd': metadata.pixelScale,
+                    },
+                    sourceLocation: ctx.sourceLocation,
+                    output: {
+                        resampling: ctx.override?.resampling ?? GdalCogBuilderDefaults.resampling,
+                        quality: ctx.override?.quality ?? GdalCogBuilderDefaults.quality,
+                        cutline: ctx.cutline,
```
blacha, line 228: is the cutline a "link"?

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                        addAlpha,
+                        nodata: metadata.nodata,
+                        bounds: metadata.targetBounds,
+                        files: metadata.files,
```
blacha, line 232: same as above

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                    },
+                    outputLocation: ctx.outputLocation,
+                    generated: {
+                        ...CliInfo,
+                        date: nowStr,
```
blacha, line 237: some question about date

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+                'eo:gsd': ctx.targetProjection.tms.pixelScale(metadata.resZoom),
+                'proj:epsg': ctx.targetProjection.tms.projection.code,
+                'linz:basemaps': {
+                    name: imageryName,
+                    source: {
+                        files: metadata.bounds,
+                        'proj:epsg': metadata.projection,
+                        'eo:gsd': metadata.pixelScale,
+                    },
+                    sourceLocation: ctx.sourceLocation,
```
blacha, line 224: are all locations links?

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+        gsd: number;
+        zoom: number;
+    };
+
+    /**
+     * The Name to used when just producing one cog to cover the full extent
+     */
+    static OneCogName = '0-0-0';
+
+    static async build({
```
blacha, line 122: not sold on this syntax, could you break it out into a type and add some comments about what the things are for

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+
+        if (interval.length == 0) {
+            const years = extractYearRangeFromName(imageryName);
+            if (years[0] != -1) {
+                interval.push(years.map((y) => `${y}-01-01T00:00:00Z`) as [string, string]);
+            }
+        }
+
+        if (ctx.cutline) {
+            links.push({ href: 'cutline.geojson.gz', type: 'application/geo+json+gzip', rel: 'cutline' });
```
blacha, line 180: I cant find any references to `application/geo+json+gzip` on the internet this seems to be made up?


I sort of feel like gzipping of this, is causing us more issues than its worth.

packages/cli/src/cog/cog.job.ts
```diff
@@ -0,0 +1,340 @@
...
+            }
+        }
+        const keywords = sourceStac.keywords ?? CogStacKeywords.slice();
+        const license = sourceStac.license ?? CogStacLicense;
+        const title = sourceStac.title ?? titleizeImageryName(imageryName);
+
+        if (interval.length == 0) {
+            const years = extractYearRangeFromName(imageryName);
+            if (years[0] != -1) {
+                interval.push(years.map((y) => `${y}-01-01T00:00:00Z`) as [string, string]);
```
blacha, line 175: do these strings, need to be full ISO date strings, or could we just do years? 

cause most of the flying is not done between jan 1st -> jan 1st? aka we are making up data that is wrong

packages/cli/src/cog/cog.job.ts
jacott: `a+b+c` is how to nest types like `tar+gzip`. See https://en.wikipedia.org/wiki/Media_type

packages/cli/src/cog/cog.job.ts
jacott: The sourceLocation is the config from the command line `roleArn` etc

packages/cli/src/cog/cog.job.ts
jacott: dates are always ISO strings

packages/cli/src/cog/cog.job.ts
jacott: Yes; though I have dropped this field

packages/cli/src/cog/cog.job.ts
jacott: No because if we just have links to the source data we would need to read, sometimes, over 3000 files to get the information we need.

packages/cli/src/cog/cog.job.ts
jacott: This is the format we have to work with. This is what LSD are doing.

packages/cli/src/cog/types.ts
```diff
@@ -2,78 +2,118 @@ import { BoundingBox, EpsgCode } from '@basemaps/geo';
...
+    id: string;
+
+    stac_version: string;
+
+    title: string;
+    description?: string;
+
+    license: string;
+
+    properties: P;
```
jacott, line 94: The actually is no properties field in a Stac Collection but this is what LDS are using. There is a summaries field but its format is a bit cumbersome.

packages/cli/src/cli/cogify/action.batch.ts
```diff
@@ -161,22 +159,28 @@ export class ActionBatchJob extends CommandLineAction {
...
+            LogConfig.get(),
+        );
     }
 
-    static async batchJob(jobPath: string, commit = false, oneCog: string | undefined, logger: LogType): Promise<void> {
-        if (!FileOperator.isS3(jobPath)) throw new Error(`AWS Batch job.json have to be in S3, jobPath:${jobPath}`);
-        const job = await FileOperator.readJson<CogJob>(jobPath);
+    static async batchJob(job: CogJob, commit = false, oneCog: string | undefined, logger: LogType): Promise<void> {
+        const jobPath = job.getJobPath();
+        if (!FileOperator.isS3(jobPath))
```
blacha, line 172: if it multi lines, use brackets

Wonder if we can configure this?

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -0,0 +1,416 @@
...
+        const links: StacLink[] = [
+            {
+                href: FileOperator.join(ctx.outputLocation.path, 'collection.json'),
+                type: 'application/json',
+                rel: 'self',
+            },
+            {
+                href: 'job.json',
+                type: 'application/json',
+                rel: 'job',
```
blacha, line 170: could you be more specific, eg `linz:basemaps:job` or something 

Is there any sort of structure for `rel` and what is in/out?

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -0,0 +1,416 @@
...
+     */
+    static async load(jobpath: string): Promise<CogStacJob> {
+        return new CogStacJob(await FileOperator.readJson<CogJobJson>(jobpath));
+    }
+
+    /**
+     * Create job.json, collection.json, source.geojson, covering.geojson, cutlint.geojson.gz and
+     * stac descriptions of the target COGs
+
+     * @param id unqiue id for this imagery set
```
blacha, line 137: These are not actually `@param`, since its a destructured object.

you only really have one parameter, `context` which is all of this stuff inside of it, to get this to work you would be best to extract this into its own interface and add comments to that.

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -0,0 +1,416 @@
...
+        await FileOperator.writeJson(jobFile, job.json, outputFs);
+
+        const covering = job.targetPtms.proj.toGeoJson(metadata.files);
+
+        const roles = ['data'];
+        const collectionLink = { href: 'collection.json', rel: 'collection' };
+
+        for (const f of covering.features) {
+            const { name } = f.properties as { name: string };
+            const href = name + '.json';
```
blacha, line 303: NIT: I don't think this is a `href`?

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -0,0 +1,416 @@
...
+            if (years[0] != -1) {
+                interval.push(years.map((y) => `${y}-01-01T00:00:00Z`) as [string, string]);
+            }
+        }
+
+        if (ctx.cutline) {
+            links.push({ href: 'cutline.geojson.gz', type: 'application/geo+json+gzip', rel: 'cutline' });
+        }
+
+        links.push({ href: 'covering.geojson', type: 'application/geo+json', rel: 'covering' });
```
blacha, line 238: same comments with `rel` if we are making them up do we want to namespace our ones?

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -0,0 +1,416 @@
...
+                        datetime: nowStr,
+                    },
+                ],
+            },
+
+            links,
+        };
+
+        // Local file systems need directories to be created before writing to them
+        if (!FileOperator.isS3(jobFile)) {
```
blacha, line 288: I think `fs.write` now makes folders

packages/cli/src/cog/cog.stac.job.ts
jacott: I'll rename `linz.basemaps.job`

blacha (approved):

PR #1033, opened 2020-08-10
feat(geo): Add an optional bias when rounding bounds

jacott:
The bias used by `tiler.ts` is 0.2 so that it will round down at < 0.7 and round up at >= 0.3 instead of down at < 0.5
and up at >= 0.5. This leads to tiles favouring covering a pixel larger area than before.

### DIFF
--- a/packages/geo/src/bounds.ts
+++ b/packages/geo/src/bounds.ts
@@ -158,11 +158,14 @@ export class Bounds implements BoundingBox {
 
     /**
      * Round dimensions to integers keeping the error a low as possible
+
+     * @param bias influence the rounding in favour of a larger area (+ve bias) or smaller area (-ve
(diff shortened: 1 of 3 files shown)


packages/geo/src/bounds.ts
```diff
@@ -160,9 +163,9 @@ export class Bounds implements BoundingBox {
      * Round dimensions to integers keeping the error a low as possible
      */
     public round(): Bounds {
```
blacha, line 165: Do we want to do this for all Rounds?

do we want to `round(bias:number = 0.2)` ? or `round(bias?:number)` and default to 0 then explicitly put the bias in?

blacha: or even `roundWithBias()`

PR #1049, opened 2020-08-12
feat(s3fs): expose standard error codes for not found and forbidden

blacha:
use 403 and 404 for errors across s3 and local instead of fs:EACESS or fs:ENOENT and s3:NotFound

### DIFF
--- a/packages/linzjs-s3fs/src/file.s3.ts
+++ b/packages/linzjs-s3fs/src/file.s3.ts
@@ -1,9 +1,16 @@
-import { S3 } from 'aws-sdk';
+import type { S3, AWSError } from 'aws-sdk';
 import { Readable, Stream } from 'stream';
 import { CompositeError } from './composite.error';
 import { FileProcessor } from './file';
 
-export class FileProcessorS3 implements FileProcessor {
+function getCompositeError(e: AWSError, msg: string): CompositeError {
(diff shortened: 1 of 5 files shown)


packages/linzjs-s3fs/src/file.s3.ts
```diff
@@ -1,9 +1,16 @@
...
+import type { S3, AWSError } from 'aws-sdk';
 import { Readable, Stream } from 'stream';
 import { CompositeError } from './composite.error';
 import { FileProcessor } from './file';
 
-export class FileProcessorS3 implements FileProcessor {
+function getCompositeError(e: AWSError, msg: string): CompositeError {
+    if (e == null) return new CompositeError(msg, 500, e);
+    if (e.code == 'NotFound') return new CompositeError(msg, 404, e);
+    if (e.code == 'AccessDenied') return new CompositeError(msg, 403, e);
```
jacott, line 9: I thought the code was 'ResourceNotFoundException'

blacha: Where did you get that code from, I was using https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList for my list

jacott: maybe that was from dynamodb; can't seem to find it anymore?

PR #1051, opened 2020-08-12
feat: allow configuration of tile resize kernels

blacha:
for some types of imagery nearest is better for resizing


blacha:
Forcing lanczos  resize sometimes causes weird artifacts:

output_17_26_z4_lanczos

Switching some resizing to use nearest removes some of the artifacts
output_17_26_z4_nearest

PR #1092, opened 2020-08-25
feat(shared): align bathymetry STAC usage with cog creation

jacott:
1. Refactor STAC interface; move to shared.
2. Add a bathymetry collection.json
3. Improve STAC Item compliance in bathymetry and cog creation

### DIFF
--- a/packages/bathymetry/src/bathy.maker.ts
+++ b/packages/bathymetry/src/bathy.maker.ts
@@ -1,17 +1,20 @@
 import { Gdal } from '@basemaps/cli';
-import { Epsg, Tile, TileMatrixSet } from '@basemaps/geo';
+import { Bounds, Epsg, Tile, TileMatrixSet } from '@basemaps/geo';
 import { LogType } from '@basemaps/shared';
 import * as fs from 'fs';
-import PLimit from 'p-limit';
+import * as os from 'os';
 import type { Limit } from 'p-limit';
+import PLimit from 'p-limit';
 import * as path from 'path';
+import { basename } from 'path';
 import { FilePath, FileType } from './file';
 import { Hash } from './hash';
 import { MapnikRender } from './mapnik';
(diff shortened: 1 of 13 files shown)


packages/bathymetry/src/bathy.maker.ts
```diff
@@ -82,17 +85,24 @@ export class BathyMaker {
...
+                    extent = extent == null ? bounds : extent.union(bounds);
                     const tileId = TileMatrixSet.tileToName(tile);
                     return this.renderTile(tile, logger.child({ index: tileId }));
                 });
                 promises.push(promise);
             }
         }
-        await Promise.all(promises).catch((err) => logger.fatal(err, 'FailedToRun'));
+        try {
+            this.createMetadata(extent!, await Promise.all(promises));
```
blacha, line 102: best not to do a not null assertion here, as this may actually be null?

Since the extent is only done in side the promise function itself?
something like
```
await Promise.all(promises);
if (extent == null) throw ..
this.createMetadata(...)
```

packages/bathymetry/src/bathy.maker.ts
```diff
@@ -107,15 +117,11 @@ export class BathyMaker {
     }
 
     /** Render a specific tile */
-    async renderTile(tile: Tile, logger: LogType): Promise<void> {
-        try {
```
blacha, line 111: This logger has a log context including the tile index that failed.

if you want to remove the log line here, maybe catch the error and re throw a composite error with which tile failed

packages/bathymetry/src/bathy.maker.ts
jacott: There is no code path where it could be null (unless tmsZoom.maxtrixHeight or matrixWidth) are null. The `!` is asserting that

blacha (approved):

PR #1122, opened 2020-09-01
feat(bathymetry): allow input and output from s3 bucket

jacott:
Also use source bucket collection.json for title, description, interval and provider details


packages/bathymetry/src/hash.ts
```diff
@@ -16,7 +16,7 @@ function hashFile(filePath: string, hashName: string): Promise<Buffer> {
 /** Create a multihash of a file */
 async function hash(filePath: string): Promise<string> {
     const hashData = await hashFile(filePath, 'sha256');
-    return Multihash.encode(hashData, 'sha2-256').toString('hex');
+    return Buffer.from(Multihash.encode(hashData, 'sha2-256')).toString('hex');
```
blacha, line 19: how come this changed?

blacha (approved):

PR #1176, opened 2020-09-15
fix: STAC files should comply to 1.0.0-beta.2 of the specification

jacott:
* Add linz basemaps schema.json
* Fix `extent/spatial/bbox` to `number[][]`
* `description` is always required
* Set `stac_version` to `1.0.0-beta.2`
* Don't use `projection` extension for `collection.json`

### DIFF
--- a/packages/cli/src/cog/cog.stac.job.ts
+++ b/packages/cli/src/cog/cog.stac.job.ts
@@ -5,6 +5,7 @@ import {
     FileConfigPath,
     FileOperator,
     ProjectionTileMatrixSet,
+    StacBaseMapsExtension,
     StacCollection,
     StacLicense,
     StacLink,
@@ -15,7 +16,7 @@ import {
 import { MultiPolygon, toFeatureCollection, toFeatureMultiPolygon } from '@linzjs/geojson';
 import { CliInfo } from '../cli/base.cli';
 import { GdalCogBuilderDefaults, GdalCogBuilderResampling } from '../gdal/gdal.config';
-import { CogStac, CogStacExtensions, CogStacItem, CogStacItemExtensions, CogStacKeywords } from './stac';
+import { CogStac, CogStacItem, CogStacItemExtensions, CogStacKeywords } from './stac';
 import {
     CogBuilderMetadata,
     CogJob,
@@ -150,7 +151,17 @@ export class CogStacJob implements CogJob {
             description = sourceStac.description;
             interval.push(...(sourceStac.extent?.temporal?.interval ?? []));
             links.push({ href: sourceCollectionPath, rel: 'sourceImagery', type: 'application/json' });
-            if (sourceStac.providers != null) providers.push(...sourceStac.providers);
+            if (sourceStac.providers != null) {
+                for (const p of sourceStac.providers) {
+                    if (p.roles.indexOf('host') == -1) {
+                        if (p.url === 'unknown') {
+                            // LINZ LDS has put unknown in some urls
+                            p.url = undefined;
+                        }
+                        providers.push(p);
+                    }
+                }
+            }
         } catch (err) {
             if (!FileOperator.isCompositeError(err) || err.code !== 404) {
                 throw err;
@@ -159,6 +170,10 @@ export class CogStacJob implements CogJob {
         const keywords = sourceStac.keywords ?? CogStacKeywords.slice();
         const license = sourceStac.license ?? StacLicense;
         const title = sourceStac.title ?? titleizeImageryName(imageryName);
+
+        if (description == null) {
+            description = 'No description';
+        }
 
         const job = new CogStacJob({
             id,
@@ -191,9 +206,11 @@ export class CogStacJob implements CogJob {
 
         const sourceProj = job.sourcePtms.proj;
 
-        const bbox = sourceProj.boundsToWgs84BoundingBox(
-            metadata.bounds.map((a) => Bounds.fromJson(a)).reduce((sum, a) => sum.union(a)),
-        );
+        const bbox = [
+            sourceProj.boundsToWgs84BoundingBox(
+                metadata.bounds.map((a) => Bounds.fromJson(a)).reduce((sum, a) => sum.union(a)),
+            ),
+        ];
 
         if (interval.length == 0) {
             const years = extractYearRangeFromName(imageryName);
@@ -218,7 +235,7 @@ export class CogStacJob implements CogJob {
             title,
             description,
             stac_version: StacVersion,
-            stac_extensions: CogStacExtensions,
+            stac_extensions: [StacBaseMapsExtension],
 
             extent: {
                 spatial: {
--- a/packages/landing/scripts/bundle.js
+++ b/packages/landing/scripts/bundle.js
@@ -9,6 +9,7 @@ const crypto = require('crypto');
 const pkg = require('../package.json');
 const fs = require('fs').promises;
 const c = require('ansi-colors');
+const { recurseDirectory } = require('./util');
 
 // Proxy environment vars into the build
 const ENV_VARS = ['API_KEY', 'NODE_ENV', 'GOOGLE_ANALYTICS', 'TILE_HOST'].map((envName) => {
@@ -27,6 +28,12 @@ const BUILD_CMD = [
     process.env.NODE_ENV == 'production' ? '--minify' : '',
     'src/index.ts',
 ];
+
+const SPECIAL_FILES = {
+    'index.css': true,
+    'index.html': true,
+    'index.js': true,
+};
 
 /**
  * List of CSS files to bundle
@@ -99,8 +106,28 @@ async function bundleHtml(opts) {
     await fs.writeFile('./dist/index.html', htmlOutput);
 }
 
+/**
+ * Copy all non special files from the static directory to the dist directory including any sub
+ * directories.
+ */
+async function copyStatic() {
+    const srcDir = './static';
+    await recurseDirectory(srcDir, async (filePath, isDir) => {
+        if (SPECIAL_FILES[filePath]) return false;
+        const srcPath = `${srcDir}/${filePath}`;
+        const destPath = `./dist/${filePath}`;
+        if (isDir) {
+            await fs.mkdir(destPath, { recursive: true });
+        } else {
+            await fs.writeFile(destPath, await fs.readFile(srcPath));
+        }
+        return true;
+    });
+}
+
 async function main() {
     await fs.mkdir('dist', { recursive: true });
+    await copyStatic();
 
     const js = await bundleJs();
     const css = await bundleCss();
--- a/packages/landing/scripts/deploy.js
+++ b/packages/landing/scripts/deploy.js
@@ -4,6 +4,7 @@ const fs = require('fs').promises;
 const crypto = require('crypto');
 const mime = require('mime-types');
 const invalidateCache = require('@basemaps/cli/build/cli/util').invalidateCache;
+const { recurseDirectory } = require('./util');
 
 const DistDir = './dist';
 const HashKey = 'linz-hash';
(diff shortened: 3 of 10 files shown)


packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -191,9 +196,11 @@ export class CogStacJob implements CogJob {
 
         const sourceProj = job.sourcePtms.proj;
 
-        const bbox = sourceProj.boundsToWgs84BoundingBox(
-            metadata.bounds.map((a) => Bounds.fromJson(a)).reduce((sum, a) => sum.union(a)),
-        );
+        const bbox = [
+            sourceProj.boundsToWgs84BoundingBox(
```
blacha, line 200: Does this break with the antimeridian?

is the reason their are multiple bbox's for the antimeridian?

packages/landing/static/json-schema/basemaps-extension/1.0/schema.json
```diff
@@ -0,0 +1,101 @@
...
+        {"$ref": "https://schemas.stacspec.org/v1.0.0-beta.2/item-spec/json-schema/item.json"},
+        {"$ref": "#/definitions/linz-common"},
+        {"$ref": "#/definitions/linz-item"}
+      ],
+    },
+    {
+      "allOf": [
+        {"$ref": "https://schemas.stacspec.org/v1.0.0-beta.2/collection-spec/json-schema/collection.json"},
+        {"$ref": "#/definitions/linz-common"},
+        {"$ref": "#/definitions/linz-collection"},
```
blacha, line 18: we cant use `linz-common` maybe `linz-basemaps-common` ?

packages/shared/src/stac/index.ts
```diff
@@ -1,5 +1,6 @@
-export const StacVersion = '1.0.0';
+export const StacVersion = '1.0.0-beta.2';
 export const StacLicense = 'CC-BY-4.0';
+export const StacBaseMapsExtension = 'https://basemaps.linz.govt.nz/json-schema/basemaps-extension/1.0/schema.json';
```
blacha, line 3: is this URL a standard for this schema?

How is the file deployed?

packages/cli/src/cog/cog.stac.job.ts
jacott: No. only one bbox is needed event when AM is crossed; it just means the first longitude will be greater than the second longitude

packages/landing/static/json-schema/basemaps-extension/1.0/schema.json
jacott: It is just a local ref; we can use whatever we want

packages/shared/src/stac/index.ts
jacott: There is no standard for vendor extensions. I am trying to use a somewhat canonical path.

The file should be deployed as part of the landing package (assuming landing deploys everything under `static`)

packages/landing/static/json-schema/basemaps-extension/1.0/schema.json
jacott: I changed it to just "common", "collection" and "item"

packages/shared/src/stac/index.ts
jacott: I see that the deploy does not deploy everything under static. How is `favicon-16x16.png` etc getting deployed?

blacha: >  I see that the deploy does not deploy everything under static. How is favicon-16x16.png etc getting deployed?

Not too sure maybe I broke it a while back? possibly should be fixed :)

jacott: It is deploying every file under dist but not descending directories; now it does.

jacott: Actually only deploy copies everything under `dist`. Bundle needs to be modified to copy everything from `static` to `dist`

jacott: `bundle` now recursively copies `static` files (not `index.css` etc) to `dist`

packages/shared/src/stac/index.ts
```diff
@@ -1,5 +1,7 @@
-export const StacVersion = '1.0.0';
+export const StacVersion = '1.0.0-beta.2';
 export const StacLicense = 'CC-BY-4.0';
+export const StacBaseMapsExtension =
```
blacha, line 3: Do we hard code `basemaps.linz.govt.nz` anywhere else? Ideally we should be injecting this later?

Could we make it a config variable inside the bundler?

jacott: Do we know the host at run time for `cli `and `bathymetry` packaes and at deploy time for `landing`?

packages/landing/static/json-schema/stac-basemaps-extension/1.0/schema.json
```diff
@@ -0,0 +1,139 @@
...
+          "$ref": "#/definitions/collection"
+        }
+      ]
+    }
+  ],
+  "definitions": {
+    "sampling-method": {
+      "type": "string",
+      "enum": [
+        "nearest",
```
blacha, line 38: do you have any thoughts on how we can keep this upto date?

I wonder if we should just use a typescript -> jsonschema to create this.

packages/landing/scripts/bundle.js
```diff
@@ -99,8 +105,31 @@ async function bundleHtml(opts) {
...
+            await fs.mkdir(destPath);
+            await recurseStatic(filePath);
+            continue;
+        }
+        await fs.writeFile(destPath, await fs.readFile(srcPath));
+    }
+}
+
 async function main() {
+    await fs.rmdir('dist', { recursive: true });
```
blacha, line 130: I dont think we want to delete this here.

I believe there is `clean` if you want to clean it.

packages/landing/scripts/bundle.js
```diff
@@ -99,8 +105,31 @@ async function bundleHtml(opts) {
     await fs.writeFile('./dist/index.html', htmlOutput);
 }
 
+async function recurseStatic(subDir = '') {
```
blacha, line 108: Could we get a comment as to what this does?

This also seems to be quite similar to the logic below could they use a shared function for the work?

packages/landing/scripts/bundle.js
```diff
@@ -99,8 +105,31 @@ async function bundleHtml(opts) {
...
+    const srcDir = './static' + subDir;
+    const files = await fs.readdir(srcDir);
+    for (const fileName of files) {
+        const filePath = `${subDir}/${fileName}`;
+
+        console.log(`DEBUG SPECIAL_FILES[filePath]`, SPECIAL_FILES[filePath], filePath);
```
blacha, line 114: do we need this console.log?

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -150,7 +151,17 @@ export class CogStacJob implements CogJob {
             description = sourceStac.description;
             interval.push(...(sourceStac.extent?.temporal?.interval ?? []));
             links.push({ href: sourceCollectionPath, rel: 'sourceImagery', type: 'application/json' });
-            if (sourceStac.providers != null) providers.push(...sourceStac.providers);
+            if (sourceStac.providers != null) {
+                for (const p of sourceStac.providers) {
+                    if (p.roles.indexOf('host') == -1) {
+                        if (p.url === 'unknown') {
+                            // LINZ LDS has put unknown in some urls
+                            p.url = undefined;
```
blacha, line 159: is unknown not allowed in the spec?

packages/shared/src/stac/index.ts
jacott: I think, because this is a link to a LINZ specific schema, the well known location should always be used.

packages/landing/static/json-schema/stac-basemaps-extension/1.0/schema.json
jacott: I don't think the list should change unless the schema version also changes.

packages/landing/scripts/bundle.js
jacott: I think a bundle should clear the directory because there could be leftover files there that are no longer in `static`. I'm okay with not though if you prefer not to.

packages/landing/scripts/bundle.js
jacott: oops debug leftover

packages/cli/src/cog/cog.stac.job.ts
jacott: correct

packages/cli/src/cog/cog.stac.job.ts
palmerj: Hi Guys, it's best practise to split across the anti-meridian for interoperability reasons with clients. See https://tools.ietf.org/html/rfc7946#section-3.1.9

packages/shared/src/stac/index.ts
blacha: > because this is a link to a LINZ specific schema

Ok good reasoning!

packages/landing/static/json-schema/stac-basemaps-extension/1.0/schema.json
blacha: If I go add another sampling-method to the code and it generates the stac files from that code, this schema is not representative of what is actually being built.

Is this caught anywhere? Like do I know that adding another sampling method to the source will break this schema?

packages/landing/scripts/bundle.js
blacha: All the other bundlers I use, webpack, parcel etc, have a very specific "clean" function if things are being deleted.

The static folder in theory is only deployed from a clean state so it shouldnt normally matter, and if you are doing a deployment you really want to `clean` the entire repo.

TLDR: yes please take it out.

packages/cli/src/cog/cog.stac.job.ts
blacha: @jacott with Jeremy's comment maybe we should split it on the antimeridian? Does the stac spec tell us not to?

packages/landing/scripts/bundle.js
jacott: DRYing up this code will involve more insertions than deletions but okay.

packages/landing/static/json-schema/stac-basemaps-extension/1.0/schema.json
jacott: I'll make the schema accept any string

packages/cli/src/cog/cog.stac.job.ts
jacott: @blacha , @palmerj The GeoJSON RFC does not split `bbox`. STAC uses the GeoJSON for items but I defer to others and will split if you think I should.

jacott: What the link above is for is polygons which we do split.

palmerj: Ok thanks. You are correct. As long as we split the footprint geometry then we are ok. Clients querying on the bbox via the STAC API will of course need to normalise their input bboxes

packages/landing/scripts/deploy.js
```diff
@@ -40,36 +41,45 @@ async function deploy() {
...
-        // Only upload files if they have changed
-        if (oldHash == hash) {
-            console.log('Skipped', fileName, `${(fileData.byteLength / 1024).toFixed(2)}Kb`, hash);
-            continue;
-        }
-        hasChanges = true;
+    /**
+     * Upload all files from the dist directory, including any sub directories, to S3 bucket .
+     */
+    const uploadDirectory = async () => {
```
blacha, line 48: NIT: Why make a function and call it directly after? could you not just execute the function

blacha (approved):

PR #1195, opened 2020-09-21
feat(lambda-analytics): allow analytics to be reprocessed by removing  the cached data

blacha:

### DIFF
--- a/packages/lambda-analytics/src/__tests__/file.process.test.ts
+++ b/packages/lambda-analytics/src/__tests__/file.process.test.ts
@@ -16,9 +16,13 @@ export const ExampleLogs = `#Version: 1.0
     .trim()
     .split('\n');
 
-export function lineReader(lines: string[]): () => AsyncGenerator<string> {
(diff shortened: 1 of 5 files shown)


packages/lambda-analytics/src/__tests__/file.process.test.ts
```diff
@@ -55,23 +57,22 @@ o.spec('FileProcess', () => {
...
-        const logData = LogStats.getDate('2020-07-28', '01');
+        FileProcess.reader = lineReader(ExampleLogs);
+        await FileProcess.process('Fake', logData, LogConfig.get());
         o(LogStats.ByDate.size).equals(1);
 
         o(logData.stats.size).equals(2);
         const devStats = logData.getStats(DevApiKey, 'bar.com');
         const clientStats = logData.getStats(ClientApiKey);
 
-        o(devStats?.total).equals(4);
```
blacha, line 69: stats were not getting reset, so these were all off by 1 hit

jacott (approved):

PR #1205, opened 2020-09-22
feat(lambda-tiler): attribution

jacott:
Adds attribution to the landing page


packages/bathymetry/src/__test__/stac.test.ts
```diff
@@ -97,12 +96,6 @@ o.spec('stac', () => {
         bm.config = bm;
 
         o('createCollection without source collection.json', async () => {
-            mockFs.jsStore['s3:///test-source-bucket/gebco-2020/collection.json'] = new CompositeError(
-                'not found',
-                404,
-                new Error(),
-            );
-
```
jacott, line 105: This is now the default action when entry not in mockFs

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -214,7 +214,10 @@ export class CogStacJob implements CogJob {
 
         if (interval.length == 0) {
             const years = extractYearRangeFromName(imageryName);
-            if (years[0] != -1) {
+            if (years[0] == -1) {
+                const now = new Date().toISOString();
+                interval.push([now, now]);
+            } else {
```
jacott, line 220: Non existent dates are not allow in stac files

packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -284,9 +285,11 @@ export class CogStacJob implements CogJob {
         for (const f of covering.features) {
             const { name } = f.properties as { name: string };
             const href = name + '.json';
+            const bbox = f.bbox as BBox;
             links.push({ href, type: 'application/geo+json', rel: 'item' });
             const item: CogStacItem = {
                 ...f,
+                bbox,
```
jacott, line 292: cause; typescript

packages/lambda-tiler/src/__test__/xyz.util.ts
```diff
@@ -50,6 +50,6 @@ export const Provider: TileMetadataProviderRecord = {
             position: 'the position',
         },
         name: 'the name',
-        site: 'the site',
+        site: 'https://example.provider.com',
```
jacott, line 53: Makes STAC results easier to run through a validator

packages/lambda-tiler/src/cli/dump.ts
```diff
@@ -49,7 +49,7 @@ async function main(): Promise<void> {
         logger,
     );
 
-    const tileData = await tile(ctx, { ...xyz, projection, name: tileSet.name, ext, type: TileType.Image });
+    const tileData = await tile(ctx);
```
jacott, line 52: `tile` function now computes the other values from `ctx`

packages/lambda-tiler/src/cli/serve.ts
```diff
@@ -32,7 +32,7 @@ async function handleRequest(
             }
         }
         if (data.status < 299 && data.status > 199) {
-            res.end(Buffer.from(data.getBody() ?? '', 'base64'));
+            res.end(data.body);
```
jacott, line 35: This should be passed through verbatim

packages/lambda-tiler/src/cli/serve.ts
```diff
@@ -58,6 +58,10 @@ function useAws(): void {
             } as any,
             logger,
         );
+        const ifNoneMatch = req.header(HttpHeader.IfNoneMatch);
+        if (ifNoneMatch != null) {
+            ctx.headers.set(HttpHeader.IfNoneMatch.toLowerCase(), ifNoneMatch);
+        }
```
jacott, line 64: Maybe I should set all the headers?

packages/lambda-tiler/src/routes/api.ts
```diff
@@ -14,6 +14,6 @@ export async function Ping(): Promise<LambdaHttpResponse> {
 export async function Version(): Promise<LambdaHttpResponse> {
     const response = new LambdaHttpResponse(200, 'ok');
     response.header(HttpHeader.CacheControl, 'no-store');
-    response.json({ version: process.env.GIT_VERSION, hash: process.env.GIT_HASH });
+    response.json({ version: process.env.GIT_VERSION ?? 'dev', hash: process.env.GIT_HASH });
```
jacott, line 17: So we always get a version

packages/cli/src/cog/cog.stac.job.ts
blacha: do we maybe want to default to 1970 then?

packages/cli/src/cog/cog.stac.job.ts
blacha: It would imply there is a structural difference between the two types? what is the difference

packages/lambda-tiler/src/cli/serve.ts
blacha: If you want improve the CLI go for it :)

packages/cli/src/cog/cog.stac.job.ts
jacott: bbox if optional for GeoJSON but compulsory for STAC

packages/lambda-tiler/src/routes/attribution.ts
```diff
@@ -0,0 +1,211 @@
...
+    for (const rule of tileSet.tileSet.rules) {
+        const im = tileSet.imagery.get(rule.imgId);
+        if (im == null) continue;
+        const stac = await stacFiles.get(im.uri);
+
+        const bbox = proj.boundsToWgs84BoundingBox(im.bounds).map(roundNumber) as BBox;
+
+        let interval: [string, string][] | undefined = stac?.extent.temporal?.interval;
+        if (interval == null) {
+            const years = extractYearRangeFromName(im.name);
```
blacha, line 115: do we have any stac files that need this? I would prefer not to guess names here, we should just fix the STAC file

packages/cli/src/cog/cog.stac.job.ts
jacott: I could make a new interface with a compulsory `bbox` for Covering files and use that instead of `GeoJSON`? At least that we I wont be executing code the get around the type difference.

packages/lambda/src/lambda.response.ts
```diff
@@ -47,7 +47,7 @@ export class LambdaHttpResponse {
         return false;
     }
 
-    json(obj: Record<string, unknown>): void {
+    json(obj: Record<string, any>): void {
```
blacha, line 50: was the unknown causing issues?

packages/landing/src/attribution.ts
```diff
@@ -0,0 +1,205 @@
...
+        } else {
+            throw new Error(`fetch attribution failed [${resp.status}] ${resp.statusText}`);
+        }
+    } catch (err) {
+        console.error(err);
+        return null;
+    }
+}
+
+function sameContents(a: any[], b: any[]): boolean {
```
blacha, line 90: this wouldnt work for objects, did you want to remove the any and  make it `null | string | number` ?

packages/landing/src/attribution.ts
```diff
@@ -0,0 +1,205 @@
...
+        this._scheduled = setTimeout(() => {
+            this._scheduled = undefined;
+            this._raf = requestAnimationFrame(this.renderAttribution);
+        }, 200);
+    }
+
+    renderAttribution = (): void => {
+        this._raf = 0;
+        if (this.attributions == null) {
+            return;
```
blacha, line 186: do we want to default this to a single copyright?

packages/landing/static/index.css
```diff
@@ -13,7 +13,10 @@ button {
 
 #map {
     width: 100%;
-    height: calc(100% - 58px);
+    left: 0;
```
blacha, line 16: the calc was here for browser compatilibty, does this work on everything, ios? safari? ie11? firefox?

packages/cli/src/cog/cog.stac.job.ts
jacott: Added new type `BBoxFeature`

packages/lambda/src/lambda.response.ts
jacott:
```
packages/lambda-tiler/src/routes/attribution.ts:208:19 - error TS2345: Argument of type 'AttributionStac' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature is missing in type 'AttributionStac'.

208     response.json(attributions);
                      ~~~~~~~~~~~~
```

packages/lambda-tiler/src/routes/attribution.ts
jacott: It is for when we don't have a stac file; in practice we should have one but hard to guarantee.

jacott: I can remove the `?` in `temporal?`; it should not be null.

packages/landing/src/attribution.ts
jacott: It does work for objects because the are guaranteed to be the same objects. However since this is non obvious I will not use it and instead compare the rendered string which it probably quicker to compute anyway.

packages/landing/static/index.css
jacott: the `- 58px` was wrong for my browser. It does work for me in firefox, google and webkit2 but will switch to a flexbox to ensure it is more progressive.

jacott: Now uses `flexbox`. There were a bunch of errors in the `index.html` so I fixed those up as well.

packages/lambda-tiler/src/cli/serve.ts
jacott: maybe later :)

packages/lambda-tiler/src/routes/attribution.ts
blacha: Can we just not have a date? Id prefer not to have this code here, we should fix the stac files if they are broken.

packages/landing/static/index.css
blacha: We might need to do a bunch of testing across browsers on this, I wonder if there is a easy way to automate it

blacha (approved):

PR #1219, opened 2020-09-28
feat: Update browser examples

jacott:
Move examples to landing package

Bundle examples to `https://[dev.]basemaps.linz.govt.nz/examples`

Add attribution library and example


scripts/examples/index.leaflet.xyz.3857.html
```diff
@@ -20,17 +20,18 @@
...
+      const start_pos = [-40.5, 173];
+      const start_zoom = 6;
 
-      var url = "https://basemaps.linz.govt.nz/v1/tiles/aerial/3857/{z}/{x}/{y}.webp?api=mykey";
-      var attributions = '<a href="https://www.linz.govt.nz/data/linz-data/linz-basemaps/data-attribution">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
+      const url = "https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/{z}/{x}/{y}.webp?api="+ApiKey;
+      const attributions = '<a href="https://www.linz.govt.nz/data/linz-data/linz-basemaps/data-attribution">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
 
-      var tiles = L.tileLayer(url, {
+      const tiles = L.tileLayer(url, {
```
blacha, line 30: NIT, could `{ attribution }` ?

scripts/examples/index.openlayers.attribution.wmts.3857.html
```diff
@@ -0,0 +1,238 @@
...
+        href="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.3.1/css/ol.css"
+        type="text/css"
+    />
+    <script src="https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v6.3.1/build/ol.js"></script>
+  </head>
+
+  <body>
+    <div id="map" style="width: 800px; height: 600px;"></div>
+    <script>
+     // Copyied from @linzjs/geojson
```
blacha, line 17: Maybe we should look at exposing these on a CDN somewhere?

scripts/examples/index.openlayers.attribution.wmts.3857.html
```diff
@@ -0,0 +1,238 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>OpenLayers WMTS Basemaps Demo</title>
+    <link
```
blacha, line 6: could we add integrity attrs when using cdns

scripts/examples/index.openlayers.attribution.wmts.3857.html
```diff
@@ -0,0 +1,238 @@
...
+
+         return lon;
+     };
+
+     function normExtent(extent) {
+         return [normLon(extent[0]), extent[1], normLon(extent[2]), extent[3]];
+     }
+     // end of @linzjs/geojson
+
+     class AttributionBounds {
```
blacha, line 61: is this copy and pasted from somewhere?

scripts/examples/index.openlayers.wmts.3857.html
```diff
@@ -4,72 +4,69 @@
...
-      var start_pos = [19467552, -5074414];
-      var start_zoom = 6;
-      var url = "https://tiles.dev.basemaps.linz.govt.nz/v1/tiles/aerial/3857/WMTSCapabilities.xml?api=mykey";
-      var attributions =
-        '<a href="https://www.linz.govt.nz/data/linz-data/linz-basemaps/data-attribution">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
+     const ApiKey = REPLACE_THIS_WITH_API_KEY;
+     var attributions =
+         '<a href="https://www.linz.govt.nz/data/linz-data/linz-basemaps/data-attribution">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
+     const projstr = "EPSG:3857";
+     const start_pos = [19467552, -5074414];
```
blacha, line 21: can we be consistent about casing? snake_case vs camelCase

scripts/examples/index.openlayers.wmts.3857.html
```diff
@@ -4,72 +4,69 @@
...
-        });
-      });
+         const source = new ol.source.WMTS(options);
+         const view  = new ol.View({
+             center: start_pos,
+             zoom: start_zoom
+         });
+         const tiles = new ol.layer.Tile({
+             source,
+             projection: projection,
```
blacha, line 61: some shorthand some not shorthand?

scripts/examples/index.openlayers.attribution.wmts.3857.html
jacott: this is from `packages/linzjs-geojson/src/wgs84.ts`

packages/landing/src/attribution.index.ts
```diff
@@ -0,0 +1,4 @@
+import { Attribution } from './attribution';
+import './global';
+
+window.basemapsAttribution = Attribution;
```
blacha, line 4: BasemapsAttribution

packages/landing/src/attribution.ts
```diff
@@ -1,24 +1,24 @@
...
 import OlPolygon from 'ol/geom/Polygon';
-import OlMap from 'ol/Map';
-import MapEventType from 'ol/MapEventType';
 import { transformExtent } from 'ol/proj';
-import Source from 'ol/source/Source';
-import { MapOptions, MapOptionType, WindowUrl } from './url';
 
-const Copyright = `© ${StacLicense} LINZ`;
+function escape(text: string): string {
+    return text.replace(/</g, '&lt;');
```
blacha, line 10: this doesnt seem the safest way to escape strings,  most people also do > and &

is there a better way of doing this?

packages/landing/src/ol.attribution.ts
```diff
@@ -0,0 +1,127 @@
...
+import { Extent } from 'ol/extent';
+import OlMap from 'ol/Map';
+import MapEventType from 'ol/MapEventType';
+import Source from 'ol/source/Source';
+import { Attribution } from './attribution';
+import { MapOptions, MapOptionType, WindowUrl } from './url';
+
+const Copyright = `© ${StacLicense} LINZ`;
+
+async function fetchAttributions(url: string): Promise<AttributionStac | null> {
```
blacha, line 13: Want to move this into the Attributions eg could do `Attributions.load(url: string)`

packages/landing/src/ol.attribution.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Initialize monitoring the OpenLayers map and set the source attributions when changed.
+     */
+    static init(source: Source, map: OlMap, config: MapOptions): void {
+        const attribution = new OlAttribution(source, map.getView(), config);
+
+        map.addEventListener(MapEventType.MOVEEND, () => {
+            attribution.updateAttribution().catch((err) => {
+                console.error(err);
```
blacha, line 58: Do we want to be just logging these to console, or just throw the error?

blacha: Also if this happens multiple times, does the catch override the previous catchcall?

packages/landing/src/ol.attribution.ts
```diff
@@ -0,0 +1,127 @@
...
+        this.view = view;
+        this.config = config;
+    }
+
+    /**
+     * Trigger an attribution text update. Will fetch attributions if needed
+     */
+    async updateAttribution(): Promise<void> {
+        if (this.attributions == null) {
+            if (this.attributions === null) return; // already fetching
```
blacha, line 74: How about, `this.attributions` is a `Promise<Attribution> | null` 

The `== null` vs `=== null` is quite confusing.

packages/landing/src/ol.attribution.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/**
+ * Handles displaying attributions for the OpenLayers interface
+ */
+export class OlAttribution {
+    source: Source;
+    view: View;
+    config: MapOptions;
+    private _scheduled: any;
```
blacha, line 42: want to throw a few comments as to what these are.

packages/landing/static/examples/index.openlayers.attribution.wmts.3857.html
```diff
@@ -0,0 +1,79 @@
...
+         matrixIds: matrixIds
+     });
+
+     const view  = new ol.View({
+         center: startPos,
+         zoom: startZoom
+     });
+
+     (async () => {
+         const attrResp = await fetch(attrUrl);
```
blacha, line 50: want to Promise.all the two fetches so they run at the same time.

packages/landing/static/examples/index.openlayers.wmts.3857.html
```diff
@@ -0,0 +1,69 @@
...
+     const matrixIds = new Array(14);
+     for (let z = 0; z < 14; ++z) {
+         resolutions[z] = size / Math.pow(2, z);
+         matrixIds[z] = z;
+     }
+
+     const parser = new ol.format.WMTSCapabilities();
+     const grid = new ol.tilegrid.WMTS({
+         origin: ol.extent.getTopLeft(projectionExtent),
+         resolutions: resolutions,
```
blacha, line 39: We use const/short hands in other places we should be able to simplify these here too

packages/landing/static/examples/index.openlayers.wmts.3857.html
```diff
@@ -0,0 +1,69 @@
...
+  </head>
+
+  <body>
+    <div id="map" style="width: 800px; height: 600px;"></div>
+    <script>
+     const apiKey = localStorage.getItem('api-key');
+     var attributions =
+         '<a href="https://www.linz.govt.nz/data/linz-data/linz-basemaps/data-attribution">LINZ CC BY 4.0 © Imagery Basemap contributors</a>';
+     const projstr = "EPSG:3857";
+     const startPos = [19467552, -5074414];
```
blacha, line 21: could make these short hands `const center = [..]` or just add them to the view itself?

packages/landing/src/attribution.ts
jacott: We can use the browser escape but it is deprecated. I think it would be the better option though. Otherwise I can at least escape `&` not much point in escaping `>`.

packages/landing/src/ol.attribution.ts
jacott: I'll leave the promise rejection unhandled.

packages/landing/src/attribution.ts
jacott: I decided to escape `<` and `&`

jacott: I am now escaping via the DOM

packages/landing/src/attribution.ts
```diff
@@ -1,24 +1,27 @@
...
 import OlPolygon from 'ol/geom/Polygon';
-import OlMap from 'ol/Map';
-import MapEventType from 'ol/MapEventType';
 import { transformExtent } from 'ol/proj';
-import Source from 'ol/source/Source';
-import { MapOptions, MapOptionType, WindowUrl } from './url';
 
-const Copyright = `© ${StacLicense} LINZ`;
+/** html escape function */
+function escapeHtml(text: string): string {
```
blacha, line 10: This is a cool way of doing this!

blacha (approved):

PR #1240, opened 2020-10-04
feat(landing): be clear about 90 day API key

jacott:


packages/landing/static/index.css
```diff
@@ -167,6 +167,10 @@ button {
     transition: right 0.5s;
 }
 
+#side-nav>h3 {
```
blacha, line 170: This seems to be a different format to everything else, do we have a css formatter somewhere

jacott: fixed the indentation

blacha (approved):

PR #1250, opened 2020-10-07
fix(landing) Deploy mime type wrong

jacott:
When deploying changed files the logic to determine the mime-type was returning the wrong
type. Fixed by passing the file extension instead of whole path.

Also deploy `examples/index.html` as `examples` and `examples/` because Cloudfront doesn't handle `index.html`
files in sub directories.

### DIFF
--- a/packages/landing/scripts/deploy.js
+++ b/packages/landing/scripts/deploy.js
@@ -3,6 +3,7 @@ const AWS = require('aws-sdk');
 const fs = require('fs').promises;
 const crypto = require('crypto');
 const mime = require('mime-types');
+const { extname, basename } = require('path');
 const invalidateCache = require('@basemaps/cli/build/cli/util').invalidateCache;
 const { recurseDirectory } = require('./util');
 
@@ -23,6 +24,9 @@ async function getHash(Bucket, Key) {
     return null;
 }
 
+// match a string containing a version number
+const HasVersionRe = /-\d+\.\d+\.\d+-/;
+
 /**
  * Deploy the built s3 assets into the Edge bucket
  *
@@ -33,6 +37,7 @@ async function deploy() {
     const stackInfo = await cf.describeStacks({ StackName: 'Edge' }).promise();
     const bucket = stackInfo.Stacks[0].Outputs.find((f) => f.OutputKey == 'CloudFrontBucket');
     if (bucket == null) throw new Error('Failed to find EdgeBucket');
+    console.log(`Upload to s3 bucket "${bucket.OutputValue}"`);
 
     const s3BucketName = bucket.OutputValue;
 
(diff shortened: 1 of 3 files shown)


packages/landing/scripts/deploy.js
```diff
@@ -33,55 +34,65 @@ async function deploy() {
...
+            console.log('Skipped', filePath, `${(fileData.byteLength / 1024).toFixed(2)}Kb`, hash);
             return true;
-        });
-    };
+        }
+        const contentType = mime.contentType(extname(filePath));
 
-    await uploadDirectory();
+        // Don't set cache control for index.html as it may change all other files are hashed and immutable
+        const cacheControl = filePath.endsWith('.html') ? undefined : 'public, max-age=604800, immutable';
```
jacott, line 79: Comment needs to change to "...for html files as they may change..."

I was wondering if we should set cache control for html files to `public, max-age=300, stale-while-revalidate=86400` which is fresh for 5 minutes then use max one day but reload then must reload.

blacha: I don't think the `index.html` gets hit that often so I dont think its a huge issue to keep it somewhat fresh.

Quickly reading the docs on `stale-while-revalidate` I think It would be best to keep the max-age quite low eg `60` seconds or less and maybe keep the `stale-while-revalidate` at say 5 minutes?

AKA I dont want to wait up to 5 minutes for a index.html change to go through.

packages/landing/scripts/deploy.js
```diff
@@ -33,55 +34,68 @@ async function deploy() {
...
+            // dirname+'/' also
+            await upload(filePath.replace('/index.html', ''), fileData, contentType, cacheControl, hash, oldHash);
+            await upload(filePath.replace('/index.html', '/'), fileData, contentType, cacheControl, hash, oldHash);
+        }
+
+        return true;
+    });
+
+    for (const filePath of changes) {
+        await invalidateCache('/' + filePath, true);
```
blacha, line 97: invalidations get quite pricey, we should batch this up into a single invalidation.

packages/landing/scripts/deploy.js
```diff
@@ -33,55 +34,68 @@ async function deploy() {
...
+        // hashed and immutable
+        const cacheControl = filePath.endsWith('.html') ? undefined : 'public, max-age=604800, immutable';
 
-    if (hasChanges) await invalidateCache('/index.html', true);
+        await upload(filePath, fileData, contentType, cacheControl, hash, oldHash);
+
+        if (filePath.endsWith('/index.html')) {
+            // s3 does not handle subdirectory index.html so need to upload to dirname and
+            // dirname+'/' also
+            await upload(filePath.replace('/index.html', ''), fileData, contentType, cacheControl, hash, oldHash);
```
blacha, line 89: what happens for the root file eg `index.html` does it try to upload a empty key?

jacott: The root index.html is handled correctly it is only sub directories where the problem exists

packages/landing/scripts/deploy.js
jacott: Will now only invalidate at most `/index.html` and `/examples*`

packages/landing/scripts/deploy.js
jacott: I am changing the logic so that all files containing a version number are marked immutable `public, max-age=604800, immutable` and all other files have a short lived cache control `public, max-age=60, stale-while-revalidate=300`

Only `index.html` will be invalidated from the cache. Arguably this is not needed either.

blacha (approved):

PR #1277, opened 2020-10-22
feat(cli): Configure TileSet metedata DB from config file

jacott:
Adds import, export commands to basemaps tool

Supports both Head and Production config and full DB contents config.

### DIFF
--- a/packages/cli/src/cli/basemaps/action.export.ts
+++ b/packages/cli/src/cli/basemaps/action.export.ts
@@ -1,0 +1,122 @@
+import { Epsg } from '@basemaps/geo';
+import {
+    Aws,
+    DefaultBackground,
+    LogConfig,
+    parseMetadataTag,
+    RecordPrefix,
+    TileMetadataTable,
+    TileMetadataTag,
+} from '@basemaps/shared';
+import { CommandLineStringParameter } from '@rushstack/ts-command-line';
+import { promises as fs } from 'fs';
+import { TileSetBaseAction } from './tileset.action';
+import {
+    compareNamePriority,
+    FullImageryConfig,
+    ImageryDefaultConfig,
+    ProjectionConfig,
+    removeDefaults,
+} from './tileset.config';
+import { defineTagParameter, primeImageryCache, rgbaToHex } from './tileset.util';
+
+/**
+ * Convert all the tilesets in the TileMetadataTable to a TileSetConfig. The Head and Production
+ * rules tags are converted to one record. If a production rule differs from the head then the rule
+ * is marked as non-production.
+
+ * @param table
+ */
+async function tilesetToConfig(
+    name: string,
+    projection: Epsg,
+    tag: TileMetadataTag,
+    defaults: ImageryDefaultConfig[] = [],
+): Promise<ProjectionConfig> {
+    const item = await Aws.tileMetadata.TileSet.get(name, projection, tag);
+
+    const imageryMap = await primeImageryCache(new Set(item.rules.map((r) => r.imgId)));
+
+    const imagery = item.rules
+        .map(
+            (rule): FullImageryConfig => {
+                const image = imageryMap.get(rule.imgId);
+                if (image == null) {
+                    throw new Error(`Can't find imagery record "${rule.imgId}" for Tileset "${item.id}"`);
+                }
+                return {
+                    id: TileMetadataTable.unprefix(RecordPrefix.Imagery, rule.imgId),
+                    name: image.name,
+                    priority: rule.priority,
+                    minZoom: rule.minZoom,
+                    maxZoom: rule.maxZoom,
+                };
+            },
+        )
+        .sort(compareNamePriority)
+        .map((r) => removeDefaults(defaults, r));
+
+    return {
+        name,
+        projection: projection.code,
+        background: rgbaToHex(item.background ?? DefaultBackground),
+        defaults,
+        imagery,
+    };
+}
+
+export class ExportAction extends TileSetBaseAction {
+    private output: CommandLineStringParameter;
+    private defaults: CommandLineStringParameter;
+    private tag: CommandLineStringParameter;
+
+    public constructor() {
+        super({
+            actionName: 'export',
+            summary: 'Export imagery and tileset rules',
+            documentation: '',
+        });
+    }
+
+    protected onDefineParameters(): void {
+        super.onDefineParameters();
+        this.output = this.defineStringParameter({
+            argumentName: 'JSON_FILE',
+            parameterLongName: '--output',
+            parameterShortName: '-o',
+            description: 'File to output to',
+            required: true,
+        });
+        this.defaults = this.defineStringParameter({
+            argumentName: 'JSON_FILE',
+            parameterLongName: '--defaults',
+            description:
+                'Config File containing rule defaults. This is added to the output and used to minimize the rule fields.',
+            required: false,
+        });
+        defineTagParameter(this);
+    }
+
+    protected async onExecute(): Promise<void> {
+        const outFile = this.output.value!;
+
+        const tileSet = this.tileSet.value!;
+        const projection = Epsg.get(this.projection.value!);
+        const tagInput = this.tag.value!;
+
+        const tag = parseMetadataTag(tagInput);
+        if (tag == null) {
+            LogConfig.get().fatal({ tag }, 'Invalid tag name');
+            console.log(this.renderHelpText());
+            return;
+        }
+
+        const defaults = this.defaults.value
+            ? JSON.parse((await fs.readFile(this.defaults.value)).toString())
+            : undefined;
+
+        const output = JSON.stringify(await tilesetToConfig(tileSet, projection, tag, defaults), undefined, 2);
+
+        await fs.writeFile(outFile, output, { flag: 'wx' });
+    }
+}
--- a/packages/cli/src/cli/basemaps/action.import.ts
+++ b/packages/cli/src/cli/basemaps/action.import.ts
@@ -1,11 +1,24 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { Epsg } from '@basemaps/geo';
-import { Aws, LogConfig, RecordPrefix, TileMetadataImageryRecord, TileMetadataTable } from '@basemaps/shared';
+import {
+    Aws,
+    LogConfig,
+    parseMetadataTag,
+    RecordPrefix,
+    TileMetadataImageryRecord,
+    TileMetadataTable,
+} from '@basemaps/shared';
 import { CommandLineAction, CommandLineFlagParameter, CommandLineStringParameter } from '@rushstack/ts-command-line';
-import { createImageryRecordFromJob } from '../cogify/action.batch';
 import { CogStacJob } from '../../cog/cog.stac.job';
+import { createImageryRecordFromJob } from '../cogify/action.batch';
+import { updateConfig } from './tileset.updater';
+import { defineTagParameter } from './tileset.util';
 
-export class ImageryImportAction extends CommandLineAction {
+/**
+ * Import a config file for a specific name and projection
+ */
+export class ImportAction extends CommandLineAction {
+    private config: CommandLineStringParameter;
+    private tag: CommandLineStringParameter;
     private job: CommandLineStringParameter;
     private commit: CommandLineFlagParameter;
     private force: CommandLineFlagParameter;
@@ -13,23 +26,33 @@ export class ImageryImportAction extends CommandLineAction {
     public constructor() {
         super({
             actionName: 'import',
-            summary: 'Import imagery from job.json',
+            summary: 'Import imagery and tileset rules from a config file"',
             documentation: '',
         });
     }
 
     protected onDefineParameters(): void {
+        this.config = this.defineStringParameter({
+            argumentName: 'JSON_FILE',
+            parameterLongName: '--config',
+            parameterShortName: '-c',
+            description: 'Configure tilesets using json file. May not be used with --job option)',
+            required: false,
+        });
+
+        defineTagParameter(this);
+
         this.job = this.defineStringParameter({
             argumentName: 'JOB',
             parameterLongName: '--job',
             parameterShortName: '-j',
-            description: 'Path to job.json',
-            required: true,
+            description: 'Import imagery from a job.json',
+            required: false,
         });
 
         this.force = this.defineFlagParameter({
             parameterLongName: '--force',
-            description: 'Force overwrite',
+            description: 'Force overwrite. Only for use with --job',
             required: false,
         });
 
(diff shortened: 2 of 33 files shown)


packages/cli/src/cli/basemaps/action.export.ts
```diff
@@ -0,0 +1,180 @@
+import { EpsgCode } from '@basemaps/geo';
+import {
+    DefaultBackground,
+    RecordPrefix,
+    TileMetadataImageRule,
+    TileMetadataSetRecord,
+    TileMetadataTable,
+    TileMetadataTag,
+} from '@basemaps/shared';
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
blacha, line 10: we should not be using non-null-assertions

packages/cli/src/cli/basemaps/action.export.ts
```diff
@@ -0,0 +1,180 @@
...
+            description: 'Export the entire tilemetadatdb',
+            required: false,
+        });
+    }
+
+    protected async onExecute(): Promise<void> {
+        const outFile = this.output.value!;
+
+        try {
+            await fs.stat(outFile);
```
blacha, line 159: stat then write is not a good idea as it causes race conditions, your better off doing a writeFile with exception if it exists 

`flag: "wx"`

packages/cli/src/cli/basemaps/action.import.ts
```diff
@@ -0,0 +1,224 @@
...
+    if (backgroundBefore !== backgroundAfter) {
+        output += `\nBackground changed from '${backgroundBefore}' to '${backgroundAfter}'\n`;
+    }
+
+    output += showDiff(changes.after, changes.before, imagery);
+    await invalidateXYZCache(changes.name, changes.projection, tag, isCommit);
+    return output;
+}
+
+async function clearDB(table: TileMetadataTable): Promise<void> {
```
blacha, line 41: This seems like a super dangerous function to have, we should never be deleting from the database.

packages/cli/src/cli/basemaps/tileset.util.ts
```diff
@@ -10,6 +10,42 @@ import * as c from 'ansi-colors';
...
+    else if (str.startsWith('#')) str = str.slice(1);
+    return {
+        r: parseHex(str.substr(0, 2)),
+        g: parseHex(str.substr(2, 2)),
+        b: parseHex(str.substr(4, 2)),
+        alpha: parseHex(str.substr(6, 2)),
+    };
+}
+
+function nTohh(n: number): string {
```
blacha, line 41: what is a hh

packages/shared/src/aws/tile.metadata.base.ts
```diff
@@ -190,6 +193,55 @@ export class TileMetadataTableBase {
...
+
+            exclusiveStartKey = items.LastEvaluatedKey;
+            if (exclusiveStartKey == null) break;
+        }
+        return null;
+    }
+
+    /**
+     * Write an array of write requests to the table.
+
```
blacha, line 223: this is weirdly formatted

packages/shared/src/aws/tile.metadata.ts
```diff
@@ -14,4 +16,16 @@ export class TileMetadataTable extends TileMetadataTableBase {
...
         this.Provider = new TileMetadataProvider(this);
     }
+
+    /**
+     * Iterate over all records in the TileMetadataTable
+     */
+    async *[Symbol.asyncIterator](): AsyncGenerator<TileMetadataRecord, null, void> {
+        for await (const items of this.scan(Const.TileMetadata.TableName)) {
+            for (const row of items) {
+                yield DynamoDB.Converter.unmarshall(row) as any;
```
blacha, line 26: Are there any records in here that are not `TileMetadataRecords` (Should we filter this some how)

packages/shared/src/aws/tile.metadata.base.ts
```diff
@@ -190,6 +193,55 @@ export class TileMetadataTableBase {
...
+     * @param tableName the name of the table to write to
+
+     * @param input the array of write requests
+
+     * @param maxBatchSize the maximum number of write requests to send in one command
+     */
+    async batchWrite(tableName: string, input: DynamoDB.WriteRequests, maxBatchSize = 25): Promise<void> {
```
blacha, line 230: this should use a `T` in the input

packages/shared/src/aws/tile.metadata.base.ts
```diff
@@ -190,6 +193,55 @@ export class TileMetadataTableBase {
...
+            const items: DynamoDB.Types.ScanOutput = await this.dynamo
+                .scan({
+                    TableName,
+                    ExclusiveStartKey: exclusiveStartKey,
+                })
+                .promise();
+
+            if (items.Items == null) break;
+
+            yield items.Items;
```
blacha, line 213: Would be better to yield these one by one, and include the type of the base object.

This should also do a migrate?

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,59 @@
+import * as z from 'zod';
+
+const zImageryConfig = z.object({
+    imgId: z.string(),
+    name: z.string(),
+    priority: z.number().refine((val) => val >= -1 && val <= 10000, { message: 'must be between -1 and 10000' }),
+    zoom: z
```
blacha, line 7: Why is zoom a array? all other configs have min/max?

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,59 @@
...
+            message: 'range must be within [0, 32]',
+        }),
+    production: z.boolean().optional(),
+});
+
+/**
+ * The projections supported by TileSetConfig
+ */
+export enum ProjectionCode {
+    Google = 'EPSG:3857',
```
blacha, line 21: Could these be extracted from somewhere else?

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,59 @@
...
+    Nztm2000 = 'EPSG:2193',
+}
+
+const zBackground = z
+    .string()
+    .refine((val) => /^[0-9a-f]{8}$/.test(val), { message: 'must have 8 lower case hex characters' });
```
blacha, line 27: would be safer to just parse this and make sure it actually works

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const zProjectionConfig = z.object({
+    headBackground: zBackground,
```
blacha, line 30: what about all the other tags?

blacha (changes requested):

packages/cli/src/cli/basemaps/action.export.ts
jacott: Cool that is what I was looking for :)

packages/cli/src/cli/basemaps/action.import.ts
jacott: I have moved it into the `loadFull` function but this is needed to populate dev with production or to restore production if something goes wrong. This is ported code from the basemaps-team repository

packages/shared/src/aws/tile.metadata.base.ts
jacott: much more useful to me the way I have written it

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: The parse function is super lax so not safer

packages/cli/src/cli/basemaps/action.export.ts
```diff
@@ -0,0 +1,141 @@
...
+    table: TileMetadataTable,
+    name: string,
+    projection: Epsg,
+    tag: TileMetadataTag,
+    defaults: ImageryDefaultConfig[] = [],
+): Promise<ProjectionConfig> {
+    const item = await Aws.tileMetadata.TileSet.get(name, projection, tag);
+    // assign an array slot for head and production records; head = 0, prod = 1
+    if (item.rules == null) {
+        table.TileSet.migrate(item as any); // migrate old format
```
blacha, line 39: Why is `any` needed here?

packages/cli/src/cli/basemaps/action.export.ts
```diff
@@ -0,0 +1,141 @@
...
+        const outFile = this.output.value!;
+        const table = new TileMetadataTable();
+
+        let output: string;
+        if (this.full.value) {
+            const lines: string[] = [];
+            for await (const item of table) {
+                lines.push(JSON.stringify(item));
+            }
+            output = `[\n  ${lines.join(',\n  ')}\n]`;
```
blacha, line 120: why are you string bashing JSON? cant you just `JSON.stringify(lines)` if you remove the internal stringify?

packages/cli/src/cli/basemaps/tileset.config.ts
blacha: Safer as in it actually matches what the parser does, this already would fail on things that use to work eg `0x000000` or `#FF00FF`

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,127 @@
...
+    minZoom: number;
+    maxZoom: number;
+}
+
+export type ImageryDefaultConfig = z.infer<typeof zImageryDefaultConfig>;
+
+const defaultFields = ['priority', 'minZoom', 'maxZoom'];
+
+export function addDefaults(defaults: ImageryDefaultConfig[], rule: ImageryConfig): FullImageryConfig {
+    const ans = { ...rule } as any;
```
blacha, line 80: Can this be done with out the any

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,127 @@
...
+export function removeDefaults(defaults: ImageryDefaultConfig[], rule: FullImageryConfig): ImageryConfig {
+    const removeFields = new Map<string, boolean>();
+    for (const def of defaults) {
+        if (def.nameContains != null && rule.name.indexOf(def.nameContains) == -1) continue;
+
+        for (const field of defaultFields) {
+            if (field in def) {
+                if (removeFields.has(field)) continue;
+
+                removeFields.set(field, (def as any)[field] == (rule as any)[field]);
```
blacha, line 110: remove any's

packages/cli/src/cli/basemaps/action.import.ts
blacha: I disagree, we should never be wiping a database with a script, this has too many ways of going wrong.

Can we just create a new head record and insert it? I don't see why we would ever need to delete the old records?

packages/cli/src/cli/basemaps/tileset.updater.ts
```diff
@@ -0,0 +1,239 @@
...
+
+function objectsDiffer(a: any, b: any): boolean {
+    try {
+        deepStrictEqual(a, b);
+        return false;
+    } catch (err) {}
+    return true;
+}
+
+function initiaTs(name: string, projection: Epsg): TileMetadataSetRecord {
```
blacha, line 42: `initial`?

packages/cli/src/cli/basemaps/tileset.updater.ts
```diff
@@ -0,0 +1,239 @@
...
+    };
+}
+
+async function loadTS(name: string, projection: Epsg, tag: TileMetadataTag): Promise<TileMetadataSetRecord> {
+    const tsData = await Aws.tileMetadata.TileSet.get(name, projection, tag);
+    if (tsData != null) return tsData;
+    return initiaTs(name, projection);
+}
+
+interface ImgPri {
```
blacha, line 63: `ImgPriority`

packages/cli/src/cli/basemaps/tileset.updater.ts
blacha: This should really live inside the TileSet database class as we use something similar somewhere else?

packages/shared/src/aws/tile.metadata.tileset.ts
```diff
@@ -46,6 +47,15 @@ export class TileMetadataTileSet extends TaggedTileMetadata<TileMetadataSetRecor
         return output;
     }
 
+    /**
+     * Is `rec` a TileSet record
+
+     * @param rec record to infer is a TileMetadataSetRecord
+     */
+    recordIsATileset(rec: BaseDynamoTable): rec is TileMetadataSetRecord {
```
blacha, line 55: Why does this include `A` but the other one doesnt?

I think we also generally use `TileSet`

packages/cli/src/cli/basemaps/action.import.ts
jacott: Okay I'll remove the full option

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: The parse function accepts `'1anyting'` I'll make the parse function more robust

packages/cli/src/cli/basemaps/action.export.ts
jacott: Otherwise just one long line but don't want to prettify the whole thing as then too many lines. All is mute because I am removing the `full` option

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: I can't see how without using hardcoded if statements but if you know how I am all ears.

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: how?

packages/shared/src/aws/tile.metadata.base.ts
jacott: I don't understand? template input? Only `WriteRequests` are allowed.

packages/shared/src/aws/tile.metadata.base.ts
jacott: No longer needed as is because `export --full` removed

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: I have made sure that the unit tests will pick up any deviation from the intended types

packages/shared/src/aws/tile.metadata.base.ts
jacott: This is no longer needed

packages/cli/src/cli/basemaps/tileset.config.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const zZoom = z.number().refine((val) => val >= ImageryConfigDefaults.minZoom && val <= ImageryConfigDefaults.maxZoom, {
+    message: `must be between ${ImageryConfigDefaults.minZoom} and ${ImageryConfigDefaults.maxZoom}`,
+});
+
+const zImageryRule = z
+    .object({
+        priority: z
+            .number()
+            .refine((val) => val >= -1 && val <= 10000, { message: 'must be between -1 and 10000' })
```
blacha, line 19: What does a priority of `-1` mean?

blacha (approved):

packages/cli/src/cli/basemaps/tileset.config.ts
jacott: -1 is turn off as it is in the other tile metadata commands

blacha: Do we not delete it by just removing it from the config file?

PR #1283, opened 2020-10-28
feat(lambda-tiler): smoke test during deployment

Wentao-Kuang:
When convenient, please have a look the changes and feedback.


packages/lambda-tiler/package.json
```diff
@@ -22,6 +22,7 @@
     "@cogeotiff/core": "^2.2.0",
     "@cogeotiff/source-aws": "^2.2.0",
     "@linzjs/geojson": "^4.16.0",
+    "node-fetch": "^2.6.1",
```
blacha, line 25: This should be a devDependency as it is not needed at runtime.

Wentao-Kuang: Good pick.

jacott: But isn't it needed at runtime because the test is run at runtime?

jacott (commented):
What about a test for `api.ts`?

PR #1309, opened 2020-11-03
build: reduce build time

blacha:
Remove base imports of `aws-sdk` as it greatly slows down the typescript build time


package.json
```diff
@@ -15,6 +15,7 @@
     "benchmark": "ospec 'packages/*/build/**/*.benchmark.js'",
     "clean": "tsc -b --clean && rimraf 'packages/*/build' && yarn clean:dist",
     "clean:dist": "rimraf 'packages/*/dist' && rimraf 'packages/_infra/cdk.out'",
+    "clean:modules": " rimraf 'packages/*/node_modules' && rimraf node_modules",
```
jacott, line 18: nitpick: `" rimraf...` space after quote

packages/cli/src/cog/__test__/builder.test.ts
```diff
@@ -39,7 +39,7 @@ o.spec('Builder', () => {
 
         o('bounds', async () => {
             const localTiff = new CogSourceFile('/local/file.tiff');
-            const s3Tiff = new CogSourceAwsS3('bucket', 's3://file.tiff');
+            const s3Tiff = new CogSourceAwsS3('bucket', 's3://file.tiff', null as any);
```
jacott, line 42: Maybe a comment as to why this is needed?

blacha: This was a API change to force you to push in a S3 object into the sources, this test overrides the reading of the tiffs.

if this was created with `CogSourceAwsS3.create` you dont need the `as any`

jacott (approved):

PR #1311, opened 2020-11-04
build(deps): update ol from 6.3.1 to 6.4.2

jacott:


packages/landing/src/map.ts
```diff
@@ -184,7 +186,7 @@ export class Basemaps {
             .transform(center, `EPSG:${this.config.projection}`, UrlProjection)
             .map((c: number) => Math.floor(c * 10e7) / 10e7); // Limit to 5 decimal places (TODO is 5 enough)
 
-        const zoom = Math.floor(view.getZoom() * 10e3) / 10e3;
+        const zoom = Math.floor(view.getZoom() ?? 0 * 10e3) / 10e3;
```
blacha, line 189: how does `??` relate to `*`

Wentao-Kuang:
All looks good to me.

packages/landing/src/ol.attribution.ts
```diff
@@ -43,8 +43,9 @@ export class OlAttribution {
     static init(source: Source, map: OlMap, config: MapOptions): void {
         const attribution = new OlAttribution(source, map.getView(), config);
 
-        map.addEventListener(MapEventType.MOVEEND, () => {
+        map.addEventListener(MapEventType.MOVEEND, (): boolean => {
             attribution.updateAttribution();
+            return true;
```
blacha, line 48: what does the return true do here?

packages/landing/src/map.ts
jacott: badly by the looks of the transpiled code! Well spotted!!

packages/landing/src/ol.attribution.ts
jacott: Continues the event propigation

blacha (approved):

PR #1319, opened 2020-11-05
feat(cli): Allow basemaps import to change title and description

jacott:
as well as background and resizeKernel

### DIFF
--- a/packages/cli/src/cli/basemaps/tileset.updater.ts
+++ b/packages/cli/src/cli/basemaps/tileset.updater.ts
@@ -7,6 +7,7 @@ import {
     TileMetadataImageRule,
     TileMetadataImageryRecord,
     TileMetadataNamedTag,
+    TileMetadataSetConfig,
     TileMetadataSetRecord,
     TileMetadataTable,
     TileMetadataTag,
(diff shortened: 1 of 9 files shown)


packages/cli/src/cli/basemaps/tileset.updater.ts
```diff
@@ -126,6 +144,21 @@ export async function updateConfig(filename: string, tag: TileMetadataTag, isCom
     }
 }
 
+function extractSetConfig(a: TileMetadataSetRecord | ProjectionConfig): TileMetadataSetConfig {
+    const ans: any = {};
```
blacha, line 148: Might be better to have this as a `Partial<TileMetadataSetConfig>`

jacott: That doesn't appear to work:

```
Type 'string | RGBAColor | { in: TileResizeKernel; out: TileResizeKernel; } | { in: string; out: string; }' is not assignable to type '(string & RGBAColor & { in: TileResizeKernel; out: TileResizeKernel; }) | undefined'.
  Type 'string' is not assignable to type '(string & RGBAColor & { in: TileResizeKernel; out: TileResizeKernel; }) | undefined'. [2322]
```

PR #1321, opened 2020-11-08
feat: Allow alternative TileMatrixSet definitions

jacott:
Esri™ does not support adding layers to basemaps which use a different tile matrix set. ArcGIS™
online provides basemaps for New Zealand which use a different tile matrix set to the NZTM2000 one
used by LINZ Basemaps. To allow LINZ Basemaps to be used as Layers with ArcGIS™ online basemaps a
compatible tiling scheme is included which is accessed from
`https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:2193:agol/WMTSCapabilities.xml?api={API_KEY}`


packages/lambda-tiler/src/wmts.capability.ts
```diff
@@ -174,7 +160,6 @@ export class WmtsCapabilities {
             V('ows:Title', tms.def.title),
             tms.def.abstract ? V('ows:Abstract', tms.def.abstract) : null,
             V('ows:Identifier', tms.projection.toEpsgString()),
-            this.buildBoundingBox(tms, tms.extent),
             V('ows:SupportedCRS', tms.projection.toUrn()),
```
jacott, line 163: BoundingBox removed for TileMatrixSet as this was breaking in ArcGIS™ Online

packages/lambda-tiler/src/cli/serve.ts
```diff
@@ -39,6 +40,7 @@ async function handleRequest(
         const duration = Date.now() - startTime;
         logger.info({ ...ctx.logContext, ...logInfo, status: data.status, duration }, 'Done');
     } catch (e) {
+        console.error(e);
```
blacha, line 43: We shouldn't use `console.log`

packages/shared/src/alternative.tms/alternative.tms.ts
```diff
@@ -0,0 +1,48 @@
...
+                    }
+                }
+            }
+            last = pTmx;
+        }
+    }
+
+    getParentZoom(z: number): number {
+        const convertZ = this.scaleMap.get(z);
+        if (convertZ == null) throw new Error(`No converted scaled level from parent Tile Matrix`);
```
blacha, line 41: should include the `z` in the error otherwise its hard to know what went wrong.

packages/shared/src/alternative.tms/alternative.tms.ts
```diff
@@ -0,0 +1,48 @@
...
+
+        // Setup the z mapping between child Tmx and parent Tmx based on the scaleDenominator
+        let last = null;
+        for (const pTmx of this.parent.def.tileMatrix) {
+            for (const cTmx of this.def.tileMatrix) {
+                if (last == null) {
+                    if (cTmx.scaleDenominator >= pTmx.scaleDenominator) {
+                        this.scaleMap.set(Number(cTmx.identifier), Number(pTmx.identifier));
+                    }
+                } else if (this.parent.def.tileMatrix.indexOf(pTmx) == this.parent.def.tileMatrix.length - 1) {
```
blacha, line 22: if you need to do `indexOf`, its best to switch this loop into  `for (let i = 0; i < ...) { ` style as `indexOf` is not a cheap operation

blacha: Might be a idea to move this into a standalone function too, that returns a map, from two tile matrix sets

blacha (approved):

PR #1351, opened 2020-11-24
feat(shared): Add iterator into TileMetadataTileSet.

Wentao-Kuang:
I have added an iterator and splitId() into TileMetadataTileSet for vacuum imagery to use. Please have a look. Thanks.

### DIFF
--- a/packages/shared/src/aws/__test__/tile.metadata.tileset.test.ts
+++ b/packages/shared/src/aws/__test__/tile.metadata.tileset.test.ts
@@ -1,5 +1,7 @@
 import { Epsg } from '@basemaps/geo';
+import DynamoDB from 'aws-sdk/clients/dynamodb';
 import o from 'ospec';
+import { Const } from '../../const';
 import { TileSetName } from '../../proj/tile.set.name';
 import { BaseDynamoTable } from '../aws.dynamo.table';
 import { TileMetadataTable } from '../tile.metadata';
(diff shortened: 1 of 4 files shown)


packages/shared/src/aws/tile.metadata.tileset.ts
```diff
@@ -137,6 +140,14 @@ export class TileMetadataTileSet extends TaggedTileMetadata<TileMetadataSetRecor
         return `ts_${record.name}_${record.projection}_${tag}`;
     }
 
+    idSplit(record: TileMetadataSetRecord): { name: string; projection: string; tag: string } | null {
```
blacha, line 143: I dont think we should leave projection as `string` we should use `projection: Epsg` or `projection: EpsgCode` which is the result of the  `Epsg.parse`

packages/shared/src/aws/__test__/tile.metadata.tileset.test.ts
```diff
@@ -97,17 +99,30 @@ o.spec('tile.metadata.tileset', () => {
...
-        o('should create version ids', () => {
-            o(ts.id('test', Epsg.Google, 5)).equals('ts_test_3857_v000005');
-            o(ts.id('test', Epsg.Google, 32)).equals('ts_test_3857_v000032');
-            o(ts.id('test', Epsg.Google, Number.MAX_SAFE_INTEGER)).equals('ts_test_3857_v9007199254740991');
+    o.spec('idSplit', () => {
+        o('should return null for unexpected ids', () => {
+            o(ts.idSplit({ id: 'wrongId' } as any)).equals(null);
+            o(ts.idSplit({ id: 'im_test_3857_head' } as any)).equals(null);
+            o(ts.idSplit({ id: 'ts_test_0000_production' } as any)).equals(null);
+            o(ts.idSplit({ id: 'ts_test_3857_v000006' } as any)).equals(null);
```
blacha, line 107: why does this fail?

packages/shared/src/aws/tile.metadata.tileset.ts
```diff
@@ -137,6 +140,14 @@ export class TileMetadataTileSet extends TaggedTileMetadata<TileMetadataSetRecor
...
+        const [prefix, name, projection, tag] = record.id.split('_');
+        if (prefix != 'ts') return null;
+        if (!Epsg.parse(projection)) return null;
+        if (!parseMetadataTag(tag)) return null;
+        return { name, projection, tag };
```
blacha, line 148: do we also want to parse `v000006` into `{ name: string, projection: .., version: number (6)}`

Wentao-Kuang: I believe for the vacuum imagery script, we only need the production metadata. However, it seem good to parse versions here as well. I will make the changes.

blacha (approved):

PR #1409, opened 2021-02-01
feat(tiler): support rendering avif tiles

blacha:


Wentao-Kuang (approved):
Looks good, and avif works on my local serve.
Thanks.

PR #1414, opened 2021-02-02
fix(landing): correct broken wmts/xyz links in side bar

blacha:
by default process.env.TILE_HOST is set to "" not undefined


Wentao-Kuang (approved):
Looks good. Thank you.

PR #1437, opened 2021-02-15
build(deps): cogeotiff 4.0.0

blacha:


blacha:
This is in preperation for when `@cogeotiff/core v4.0.0` is released, should have some nice performance gains for tile fetching.

packages/cli/src/cog/builder.ts
```diff
@@ -148,16 +146,16 @@ export class CogBuilder {
             return Epsg.get(projection);
         }
 
-        const imgWkt = image.value(TiffTag.GeoAsciiParams);
+        const imgWkt = image.value<string>(TiffTag.GeoAsciiParams);
```
Wentao-Kuang, line 149: Sorry, I am not fully understand what is happening here. Why we add ! for `imgWkt!` after specify the `string` type.
!`

blacha: `.value` can return a `string | number | string[] | number[] | undefined`  based off the TiffTag requested.

by doing `<string>` I am forcing the type to be `string | undefined` 

I should really get rid of the `!` though, Ill have another look at this logic

Wentao-Kuang (approved):

PR #1438, opened 2021-02-16
feat(lambda): log the lambda request id at the end of every request

blacha:


Wentao-Kuang (approved):
Nice and easy. Thanks.

PR #1439, opened 2021-02-16
feat(lambda-tiler): show number of bytes served with WMTS requests

blacha:


packages/lambda-tiler/src/routes/tile.ts
```diff
@@ -164,6 +164,7 @@ export async function wmts(req: LambdaContext): Promise<LambdaHttpResponse> {
     response.header(HttpHeader.ETag, cacheKey);
     response.header(HttpHeader.CacheControl, 'max-age=0');
     response.buffer(data, 'text/xml');
+    req.set('bytes', data.byteLength);
```
Wentao-Kuang, line 167: Does it suppose to set to response?

blacha: This `set` set's the request context.

The request context is logged at the end of the request.

Wentao-Kuang: I see. Now good to go.

Wentao-Kuang (approved):

PR #1440, opened 2021-02-16
fix(lambda-tiler): only export the tile matrix set once per epsg code

blacha:
This was creating a TileMatrixSet definition per use which was causing the file size to explode when multiple layers were used as each layer would export a copy of each TileMatrixSet they used.


Wentao-Kuang (approved):
Looks all good. Thank you.

PR #1467, opened 2021-03-03
refactor: remove alternative tile matrix sets

blacha:
Fixes: #


packages/shared/src/proj/projection.tile.matrix.set.ts
```diff
@@ -59,11 +53,8 @@ export class ProjectionTileMatrixSet {
      * @param epsgCode
```
Wentao-Kuang, line 53: Do we still need this?
`static altMap = new Map<EpsgCode, Map<string, ProjectionTileMatrixSet>>();`

packages/lambda-tiler/src/tiler.ts
```diff
@@ -1,10 +1,9 @@
 import { Epsg, TileMatrixSet } from '@basemaps/geo';
 import { GoogleTms } from '@basemaps/geo/build/tms/google';
 import { Nztm2000Tms } from '@basemaps/geo/build/tms/nztm2000';
-import { Nztm2000AgolTms } from '@basemaps/shared/build/alternative.tms/nztm2000.agol';
 import { Tiler } from '@basemaps/tiler/build/tiler';
 
-export const DefaultTilers = [new Tiler(GoogleTms), new Tiler(Nztm2000Tms), new Tiler(Nztm2000AgolTms)];
+export const DefaultTilers = [new Tiler(GoogleTms), new Tiler(Nztm2000Tms)];
```
Wentao-Kuang, line 6: ` get(epsg: Epsg, alt?: string): Tiler | undefined {
        return Tilers.map.get(TileMatrixSet.getId(epsg, alt));
    },
`
This Tile.get and TileMatrixSet.getId can remove the alt parameter as well.

packages/shared/src/proj/projection.tile.matrix.set.ts
blacha: You are right we should delete this, however it gets deleted when I remove `ProjectionTileMatrixSet` in the next PR :)

Wentao-Kuang (approved):

packages/lambda-tiler/src/tiler.ts
blacha: Tile.get also gets removed in the next PR I think

PR #1474, opened 2021-03-04
feat(lambda-tiler): support NZTM2000Quad when serving via WMTS

blacha:
Fixes: #

### DIFF
--- a/packages/lambda-tiler/src/wmts.capability.ts
+++ b/packages/lambda-tiler/src/wmts.capability.ts
@@ -1,4 +1,4 @@
-import { Bounds, TileMatrixSet, WmtsProvider } from '@basemaps/geo';
+import { Bounds, Nztm2000QuadTms, TileMatrixSet, WmtsProvider } from '@basemaps/geo';
 import { Projection, TileMetadataProviderRecord, V, VNodeElement } from '@basemaps/shared';
 import { ImageFormatOrder } from '@basemaps/tiler';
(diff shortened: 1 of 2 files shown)


packages/lambda-tiler/src/wmts.capability.ts
```diff
@@ -19,6 +19,15 @@ function wgs84Extent(layer: TileSet): BBox {
     return Projection.get(layer.tileMatrix).boundsToWgs84BoundingBox(layer.extent);
 }
 
+/**
+ * Default the tile matrix id to the projection of the TileMatrixSet
+ */
+function getTileMatrixId(tileMatrix: TileMatrixSet): string {
```
blacha, line 25: I am not a huge fan of this, we should maybe change all WMTS's to use the tile matrix id but this would be a user facing change, 

should validate with @JBLINZ if we are ok changing the name and make a specific pull request for it.

Wentao-Kuang: Agree, this seems just a temporary solution.

Wentao-Kuang (approved):

packages/lambda-tiler/src/wmts.capability.ts
blacha: @JBLINZ has agreed to move to the Tile Matrix Id as the layer names I will make another pull request for it.

PR #1475, opened 2021-03-04
feat(bathymetry): support other tile matrix sets for rendering bathymetry

blacha:

### DIFF
--- a/packages/bathymetry/src/index.ts
+++ b/packages/bathymetry/src/index.ts
@@ -1,17 +1,19 @@
 import { BaseCommandLine } from '@basemaps/cli/build/cli/base.cli';
 import { makeTempFolder } from '@basemaps/cli/build/cli/folder';
-import { GoogleTms } from '@basemaps/geo';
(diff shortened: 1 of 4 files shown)


packages/bathymetry/src/index.ts
```diff
@@ -53,6 +62,16 @@ class CreateAction extends CommandLineAction {
                 process.env[Env.Gdal.DockerContainerTag] = 'ubuntu-full-latest';
             }
         }
+        const tileMatrixInput = this.tileMatrix.value ?? GoogleTms.identifier;
+        const tileMatrix = TileMatrixSets.find(tileMatrixInput);
+        if (tileMatrix == null) {
```
Wentao-Kuang, line 67: Do we still need checking here since the default is already `GoogleTms.identifier`

Wentao-Kuang (approved):

packages/bathymetry/src/index.ts
blacha: `.value`  is `string | undefined` so typescript wouldn't let me pass it to the `.find()` method

PR #1483, opened 2021-03-09
feat(cli): allow cusomising the tile matrix to use when building COGs

blacha:


packages/cli/src/cog/__test__/cog.test.ts
```diff
@@ -29,11 +28,11 @@ o.spec('cog', () => {
             const job = SourceTiffTestHelper.makeCogJob();
             const logger = LogConfig.get();
 
-            const { targetTms } = job;
+            // const { targetTms } = job;
```
Wentao-Kuang, line 31: Some leftover comments.

Wentao-Kuang (dismissed):

packages/cli/src/cog/__test__/cog.test.ts
blacha: good point ill remove

Wentao-Kuang (approved):

PR #1503, opened 2021-03-16
fix(geo): use the closest zoom mapping

blacha:


packages/lambda-tiler/src/routes/attribution.ts
```diff
@@ -23,6 +25,7 @@ import {
     titleizeImageryName,
 } from '@basemaps/shared';
 import { BBox, MultiPolygon, multiPolygonToWgs84, Pair, union, Wgs84 } from '@linzjs/geojson';
+import 'source-map-support/register';
```
Wentao-Kuang, line 28: Remove this plz

Wentao-Kuang (approved):

PR #1529, opened 2021-03-30
feat(shared): Add support of vector tile metadata table.

Wentao-Kuang:


packages/shared/src/aws/tile.metadata.base.ts
```diff
@@ -154,6 +173,7 @@ export type TileMetadataProviderRecord = TaggedTileMetadataRecord & WmtsProvider
 export enum RecordPrefix {
     Imagery = 'im',
     TileSet = 'ts',
+    Vector = 'vt',
```
blacha, line 176: Vector tiles are still a tile set, they should be a sub group of the tile set

blacha:
This was merged as part of #1535

PR #1548, opened 2021-04-12
refactor: split database configuration out into `@basemaps/config`

blacha:
This moves a lot of configuration around but should simplify the database logic and usage.

Working with record prefixes are now on the config object
```typescript
Config.prefix(Config.Prefix.Imagery, id)
```

Named tag are now on Config
```typescript
Config.Tag.Production
```

Creating a new production tag from record v32
```typescript
const record = await Config.TileSet.get(Config.TileSet.id({name: 'aerial', projection: 3857}, 32);
await Config.Imagery.tag(record, Config.Tag.Production);
```


packages/cli/src/cli/basemaps/action.export.ts
```diff
@@ -31,12 +25,13 @@ import { primeImageryCache, rgbaToHex } from './tileset.util';
 async function tilesetToConfig(
     name: string,
     projection: Epsg,
-    tag: TileMetadataTag,
+    tag: ConfigTag | string,
```
Wentao-Kuang, line 28: Does this also can be `string` now?

packages/cli/src/cli/basemaps/action.tileset.history.ts
```diff
@@ -1,14 +1,9 @@
...
-    Aws,
-    LogConfig,
-    TileMetadataImageryRecord,
-    TileMetadataNamedTag,
-    TileMetadataSetRecord,
-    TileMetadataTag,
-} from '@basemaps/shared';
+import { Config, LogConfig } from '@basemaps/shared';
 import { CommandLineIntegerParameter, CommandLineStringParameter } from '@rushstack/ts-command-line';
+import { ConfigTag } from 'packages/config/src/config/tag';
```
Wentao-Kuang, line 6: `CongiTag` might be better exported from `'@basemaps/config';`

packages/cli/src/cli/basemaps/tileset.util.ts
```diff
@@ -1,15 +1,9 @@
...
-    TileMetadataImageryRecord,
-    TileMetadataNamedTag,
-    TileMetadataSetRecord,
-    TileSetRasterRecord,
-    TileMetadataTag,
-    TileSetType,
-} from '@basemaps/shared';
+import { Config } from '@basemaps/shared';
 import * as c from 'ansi-colors';
+import { ConfigImagery } from 'packages/config/src/config/imagery';
```
Wentao-Kuang, line 5: `ConfigImagery` exported from `@basemap/config`?

packages/cli/src/cli/provider/action.provider.history.ts
```diff
@@ -1,5 +1,7 @@
-import { Aws, TileMetadataNamedTag, TileMetadataProviderRecord, TileMetadataTag } from '@basemaps/shared';
+import { Config } from '@basemaps/shared';
 import { CommandLineAction } from '@rushstack/ts-command-line';
+import { ConfigProvider } from 'packages/config/src/config/provider';
```
Wentao-Kuang, line 3: `ConfigProvider` exported from `@basemaps/config`

packages/cli/src/cog/types.ts
```diff
@@ -1,5 +1,6 @@
 import { BoundingBox, EpsgCode, TileMatrixSet } from '@basemaps/geo';
-import { FileConfig, NamedBounds } from '@basemaps/shared';
+import { FileConfig } from '@basemaps/shared';
+import { NamedBounds } from 'packages/geo/build';
```
Wentao-Kuang, line 3: `NamedBounds` from '@basemaps/geo';

packages/cli/src/cli/basemaps/action.export.ts
blacha: Not sure, we could look at forcing it to always be `ConfigTag` rather than any string. that way a string would have to be parsed before being used.

packages/landing/src/tile.matrix.ts
```diff
@@ -4,7 +4,7 @@ import TileSource from 'ol/source/Tile';
 import WMTS from 'ol/source/WMTS';
 import XYZ from 'ol/source/XYZ';
 import WMTSTileGrid from 'ol/tilegrid/WMTS.js';
-import { Extent } from 'packages/attribution/node_modules/@types/ol/extent';
+// import { Extent } from 'packages/attribution/node_modules/@types/ol/extent';
```
Wentao-Kuang, line 7: comments not delete?

Wentao-Kuang (approved):

PR #1553, opened 2021-04-14
feat(lambda-tiler): serve vector map style json.

Wentao-Kuang:


packages/config/src/config/vector.style.ts
```diff
@@ -1,6 +1,32 @@
...
-    tileSetName: string;
-    style: string;
+export interface Sources {
+    [name: string]: {
+        type: 'vector';
+        url: string;
+    };
 }
+
+export interface StyleJSon {
```
blacha, line 10: `Json`

packages/config/src/config/vector.style.ts
```diff
@@ -1,6 +1,32 @@
 import { VersionedConfig } from './base';
 
-export interface ConfigVectorStyle extends VersionedConfig {
-    tileSetName: string;
-    style: string;
+export interface Sources {
```
blacha, line 3: Generally its best to use `Record<string, { type: 'vector', url: string}>`

packages/lambda-tiler/src/routes/tile.ts
```diff
@@ -107,6 +108,56 @@ export async function tileJson(req: LambdaContext): Promise<LambdaHttpResponse>
...
+
+    // Get style Config from db
+    const dbId = Config.Style.id({ name: style }, nameComp.tag);
+    const styleConfig = await Config.Style.get(dbId);
+    if (styleConfig == null) return NotFound;
+
+    // Prepare sources and add linz source
+    const sources: Sources = {};
+    const tileJsonUrl = `${host}/${version}/${name}/${rest[0]}/${rest[1]}/tile.json?api=${req.apiKey}`;
+    for (const [key, value] of Object.entries(styleConfig.sources)) {
```
blacha, line 125: I'm not 100% sold on this logic, I think we need something more complex here. but maybe this will be enough for what we need to do.

blacha (approved):

PR #1592, opened 2021-05-06
fix(attribution): attribution not working due with incorrect import of Ol Polygon

Wentao-Kuang:
image
This is causing by the incorrect import of `import OlPolygon from 'ol/geom/Polygon';` since the package has a UMD bundle.


blacha:
Something weird is going on here... There looks to be a bunch of bad default imports from open layers and its not just `Polygon` i

Eg GeometryLayout is also having this issue but failing silently

`GeometryLayout_1.default.XY` => `undefined`
`GeometryLayout_1.default.default.XY` => `"XY"`

blacha:
This was fixed in #1599

PR #1650, opened 2021-05-31
feat(landing): Update the UI NZTM button to return WMTS with NZTM2000Quad TileMatrix

Wentao-Kuang:


packages/landing/src/ui.ts
```diff
@@ -161,7 +161,7 @@ Your Service/App URL:
             this.projectionNztm.classList.remove('lui-button-active');
             this.apiXyz.classList.remove('display-none');
         }
-        const cfg: MapOptions = { ...this.basemaps.config, tileMatrix: TileMatrixSets.get(projection) };
+        const cfg: MapOptions = { ...this.basemaps.config, tileMatrix: tileMatrix };
```
blacha, line 164: NIT: you don't need the `: tileMatrix`

Wentao-Kuang: I see, `this.basemaps.config` have tileMatrix already.

Wentao-Kuang: @blacha. Wait, I think need it, it used to prepare the url.

blacha (approved):

PR #1673, opened 2021-06-15
feat(s3fs): provide basic file information with "fs.head"

blacha:


packages/linzjs-s3fs/src/file.local.ts
```diff
@@ -29,8 +39,8 @@ export class FsLocal implements FileProcessor {
         }
     }
 
-    async exists(filePath: string): Promise<boolean> {
-        return await new Promise((resolve) => fs.exists(filePath, resolve));
+    exists(filePath: string): Promise<boolean> {
```
Wentao-Kuang, line 42: Is this suppose to be a async function? and `await` before `head()`?

blacha: since we no longer "await" anything in the function its doest have to be async

Wentao-Kuang (approved):

PR #1691, opened 2021-06-23
feat: switch to a binary cotar index

blacha:


packages/landing/src/ol.attribution.ts
```diff
@@ -18,7 +18,8 @@ function sameExtent(a: Extent, b: Extent): boolean {
     }
     return true;
 }
-
+import { Projection } from '@basemaps/shared';
+console.log(Projection);
```
Wentao-Kuang, line 22: You forgot this.

Wentao-Kuang (approved):

PR #1779, opened 2021-07-23
feat: upgrade to `@cogeotiff` 5.0.0

blacha:


packages/lambda-tiler/src/tile.set.vector.ts
```diff
@@ -19,11 +19,10 @@ class CotarCache {
     }
 
     async loadCotar(uri: string): Promise<Cotar | null> {
+        if (uri.endsWith('.tar.co')) throw new Error(`URI is not a cloud optimized tar ${uri}`);
```
Wentao-Kuang, line 22: unless instead of if?

blacha: Good catch ill update this

blacha:
No longer needed as we upgraded to a newer version of `@cogeotiff`

PR #1790, opened 2021-07-28
fix(config): do not cache tile sets forever as they can be updated

blacha:
This fixes a issue where a database update is applied but some lambdas will hold onto the old config until they get restarted

### DIFF
--- a/packages/lambda-tiler/src/cli/serve.ts
+++ b/packages/lambda-tiler/src/cli/serve.ts
@@ -110,7 +110,7 @@ async function useLocal(): Promise<void> {
         const requestId = ulid.ulid();
         const logger = LogConfig.get().child({ id: requestId });
 
-        const tileSets = await Promise.all([...TileSets.cache.values()]);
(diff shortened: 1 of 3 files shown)


packages/lambda-tiler/src/cli/serve.ts
```diff
@@ -110,7 +110,7 @@ async function useLocal(): Promise<void> {
         const requestId = ulid.ulid();
         const logger = LogConfig.get().child({ id: requestId });
 
-        const tileSets = await Promise.all([...TileSets.cache.values()]);
+        const tileSets = await Promise.all([...TileSets.cache.values()].map((c) => c.value));
```
Wentao-Kuang, line 113: Is this change for Cloning the TileSets values?

blacha: this change is because we now store the time of when we fetched the object last.

so the `[....cache.values()]`  has gone form `TileSet[]` to
```typescript
{ value: TileSet, time: number}[]
```

Wentao-Kuang (approved):

PR #1811, opened 2021-08-12
fix(lambda-tiler) Skip all the url settings for aerial sources in style json.

Wentao-Kuang:


packages/lambda-tiler/src/routes/tile.ts
```diff
@@ -122,12 +122,14 @@ export async function styleJson(req: LambdaContext, fileName: string): Promise<L
...
     const tileJsonUrl = `${host}/${version}/${name}/${rest[0]}/${rest[1]}/tile.json?api=${req.apiKey}`;
+    const rasterUrl = `${host}/${version}/${name}/aerial/${rest[1]}/{z}/{x}/{y}.webp?api=${req.apiKey}`;
     for (const [key, value] of Object.entries(style.sources)) {
-        if (value.url === '' || value.url.startsWith(host)) {
-            sources[key] = { type: 'vector', url: tileJsonUrl };
-        } else {
-            sources[key] = value;
+        if (value.type === 'vector' && value.url === '') {
+            value.url = tileJsonUrl;
+        } else if (value.type === 'raster' && value.tiles.length === 0) {
```
blacha, line 129: could we also check `Array.isArray(value.tiles)`  then set `value.tiles = [rasterUrl]`

might be safer if people forget the tiles.

blacha (approved):

PR #1818, opened 2021-08-18
feat(config): enable swapping of configuration providers dynamically

blacha:


packages/config/src/dynamo/__tests__/config.imagery.test.ts
```diff
@@ -1,8 +1,9 @@
 import { Epsg, EpsgCode, NamedBounds, QuadKey, TileMatrixSet, TileMatrixSets } from '@basemaps/geo';
 import o from 'ospec';
 import { createSandbox } from 'sinon';
+import { Config } from '../..';
```
Wentao-Kuang, line 4: My first time see this import? `'../..'`

blacha: `..` resolves to `../index.ts` so `../..` is `../../index.ts`

Wentao-Kuang (approved):

PR #1819, opened 2021-08-19
feat(server): create a standalone express server

blacha:
this adds a `basemaps-server` cli that can be used with the linz/basemaps-config repository to create a self hosted cli for the imagery

### DIFF
--- a/packages/config/src/base.config.ts
+++ b/packages/config/src/base.config.ts
@@ -58,6 +58,23 @@ export class ConfigInstance {
         return output;
     }
 
+    /**
+     * Prefix a dynamoDb id with the provided prefix if it doesnt already start with it.
+     */
+    prefix(prefix: ConfigPrefix, id: string): string {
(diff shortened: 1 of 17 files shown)


packages/config/src/base.config.ts
```diff
@@ -58,6 +58,23 @@ export class ConfigInstance {
         return output;
     }
 
+    /**
+     * Prefix a dynamoDb id with the provided prefix if it doesnt already start with it.
+     */
+    prefix(prefix: ConfigPrefix, id: string): string {
+        if (id === '') return id;
```
Wentao-Kuang, line 65: Should we return id here? I think the empty id is invalid.

blacha: This was just copy and pasted from the other spot, I haven't changed the logic

Wentao-Kuang (approved):

PR #1821, opened 2021-08-19
feat(lambda-tiler): remove `@basemaps/lambda` and replace with `@linzjs/lambda`

blacha:


packages/geo/src/stac/stac.attribution.ts
```diff
@@ -21,7 +21,7 @@ export type AttributionItem = StacItem<{
 /**
  * A Single File STAC compliant attribution json structure.
  */
-export interface AttributionStac extends StacCatalog, GeoJSON.GeoJsonObject {
+export interface AttributionStac extends StacCatalog, GeoJSON.GeoJsonObject, Record<string, unknown> {
```
blacha, line 24: This was complaining about being cast to a `Record<string, unknown>` not sure if there is a better way around this

packages/lambda-tiler/src/router.ts
```diff
@@ -0,0 +1,56 @@
+import { Const } from '@basemaps/shared';
```
blacha, line 1: Imported directly from `@basemaps/lambda` as it was not pushed across to `@linzjs/lambda`

packages/lambda-tiler/src/routes/tile.ts
```diff
@@ -55,7 +59,8 @@ export async function wmts(req: LambdaContext): Promise<LambdaHttpResponse> {
     const provider = await Config.Provider.get(providerId);
     if (provider == null) return NotFound;
 
-    const xml = WmtsCapabilities.toXml(host, provider, tileSets, req.apiKey);
+    const apiKey = Router.apiKey(req);
+    const xml = WmtsCapabilities.toXml(host, provider, tileSets, apiKey!);
```
blacha, line 63: I do not like this `!` but the apiKey is checked before this function is called. Im not sure it needs to exist. I might try and refactor this out

Wentao-Kuang (approved):

PR #1861, opened 2021-09-16
fix: bundle esm into commonjs for serving

blacha:


scripts/bundle.js
```diff
@@ -65,14 +65,15 @@ function bundleJs(basePath, cfg, outfile) {
...
-        `--target=${cfg.target || 'es2018'}`,
+        `--target=${cfg.target || 'es2020'}`,
         `--format=${cfg.format || 'cjs'}`,
         ...Object.entries(cfg.env || {}).map(defineEnv),
         ...Object.entries(DefaultEnvVars).map(defineEnv),
         ...(cfg.external || []).map((c) => `--external:${c}`),
         `--outfile=${outfile}`,
         joinPath(basePath, cfg.entry),
     ];
+    console.log(buildCmd);
```
Wentao-Kuang, line 76: `console.log` forgot to remove

packages/lambda-tiler/src/routes/health.ts
```diff
@@ -87,6 +86,7 @@ export async function Health(req: LambdaHttpRequest): Promise<LambdaHttpResponse
 
         // Get test tile to compare
         const testBuffer = await getTestBuffer(test);
+        test.buf = testBuffer;
```
Wentao-Kuang, line 89: Is this `test.buf` been used by anything?

scripts/bundle.js
blacha: I kinda like the build command being logged :D

packages/lambda-tiler/src/routes/health.ts
blacha: I believe it is somewhere, I didnt want to break backwards compatibility

Wentao-Kuang (approved):

PR #1870, opened 2021-09-21
feat(server): use the lambda handler directly

blacha:
This gives a much better logging output by default

This also enforces pretty printing of logs when run from the console

### DIFF
--- a/packages/server/src/server.ts
+++ b/packages/server/src/server.ts
@@ -1,50 +1,40 @@
-import { HttpHeader, HttpHeaderRequestId, LambdaAlbRequest } from '@linzjs/lambda';
-import { handleRequest } from '@basemaps/lambda-tiler';
-import * as ulid from 'ulid';
+import { handler } from '@basemaps/lambda-tiler';
+import { ALBEvent, ALBResult, APIGatewayProxyResultV2, CloudFrontRequestResult, Context } from 'aws-lambda';
+import ulid from 'ulid';
 import express from 'express';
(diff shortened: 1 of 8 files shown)


packages/server/src/server.ts
```diff
@@ -1,50 +1,40 @@
...
-    ctx.headers.set(HttpHeaderRequestId.RequestId, 'c' + requestId);
-  }
-  const ifNoneMatch = req.header(HttpHeader.IfNoneMatch);
-  if (ifNoneMatch != null && !Array.isArray(ifNoneMatch)) {
-    ctx.headers.set(HttpHeader.IfNoneMatch.toLowerCase(), ifNoneMatch);
-  }
+BasemapsServer.get('/v1/*', async (req: express.Request, res: express.Response) => {
+  const event: ALBEvent = {
+    httpMethod: 'GET',
+    requestContext: { elb: { targetGroupArn: 'arn:fake' } },
```
Wentao-Kuang, line 20: fake Arn?

blacha: it needed a 'string' for type completeness, doesn't matter what the string is as we dont use it anywhere

Wentao-Kuang: Weird, this cannot be optional?

blacha: Nope, this is a AWS ALB event type, so all ALB events should have a arn value there.

Wentao-Kuang (approved):

PR #1882, opened 2021-09-23
fix(lambda-tiler): limit the tiff memory cache to 256MB of imagery

blacha:


packages/lambda-tiler/src/tiff.cache.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.cacheB.clear();
+  }
+
+  static set(id: string, tiff: CogTiff): void {
+    TiffCache.cacheA.set(id, tiff);
+  }
+
+  /** Validate the size of the cache has not exploded */
+  static check(): void {
+    if (TiffCache.MaxCacheSizeBytes <= 0) return;
```
Wentao-Kuang, line 36: Why do we need to check `MaxCacheSizeBytes <= 0` since it is a static value?

blacha: Paranoia, there is nothing stopping a consumer of this package from just changing the value to -1

Wentao-Kuang (approved):

PR #1992, opened 2021-11-28
fix(landing): Fix the loading map by url coordinates for nztm

Wentao-Kuang:
`Config.map.transformedLocation` is used by update url location which transform current map location to google. But the loadiing map by url need to transform location from google to current map projection.

### DIFF
--- a/packages/landing/src/components/map.tsx
+++ b/packages/landing/src/components/map.tsx
@@ -52,7 +52,7 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
     const cfg = Config.map;
     const tileGrid = getTileGrid(cfg.tileMatrix.identifier);
     const style = tileGrid.getStyle(cfg.layerId, cfg.style);
-    const location = locationTransform(cfg.location, Config.map.tileMatrix, GoogleTms);
+    const location = locationTransform(cfg.location, cfg.tileMatrix, GoogleTms);
 
     this.map = new maplibre.Map({
       container: this.el,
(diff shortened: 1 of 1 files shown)


packages/landing/src/components/map.tsx
```diff
@@ -112,7 +112,9 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
     const center = this.map.getCenter();
     if (center == null) throw new Error('Invalid Map location');
     const zoom = Math.floor((this.map.getZoom() ?? 0) * 10e3) / 10e3;
-    return Config.map.transformLocation(center.lat, center.lng, zoom);
+    const location = { lat: center.lat, lon: center.lng, zoom };
+    Config.map.setLocation(location);
```
blacha, line 116: I dont think a `getter` should also `set`, I think we only use this function in one place did you want to just inline it into the `setLocationUrl` function

packages/landing/src/components/map.tsx
```diff
@@ -52,7 +52,7 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
     const cfg = Config.map;
     const tileGrid = getTileGrid(cfg.tileMatrix.identifier);
     const style = tileGrid.getStyle(cfg.layerId, cfg.style);
-    const location = cfg.transformedLocation;
+    const location = locationTransform(cfg.location, Config.map.tileMatrix, GoogleTms);
```
blacha, line 55: everything else is using `cfg` here so `cfg.tileMatrix`

Wentao-Kuang: Good pick.

blacha (approved):

PR #2026, opened 2021-12-15
fix(landing): enable debug page

Wentao-Kuang:


packages/landing/src/components/debug.tsx
```diff
@@ -0,0 +1,79 @@
...
+export class Debug extends Component<{ map: maplibre.Map }> {
+  componentDidMount(): void {
+    setTimeout(() => this.props.map.resize(), 100);
+  }
+
+  render(): ComponentChild {
+    return (
+      <div className="debug">
+        <div className="debug__info">
+          <label className="debug__label">
```
blacha, line 14: the label should be the title not the title and value

`<label>ImageId</label><div>{Config.map.layerId}</div>`

Wentao-Kuang: Thanks fixed.

blacha (approved):

PR #2027, opened 2021-12-15
fix(landing): Trigger change event when debug value changes and disable switcher for debug mode.

Wentao-Kuang:


packages/landing/src/config.map.ts
```diff
@@ -85,7 +85,7 @@ export class MapConfig extends Emitter<MapConfigEvents> {
     this.style = style ?? null;
     this.layerId = layerId.startsWith('im_') ? layerId.slice(3) : layerId;
     this.tileMatrix = tileMatrix;
-    this.debug = debug;
```
blacha, line 88: can you also rename it to `isDebug` to show its a boolean

Wentao-Kuang: Thanks fixed.

blacha: was meaning `this.isDebug = debug` not the setter function name.

Wentao-Kuang: Sorry, updated.

blacha (approved):

PR #2030, opened 2021-12-16
feat(landing): allow using the topographic vector map as a debug layer

blacha:


packages/landing/src/components/debug.tsx
```diff
@@ -34,7 +43,75 @@ export class Debug extends Component<{ map: maplibre.Map }> {
     );
   }
 
-  adjustOSM = (e: Event): void => {
+  _styleJson: Promise<Style> | null = null;
+  get styleJson(): Promise<Style> {
+    if (this._styleJson == null) {
+      this._styleJson = fetch(
+        WindowUrl.toTileUrl(MapOptionType.TileVectorStyle, Config.map.tileMatrix, 'topographic', 'topographic'),
```
Wentao-Kuang, line 50: `Config.map.tileMatrix` probably not working for NZTM tileMatrix, do we want to disable the slider for NZTM?

blacha: We don't disable OSM, we could make another PR to fix that.

Wentao-Kuang: Sorry, I mean we can just disable the Topographic one for NZTM.
` <div className="debug__info">
          <label className="debug__label">Topographic</label>
          {debugSlider(this.adjustTopographic)}
        </div>`

Wentao-Kuang (approved):

PR #2032, opened 2021-12-16
fix(lambda-tiler): remove the host check to add api keys for all stylejson sources.

Wentao-Kuang:


packages/lambda-tiler/src/routes/tile.ts
```diff
@@ -132,16 +132,16 @@ export const TileRoute = {
     const sources: Sources = {};
     for (const [key, value] of Object.entries(style.sources)) {
       if (value.type === 'vector') {
-        if (value.url.includes(host)) {
-          const url = new URL(value.url);
```
blacha, line 136: what happens with `https://basemaps.linz.govt.nz/` does this duplicate `/`? I normally use `fsa.join` to join urls

`path.join` with urls work very well either

blacha (approved):

PR #2035, opened 2021-12-19
feat(landing): rendering source imagery bounds

blacha:


packages/landing/src/components/debug.tsx
```diff
@@ -33,10 +33,42 @@ export class Debug extends Component<{ map: maplibre.Map }> {
...
           <input type="checkbox" onClick={this.togglePurple} />
         </div>
+        {this.renderSourceToggle()}
       </div>
     );
   }
 
+  renderSourceToggle(): ComponentChild {
+    // TODO this is a nasty hack to detect if a direct imageryId is being viewed
+    if (!Config.map.layerId.startsWith('01')) return null;
```
Wentao-Kuang, line 43: Might be good to have boolean property in `Config.map` to catch the `im_` as imagery?

Wentao-Kuang, line 43: Might be good to validate this as ulid?

blacha: I think this is fine, as if its not a valid id the service will  `404` saying it cannot find the imagery.

Wentao-Kuang (approved):

PR #2107, opened 2022-03-16
Revert "release: v6.21.0 (#2104)"

blacha:
This reverts commit d57532186c445aa3b03efe3d91645c4efb2d714d.

Since v6.21.0 was never published to NPM the deployment process will fail until it is, this reverts all the versions back to pre v6.21.0 numbers a new v6.22.0 will be published later.

### DIFF
--- a/packages/_infra/package.json
+++ b/packages/_infra/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@basemaps/infra",
-  "version": "6.21.0",
+  "version": "6.20.0",
   "private": true,
   "repository": {
     "type": "git",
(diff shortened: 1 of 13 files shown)


Wentao-Kuang (approved):

Wentao-Kuang (changes requested):
Do we want to revert change logs as well?

blacha:
> Do we want to revert change logs as well?

The changes are still valid for v6.21.0 its just it was never published. so I think we should keep the changelogs

Wentao-Kuang (approved):

PR #2120, opened 2022-03-18
feat(cli): support multiple invalidations for cloudfront

blacha:


packages/cli/src/cli/util.ts
```diff
@@ -38,12 +38,13 @@ export async function invalidateCache(path: string, commit = false): Promise<voi
...
 
   LogConfig.get().info({ path, cfId: cf.Id }, 'Invalidating');
+  const Items = Array.isArray(path) ? path : [path];
   if (commit) {
     await cloudFront
       .createInvalidation({
         DistributionId: cf.Id,
         InvalidationBatch: {
-          Paths: { Quantity: 1, Items: [path] },
+          Paths: { Quantity: 1, Items },
```
Wentao-Kuang, line 47: Do we also need to update the `Quantity: Items.length`?

blacha: I was wondering that, I'm not sure, let me run some tests

blacha: You are 100% correct, the Quantity needs to match the `Items.length`

> Error InconsistentQuantities: The specified quantity of Paths (1) does not match the actual quantity supplied (2).

Wentao-Kuang: That's logically make sense.

Wentao-Kuang (approved):

PR #2158, opened 2022-04-19
build: upgrade to aws-cdk v2

blacha:


Wentao-Kuang (approved):
Nice, removed 1500+ lines from yarn.lock.

PR #2162, opened 2022-04-20
feat(config): Add config job for imagery processing api

Wentao-Kuang:


packages/config/src/config/job.ts
```diff
@@ -0,0 +1,15 @@
...
+  Complete = 'complete',
+  Fail = 'fail',
+}
+
+export interface ConfigJob extends BaseConfig {
+  /** Job Status for the imagery importing batch jobs */
+  status: JobStatus;
+
+  /** Job Batch processing error messages */
+  error?: string;
```
blacha, line 14: error should be on a seperate type, we should have something like

```
export interface ConfigJobFialed  {
 status: 'failed'
 error: ...
}

export interface ConfigJob = ConfigJobFailed | ConfigJobSuccess
```
 

then the success option could have extra keys

packages/config/src/config/prefix.ts
```diff
@@ -4,4 +4,5 @@ export enum ConfigPrefix {
   Provider = 'pv',
   ImageryRule = 'ir',
   Style = 'st',
+  Job = 'job',
```
blacha, line 7: have three letters here will break other logic, there are quite a few places where it expects the prefix to be three lettters total `pv_`

packages/config/src/dynamo/dynamo.config.ts
```diff
@@ -19,6 +20,7 @@ export class ConfigProviderDynamo extends BasemapsConfigProvider {
   Style = new ConfigDynamoCached<ConfigVectorStyle>(this, ConfigPrefix.Style);
   TileSet = new ConfigDynamoBase<ConfigTileSet>(this, ConfigPrefix.TileSet);
   Provider = new ConfigDynamoCached<ConfigProvider>(this, ConfigPrefix.Provider);
+  Job = new ConfigDynamoCached<ConfigJob>(this, ConfigPrefix.Job);
```
blacha, line 23: We use job as a processing job, this could get confusing as to what "job" are we talking about?

Any thoughts on other names?

Wentao-Kuang: How about BatchJob? with prefix with `bt_`

blacha: Some random thoughts
- ImageryImport `ii_`
- CogCreation `cc_`
- ImportJob `ij_`

Wentao-Kuang: My preference is CogCreation, what do you think?

blacha (approved):

PR #2169, opened 2022-04-29
feat(cli): allow overriding imagery names

blacha:


Wentao-Kuang:
Build failure.

packages/lambda-tiler/src/routes/import.ts
```diff
@@ -26,15 +28,24 @@ export async function Import(req: LambdaHttpRequest): Promise<LambdaHttpResponse
     targetTms = tileMatrix;
   }
 
-  // Find the imagery from s3
+  // Validate the s3 path
```
Wentao-Kuang, line 31: @blacha I added some more here, to validate the imagery name before pass it to the CogJobContext.

blacha: this is going to give very weird imagery names, maybe we should just use `:ulid_:year`  to show its a completly made up name?

Some of these import paths will be completely meaningless, imagine a folder  `s3://a/2022/04/eca/RGB/` what sort of "name" would we get from this?

Wentao-Kuang: Updated.

Wentao-Kuang (dismissed):

packages/lambda-tiler/src/routes/import.ts
Wentao-Kuang: Wait, we cannot just add uild_year into `CogJobContext`. This will make a different hash each time from `CogJobContext`.

Wentao-Kuang: I am going to make this imageryName as optional attributions for CogJobContext. And add this after creating hash.

packages/cli/src/cog/job.factory.ts
```diff
@@ -22,7 +22,9 @@ export const CogJobFactory = {
    */
   async create(ctx: JobCreationContext): Promise<CogJob> {
     const id = ctx.override?.id ?? ulid.ulid();
-    const imageryName = basename(ctx.sourceLocation.path).replace(/\./g, '-'); // batch does not allow '.' in names
+    let imageryName = ctx.imageryName;
+    if (imageryName == null) imageryName = basename(ctx.sourceLocation.path).replace(/\./g, '-'); // batch does not allow '.' in names
```
blacha, line 26: this was removed, we dont use the imagery name inside of batch so we can remove the .replace

Wentao-Kuang: Fixed. Thanks.

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -96,6 +96,7 @@ export class ActionCogCreate extends CommandLineAction {
     // Output file exists don't try and overwrite it
     if (outputExists) {
       logger.warn({ targetPath }, 'CogCreate:OutputExists');
+      await this.checkJobStatus(job, logger);
```
Wentao-Kuang, line 99: @blacha  I have added this check in here. In cases, partially processed file list never get to `CogCreate:JobComplete` to create 'ts_' and 'im_' records.

Wentao-Kuang (approved):

PR #2170, opened 2022-04-29
feat(lambda-tiler): Import api for import imagery jobs.

Wentao-Kuang:


packages/shared/src/const.ts
```diff
@@ -30,6 +30,12 @@ export const Env = {
   /** Number of hours to assume a role for, @default 8 */
   AwsRoleDurationHours: 'AWS_ROLE_DURATION_HOURS',
 
+  /** AWS role config bucket */
+  AwsRoleConfigBucket: 'ROLE_CONFIG_BUCKET',
```
Wentao-Kuang, line 34: @blacha  Do you know where to set the values for these environments variables?

blacha: It should be inside of `_infra` where the lambda is deployed

packages/lambda-tiler/src/__test__/tile.import.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    // When ...Then ...
+    const res = await Import(req);
+    o(res.body).equals('{"status":404,"message":"Imagery Not Found"}');
+  });
+
+  o('should return 200 with new import', async () => {
+    // Given... A new import with no existing config.
+    sandbox.stub(fsa, 'readJson').resolves({ buckets: [role] });
+    sandbox.stub(fsa, 'list').callsFake(async function* () {
+      yield files[0];
```
blacha, line 104:
```
for (file in files) yield file
```

blacha:
```
async * function listFiles() {
}

o.spec(() => {sandbox.stub(fsa, 'list').callsFake(listFiles) }
```

packages/lambda-tiler/src/import/imagery.find.ts
```diff
@@ -0,0 +1,54 @@
+import { AwsCredentials } from '@chunkd/source-aws-v2';
+import { fsa } from '@chunkd/fs';
+import { Env } from '@basemaps/shared';
+
+export type RoleConfig = {
+  bucket: string;
+  accountId: string;
+  roleArn: string;
+  externalId?: string;
```
blacha, line 9:
```typescript
/** Duration in seconds for how long the role can be assumed */
roleSessionDuration?: number
```

packages/lambda-tiler/src/import/imagery.find.ts
```diff
@@ -0,0 +1,54 @@
+import { AwsCredentials } from '@chunkd/source-aws-v2';
+import { fsa } from '@chunkd/fs';
+import { Env } from '@basemaps/shared';
+
+export type RoleConfig = {
```
blacha, line 5:
```
export interface 
```

packages/lambda-tiler/src/import/imagery.find.ts
```diff
@@ -0,0 +1,54 @@
...
+  version: string;
+  package: string;
+  hash: string;
+  updatedAt: string;
+};
+
+export class RoleRegister {
+  /** Get all imagery source aws roles */
+  static async _loadRoles(): Promise<RoleConfig[]> {
+    const configPath = `s3://${Env.AwsRoleConfigBucket}/config.json`;
```
blacha, line 24:
```
s3://ROLE_CONFIG_BUCKET/config.json
```

I think its 

```
Env.get(Env.AwsRoleConfigBucket)
```

blacha: Catch errors return []
if bucket == null return []

packages/lambda-tiler/src/import/imagery.find.ts
blacha: default 3600?

packages/lambda-tiler/src/import/make.cog.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (path.includes('geographx')) {
+    resampling = {
+      warp: 'bilinear',
+      overview: 'bilinear',
+    } as const;
+  }
+
+  const ctx: JobCreationContext = {
+    override: { projection: tileMatrix.projection, resampling },
+    outputLocation: { type: 's3' as const, path: `s3://${Env.ImportImageryBucket}` },
```
blacha, line 31: `Env.get`

packages/lambda-tiler/src/import/make.cog.ts
```diff
@@ -0,0 +1,38 @@
...
+export async function makeCog(
+  path: string,
+  tileMatrix: TileMatrixSet,
+  role: RoleConfig,
+  files: string[],
+): Promise<JobCreationContext> {
+  let resampling;
+
+  /** Process Gebco 2193 as one cog of full extent to avoid antimeridian problems */
+  if (tileMatrix.projection === Epsg.Nztm2000 && path.includes('gebco')) {
```
blacha, line 15: drop these.

packages/lambda-tiler/src/import/make.cog.ts
blacha: I think this is the full path?

```
s3://linz-basemaps-cache/:processingId/
```

packages/lambda-tiler/src/routes/import.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @example
+ * - /v1/import?path=s3://linz-imagery-staging/2022-03/wellington_rural_2022_delivery_1
+ */
+export async function Import(req: LambdaHttpRequest): Promise<LambdaHttpResponse> {
+  const path = req.query.get('path');
+  const projection = req.query.get('p');
+
+  // Parse projection as target, default to process both NZTM2000Quad
+  let target = Nztm2000Tms;
```
blacha, line 21: I think there is a function that does this for us?
```
getTileMatrix(randomSTring)
```

possibly in @basemaps/shared

blacha: const `targetTileMatrix = getTIleMatrix(projection) ?? Nztm2000Tms`

packages/lambda-tiler/src/import/imagery.find.ts
Wentao-Kuang: Doesn't need to handle default value here as `AwsCredentials.fsFromRole` already handle it to 3600 default for us.

packages/lambda-tiler/src/routes/import.ts
Wentao-Kuang: I think we need to keep the logic here to return 404 if projection url parameter is incorrect. In case someone put a wrong projection code but lead to default 2193.

packages/lambda-tiler/src/import/make.cog.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (bucket == null) throw new Error('Output AWS s3 bucket Not Found.');
+  const ctx: JobCreationContext = {
+    override: {
+      projection: tileMatrix.projection,
+      resampling: {
+        warp: 'bilinear',
+        overview: 'lanczos',
+      },
+    },
+    outputLocation: { type: 's3' as const, path: `s3://${bucket}` },
```
blacha, line 22: does this path need to be more specific? `s3://${bucket}/${ulid}` ?

Wentao-Kuang: Yes, we can't add id here until we got the context first. So I add it in the `import()` function.
`// Add id back to JobCreationContext
    ctx.override!.id = id;
    ctx.outputLocation.path = fsa.join(ctx.outputLocation.path, id);`

blacha (approved):

PR #2180, opened 2022-05-09
feat(cli): make cogs will update the process job status if exists.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -126,6 +128,20 @@ export class ActionCogCreate extends CommandLineAction {
       } else {
         logger.warn('DryRun:Done');
       }
+    } catch (e) {
+      const jobId = Config.ProcessingJob.id(job.id);
+      const jobConfig = await Config.ProcessingJob.get(jobId);
+      if (jobConfig != null) {
+        const jobFailed = jobConfig as ProcessingJobFailed;
+        jobFailed.status = 'failed';
+        if (typeof e === 'string') {
```
blacha, line 137: `jobFailed.error = String(e)`

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -126,6 +128,20 @@ export class ActionCogCreate extends CommandLineAction {
...
+          jobFailed.error = e;
+        } else if (e instanceof Error) {
+          jobFailed.error = e.message; // works, `e` narrowed to Error
+        }
+        const config = new ConfigProviderDynamo(Const.TileMetadata.TableName);
```
blacha, line 142: Why are you making a new config here? you already have `await Config.ProcessingJoob.get(...)`

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -148,6 +164,13 @@ export class ActionCogCreate extends CommandLineAction {
 
     if (expectedTiffs.size === 0) {
       logger.info({ tiffCount: jobSize, tiffTotal: jobSize }, 'CogCreate:JobComplete');
+      const jobId = Config.ProcessingJob.id(job.id);
+      const jobConfig = await Config.ProcessingJob.get(jobId);
+      if (jobConfig != null) {
```
blacha, line 169: This should happen before the "JobComplete" message.

packages/cli/src/cli/cogify/action.cog.ts
Wentao-Kuang: I need to use `config.ProcessingJob.put()`, we only have get function in `Config.ProcessingJob`

blacha: I'm not sure I follow?

`config.ProcessingJob.put` is a function and it exists?

Wentao-Kuang: Sorry, I need a put function but it is not exists in the `Config.ProcessingJob`, this `Config` is import from `@basemaps/shared` as the following code `export const Config = new ConfigInstance();`. 

To be able to put item into dynamodb, I need a `const config = new ConfigProviderDynamo(Const.TileMetadata.TableName);`

blacha: You could use a `instanceof` check to see if it has a `.put()` like we do in basemaps-config?

https://github.com/linz/basemaps-config/blob/master/src/config/tileset.updater.ts#L122

Wentao-Kuang: Let me try. Thanks.

Wentao-Kuang: Works for me. Just updated.

packages/cli/src/cli/cogify/action.cog.ts
blacha: You don't need `instanceof Error` this can be simplified to remove the if statement.

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -126,6 +128,22 @@ export class ActionCogCreate extends CommandLineAction {
...
         logger.warn('DryRun:Done');
       }
+    } catch (e) {
+      if (job.processingId != null) {
+        // Update job status if this is the processing job.
+        const jobConfig = await Config.ProcessingJob.get(job.processingId);
+        if (jobConfig != null) {
+          const jobFailed = jobConfig as ProcessingJobFailed;
+          jobFailed.status = 'failed';
+          if (typeof e === 'string') {
```
blacha, line 138: this if statement can be replaced with `jobFailed.error = String(e)`

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -147,6 +165,18 @@ export class ActionCogCreate extends CommandLineAction {
     }
 
     if (expectedTiffs.size === 0) {
+      if (job.processingId != null) {
+        // Update job status if this is the processing job.
+        const jobConfig = await Config.ProcessingJob.get(job.processingId);
+        if (jobConfig != null) {
+          const jobComplete = jobConfig as ProcessingJobComplete;
+          jobComplete.status = 'complete';
+          jobComplete.projection = job.tileMatrix.projection.code;
```
blacha, line 174: this should be set well before the end?

Also we should use `tileMatrix` over `projection`

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -147,6 +165,18 @@ export class ActionCogCreate extends CommandLineAction {
...
+          jobComplete.url = await prepareUrl(job);
```
blacha, line 175: this url should be logged in the `JobComplete` log just below

packages/config/src/config/processing.job.ts
```diff
@@ -1,15 +1,20 @@
...
 export interface ProcessingJob extends BaseConfig {
-  /** Job Status for the imagery importing batch jobs */
-  status: JobStatus;
+  status: 'processing';
+}
+
+export interface ProcessingJobComplete extends BaseConfig {
+  status: 'complete';
+  /** Processed Imagery projection */
+  tileMatrix: TileMatrixSet;
```
blacha, line 13: you cannot store the full class inside the database? this should be a `string`

Wentao-Kuang: Do you mean the tileMatrix.identifier? or just String the whole object?

blacha: > Do you mean the tileMatrix.identifier? or just String the whole object?

Yup just the identifier, we could type this as `NZTM2000Quad | WebMercatorQuad` as those are the only two we should be processing!

blacha (approved):

PR #2182, opened 2022-05-10
fix(lambda-tiler): Fix the insert of processing job config and actually start job after inserting config.

Wentao-Kuang:


packages/lambda-tiler/src/routes/import.ts
```diff
@@ -43,16 +43,18 @@ export async function Import(req: LambdaHttpRequest): Promise<LambdaHttpResponse
...
+    // Insert Processing job config
     jobConfig = {
       id: jobId,
       name: path,
       status: 'processing',
     } as ConfigProcessingJob;
 
-    const config = new ConfigProviderDynamo(Const.TileMetadata.TableName);
-    await config.ProcessingJob.put(jobConfig);
+    if (Config.ProcessingJob instanceof ConfigDynamoBase) await Config.ProcessingJob.put(jobConfig);
```
blacha, line 53: I made https://github.com/linz/basemaps/pull/2183 to make this easier in the future.

Wentao-Kuang: Thanks. Just Updated. Can you plz give another go.

blacha (approved):

PR #2184, opened 2022-05-10
feat(config): Relocate imagery config from basemaps-config into config package.

Wentao-Kuang:


packages/config/src/imagery.config.ts
```diff
@@ -0,0 +1,99 @@
+import { Bounds, ImageFormat, TileMatrixSet } from '@basemaps/geo';
+import { fsa } from '@chunkd/fs';
+import { basename } from 'path';
+import ulid from 'ulid';
+import { ConfigImagery } from './config/imagery.js';
+import { ConfigTileSet, TileSetType } from './config/tile.set.js';
+
+function guessIdFromUri(uri: string): string {
```
Wentao-Kuang, line 8: Hi @blacha, 

I am relocating this into `basemaps/package/config` to make the use of for both `basemap/cli` and `basemaps-config`.

I found that this function might not working for hashed `CogJobContext` id. Do you have any ideal to make this better?

blacha: We have all the ids inside the `job.json` so I don't think we need to guess them.

blacha: the output path should still be a `ulid` so the `job.json` has all the information we need to make the configuration file.

I don't think we "need" this logic for guessing things.

Wentao-Kuang: image
I am a little confused, I thought we are using the processingId as tileset id in s3 path??

blacha: I thought we were storing them as something like

`s3://linz-basemaps-cache/:ulid`

so even though the processing id is a super long hash eg `b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c`  the stored files will have a date associated with them,

the `:ulid` should be stored somewhere so we dont have to guess it from a URI.

Wentao-Kuang: This should be ready if we using ulid.

Wentao-Kuang:
This change is not necessary. Closed  the PR.

PR #2191, opened 2022-05-12
feat(cli): Insert imagery and tileset config after cog creation complete

Wentao-Kuang:


packages/cli/src/cli/cogify/imagery.config.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export async function insertConfigImagery(job: CogStacJob, logger: LogType): Promise<void> {
+  const now = Date.now();
+  const imgId = Config.Imagery.id(job.id);
+  const configImagery: ConfigImagery = {
+    id: imgId,
+    name: job.name,
+    createdAt: now,
+    updatedAt: now,
+    projection: job.tileMatrix.projection.code,
```
blacha, line 18: we should add tileMatrix identifier here too.

Wentao-Kuang: Updated. We might need to update `basemap-config` repo, next time when we upgrade config package versions.

blacha: Build failure but otherwise :lgtm:

blacha (approved):

PR #2197, opened 2022-05-16
fix(infra): Fix the cog batch job to put record to dynamodb.

Wentao-Kuang:
This should fix the following problems, when insert and update records to dynamodb.

image
image

### DIFF
--- a/packages/_infra/src/serve/db.ts
+++ b/packages/_infra/src/serve/db.ts
@@ -29,7 +29,7 @@ export const TileMetadataTableArn = {
     return cdk.Arn.format(
       {
         service: 'dynamodb',
-        region: '*',
+        region: Const.Aws.Region,
         resource: 'table',
         resourceName: Const.TileMetadata.TableName,
       },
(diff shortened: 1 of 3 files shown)


packages/_infra/src/serve/db.ts
```diff
@@ -29,7 +29,7 @@ export const TileMetadataTableArn = {
     return cdk.Arn.format(
       {
         service: 'dynamodb',
-        region: '*',
```
blacha, line 32: I dont think this will fix the problem, this was to allow other regions to write to dynamodb

Wentao-Kuang: I see. I need to have a look what cause the problem. Just don't how to test this.

blacha: where is it coming from, do you have the bundled index.js to see the stack trace?

Wentao-Kuang: Yes, That's what I am thinking to get a bundle index.js first.

Wentao-Kuang: image
This the place of error, I can't find where is the place to using this. 
However, I am sure ti broke at the place for puting item to dynamodb.

blacha (approved):

PR #2200, opened 2022-05-17
feat(config): add configuration parser and bundler

blacha:
- feat(config-cli): import configuration cli from linz/basemaps-config
- fix: include bin/
- feat: load screenshot config from file
- feat: add name to log lines

### DIFF
--- a/packages/config/src/json/json.config.ts
+++ b/packages/config/src/json/json.config.ts
@@ -1,0 +1,221 @@
+import { Bounds, GoogleTms, ImageFormat, Nztm2000QuadTms, TileMatrixSet, VectorFormat } from '@basemaps/geo';
+import { fsa } from '@chunkd/fs';
+import { basename } from 'path';
+import ulid from 'ulid';
+import { Config } from '../base.config.js';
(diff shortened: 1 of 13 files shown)


packages/config/src/json/json.config.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  async _loadImagery(uri: string, tileMatrix: TileMatrixSet, name: string): Promise<ConfigImagery> {
+    this.logger.trace({ uri }, 'FetchImagery');
+
+    // TODO is there a better way of guessing the imagery id & tile matrix?
+    const id = Config.prefix(ConfigPrefix.Imagery, guessIdFromUri(uri));
+
+    const fileList = await fsa.toArray(fsa.list(uri));
+    const tiffFiles = fileList.filter((f) => f.endsWith('.tiff'));
```
Wentao-Kuang, line 168: Do we consider imagery end with `.tif` ?

blacha: these our COGs so we decide what they end with

Wentao-Kuang (approved):

PR #2203, opened 2022-05-17
feat(lambda-tiler): Add file number limitation to import api.

Wentao-Kuang:


packages/cli/src/cog/cog.stac.job.ts
```diff
@@ -217,11 +217,9 @@ export class CogStacJob implements CogJob {
         bounds: metadata.targetBounds,
         oneCogCovering: ctx.oneCogCovering,
       },
+      processingId: ctx.override?.processingId,
```
Wentao-Kuang, line 220: This fix a problem for passing `processingId` to `job.json`. Otherwise, I need to set it as `job.json.processingId = ctx.override?.processingId;`

packages/lambda-tiler/src/routes/import.ts
```diff
@@ -44,6 +44,7 @@ export async function Import(req: LambdaHttpRequest): Promise<LambdaHttpResponse
   if (role == null) return new LambdaHttpResponse(403, 'Unable to Access the s3 bucket');
   const files = await findImagery(path);
   if (files.length === 0) return new LambdaHttpResponse(404, 'Imagery Not Found');
+  if (files.length >= 500) return new LambdaHttpResponse(400, `Too many files to process. Files: ${files.length}`);
```
blacha, line 47: this limit would be way too low, maybe 5_000 ?

Wentao-Kuang: Sure. Updated.

blacha (approved):

PR #2206, opened 2022-05-19
feat(lambda-tiler): Load MaxImagePixelSize from import api.

Wentao-Kuang:


packages/cli/src/cog/cutline.ts
```diff
@@ -155,7 +155,7 @@ export class Cutline {
    * Generate an optimized WebMercator tile cover for the supplied source images
    * @param sourceMetadata contains images bounds and projection info
    */
-  optimizeCovering(sourceMetadata: SourceMetadata): NamedBounds[] {
+  optimizeCovering(sourceMetadata: SourceMetadata, maxImageSize?: number): NamedBounds[] {
```
blacha, line 158: this can be defaulted in the function declaration with `, maxImageSize = MaxImagePIxelWidth` and means you can remove l168

blacha (approved):

blacha (approved):

PR #2207, opened 2022-05-22
feat(lambda-cog): New lambda Cog for import api.

Wentao-Kuang:


packages/lambda-cog/src/api.key.ts
```diff
@@ -0,0 +1,23 @@
+import * as ulid from 'ulid';
```
Wentao-Kuang, line 1: We might need to do a different api key logic here?

packages/_infra/src/serve/lambda.cog.ts
```diff
@@ -0,0 +1,74 @@
...
+        statements: [stsPolicy],
+      }),
+    );
+
+    this.functionUrl = new lambda.FunctionUrl(this, 'LambdaCogUrl', {
+      function: this.lambda,
+      authType: lambda.FunctionUrlAuthType.NONE,
+      cors: {
+        allowedOrigins: ['*'],
+        allowedMethods: [lambda.HttpMethod.GET, lambda.HttpMethod.POST],
```
blacha, line 68: do we need POST?

packages/lambda-cog/bundle.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# Create a deployment bundle with sharp/libvips prebuilt included
+#
+../../scripts/bundle.js package.json
+cd dist
```
blacha, line 6: We do not include any native modules so we shouldn't need this.

packages/lambda-cog/scripts/create.deployment.package.mjs
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env node
```
blacha, line 1: we dont install sharp so we dont need this.

packages/lambda-cog/src/api.key.ts
blacha, line 1: this endpoint doesnt need a api key.

packages/lambda-cog/bundle.sh
Wentao-Kuang: If we remove the bundle, do we also need to update the code path to point to build instead of dist? `const CODE_PATH = '../lambda-cog/build;`

blacha: it should be bundled into `dist` for us by the bundler.

What happens when you run bundle in the root?
```
npx lerna run bundle
```

`lambda-analytics` would be a better example as it hast a lot less stuff in it.

Wentao-Kuang: I see, just fixed that, I point to the correct bundle script in the package.json.

packages/_infra/src/serve/lambda.cog.ts
```diff
@@ -0,0 +1,74 @@
...
+  public constructor(scope: cdk.Stack, id: string, props: LambdaCogProps) {
+    super(scope, id);
+
+    const config = getConfig();
+    /**
+     * WARNING: changing this lambda name while attached to a alb will cause cloudformation to die
+     * see: https://github.com/aws/aws-cdk/issues/8253
+     */
+    this.lambda = new lambda.Function(this, 'Cog', {
+      vpc: props.vpc,
```
blacha, line 34: we dont need a VPC

packages/_infra/src/serve/lambda.cog.ts
```diff
@@ -0,0 +1,74 @@
...
+  public lambda: lambda.Function;
+  public functionUrl: lambda.FunctionUrl;
+  public version: lambda.Version;
+
+  public constructor(scope: cdk.Stack, id: string, props: LambdaCogProps) {
+    super(scope, id);
+
+    const config = getConfig();
+    /**
+     * WARNING: changing this lambda name while attached to a alb will cause cloudformation to die
```
blacha, line 30: we are not using ALBs so this warning doesnt matter.

packages/_infra/src/serve/lambda.cog.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+    const configBucket = s3.Bucket.fromBucketName(this, 'ConfigBucket', config.AwsRoleConfigBucket);
+    configBucket.grantRead(this.lambda);
+
+    // We need to assume role to find imagery from data lake
+    const stsPolicy = new iam.PolicyStatement();
+    stsPolicy.addActions('sts:AssumeRole');
+    stsPolicy.addAllResources();
+    this.lambda.role?.attachInlinePolicy(
+      new iam.Policy(this, 'assume-role-policy', {
```
blacha, line 58: we generally use TitleCase for resouceids `AssumeRolePolicy`

Wentao-Kuang: All fixed. Thank you.

packages/lambda-cog/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "name": "Land Information New Zealand",
+    "url": "https://linz.govt.nz",
+    "organization": true
+  },
+  "type": "module",
+  "engines": {
+    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
```
blacha, line 19: I dont think we need to publish this to NPM, can you set this package to "private: true" and remove the publish config

yarn.lock
```diff
@@ -6906,6 +6906,13 @@ semver@^7.1.1, semver@^7.1.3, semver@^7.3.2, semver@^7.3.4, semver@^7.3.5:
   dependencies:
     lru-cache "^6.0.0"
 
+semver@^7.3.7:
```
blacha, line 6909: this is a little odd, all of these should have had the same version here.

have you run a `yarn` in the root directory.

packages/lambda-cog/package.json
Wentao-Kuang: Can we add this into package/server later?

yarn.lock
Wentao-Kuang: Yes, this is the yarn.lock after running in root. The only changes I have made is the cdk version to be able to use FuntionUrl.

packages/lambda-cog/package.json
blacha: I dont think we need this to be published ever? well not as a lambda function anyway? this is all very specific to LINZ's internal storage structure.

Wentao-Kuang: Done, thanks.

packages/_infra/src/index.ts
```diff
@@ -44,8 +45,10 @@ async function main(): Promise<void> {
    */
   const edge = new EdgeStack(basemaps, 'Edge', { env: { region: 'us-east-1', account }, cloudfrontCertificateArn });
   const serve = new ServeStack(basemaps, 'Serve', { env: { region: BaseMapsRegion, account }, albCertificateArn });
+  const cog = new CogStack(basemaps, 'LambdaCog', { env: { region: BaseMapsRegion, account } });
 
   edge.addDependency(serve);
+  edge.addDependency(cog);
```
blacha, line 51: edge doesnt need a dependency on this. this is a seperate url?

packages/cli/src/cog/builder.ts
```diff
@@ -226,6 +226,7 @@ export class CogBuilder {
       else union = Bounds.fromJson(bounds).union(union);
     }
     if (union == null) throw new Error('Bug! union can not be null');
+    if (existsSync(CacheFolder)) rmdirSync(CacheFolder, { recursive: true });
```
blacha, line 229: This defeats the purpose of having a cache.

the goal was to cache things between runs. can we disable the cache instead?

Wentao-Kuang: What do you mean disable cache? Just create meta data on fly?

blacha: this cache folder was so that if you run the command twice you can use the cached data

eg
```
./import s3://a
./import s3://a --commit
```

So it doesnt have to recompute everything

If we delete the cache after the first import invocation there is no point to having a cache.

so rather than just deleting the cache folder it would be better just to stop using the cache.

Wentao-Kuang: I see what you mean. Do we want to make this as a option parameter? For the import api, it seems we always creating new job if there is any changes. But we can still using this for import cli?

blacha: If we are moving this all into the cloud then lets just delete it, not point storing the cache here.

Wentao-Kuang: All fixed. Thank you.

packages/_infra/src/index.ts
Wentao-Kuang: Fixed.

blacha (approved):

PR #2210, opened 2022-05-25
feat(cli): Upgrade action cog to be able to group small jobs into a single batch job.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -45,25 +46,34 @@ export class ActionCogCreate extends CommandLineAction {
...
-    const cogIndex = this.cogIndex?.value;
-    if (cogIndex != null) {
-      const { name } = files[cogIndex];
-      if (name == null) {
-        throw new LoggerFatalError({ cogIndex, tileMax: files.length - 1 }, 'Failed to find cog name from index');
+    const cogIndex = this.cogIndex?.values;
+    if (cogIndex != null && cogIndex.length > 0) {
+      const names: string[] = [];
+      for (const i of cogIndex) {
+        const { name } = files[i];
```
blacha, line 56: This is kinda weirrd, if i > files.length, files[i]  will be undefined and  will throw

```typescript
const file = files[i]
if (files == null) throw new LoggerFatalError(...)
names.push(file.name);
```

blacha: might also be a good idea to use a `Set` then return `[...set.values()]` as it will dedupe files.

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -85,50 +95,52 @@ export class ActionCogCreate extends CommandLineAction {
...
-    const outputExists = await fsa.exists(targetPath);
-    logger.info({ targetPath, outputExists }, 'CogCreate:CheckExists');
-    // Output file exists don't try and overwrite it
-    if (outputExists) {
-      logger.warn({ targetPath }, 'CogCreate:OutputExists');
-      await this.checkJobStatus(job, logger);
-      return;
-    }
+    const names = this.getName(job);
+    if (names == null) return;
```
blacha, line 99: validate names.length?

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -34,7 +35,7 @@ export class ActionCogCreate extends CommandLineAction {
     });
   }
 
-  getName(job: CogJob): string | null {
+  getName(job: CogJob): string[] | null {
```
blacha, line 38: getName*s* ?

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -85,50 +95,52 @@ export class ActionCogCreate extends CommandLineAction {
...
+
+        const tmpTiff = fsa.join(tmpFolder, `${name}.tiff`);
+
+        await buildCogForName(job, name, tmpVrtPath, tmpTiff, logger, isCommit);
+        logger.info({ target: targetPath }, 'CogCreate:StoreTiff');
+        if (isCommit) {
+          await fsa.write(targetPath, createReadStream(tmpTiff));
+          await this.checkJobStatus(job, logger);
+        } else {
+          logger.warn('DryRun:Done');
```
blacha, line 142: add `{name}` ?

packages/cli/src/cli/cogify/batch.job.ts
```diff
@@ -29,34 +30,41 @@ export class BatchJob {
...
-    const memory = 3900 * resDiff;
+    const jobName = BatchJob.id(job, names);
+
+    // Calculate the total size to provision memory
+    let res = 1;
+    for (const name of names) {
+      const tile = TileMatrixSet.nameToTile(name);
+      const alignmentLevels = Projection.findAlignmentLevels(job.tileMatrix, tile, job.source.gsd);
+      // Give 25% more memory to larger jobs
+      res += Math.max(alignmentLevels - MagicAlignmentLevel, 0) * 0.25;
```
blacha, line 58: is this adding a number per names? what is memory for a job with 3 images?

packages/cli/src/cli/cogify/batch.job.ts
```diff
@@ -111,24 +119,43 @@ export class BatchJob {
...
+          logger.info({ targetPath }, 'FileExists');
+          return { names: [name], ok: true };
+        }
+
+        // Calculate the imagery size and chunk the small ones into one job.
+        const size = width / job.output.gsd;
+        if (size < ChunkJobSizeLimit) chunkJob.push(name);
+
+        // Check existence of the job running.
+        const jobName = BatchJob.id(job, [name]);
```
blacha, line 138: I'm not sure this logic is going to work, I think you need to create a list of jobs first. then iterate to see if the jobs exist and then submit them.

as the chunked job isnt going to have the same name as `BatchJob.id(job, [name])` vs `BatchJob.id(job, [a,b,c])`

blacha (changes requested):

packages/cli/src/cli/cogify/batch.job.ts
Wentao-Kuang: I am not 100% sure the logic here. I think we might only need to do this for the `names.lenght === 1`, as others just small images chunk job.

Wentao-Kuang:
@blacha All updated. Thanks.

packages/cli/src/cli/cogify/batch.job.ts
```diff
@@ -29,34 +30,41 @@ export class BatchJob {
...
-   * @param fileName output filename
+   * @param fileNames output filename
    * @returns job id
    */
-  static id(job: CogJob, fileName: string): string {
+  static id(job: CogJob, fileNames: string[]): string {
     // Job names are uncontrolled so hash the name and grab a small slice to use as a identifier
     const jobName = createHash('sha256').update(job.name).digest('hex').slice(0, 16);
-    return `${job.id}-${jobName}-${fileName}`.slice(0, 128);
+    fileNames.sort((a, b) => (a > b ? 1 : -1));
```
blacha, line 39: when they are strings you can use `a.localeCompare(b)`

packages/cli/src/cli/cogify/batch.job.ts
```diff
@@ -111,24 +119,51 @@ export class BatchJob {
...
-        const jobName = BatchJob.id(job, name);
-        const isRunning = runningJobs.get(jobName);
-        if (isRunning) {
-          logger.info({ jobName }, 'JobRunning');
-          return { name, ok: true };
-        }
+    // Prepare chunk job and individual jobs based on imagery size.
+    const jobs: string[][] = [];
+    const chunkJob: string[] = [];
+    for (const file of job.output.files) {
```
blacha, line 125:
```typescript
const MaxChunkJob = 10

function getJobs(files) {
 const jobs = [];
 let chunkJob = [];
 
for (const file of files) {
  const imageSize = file.width / job.output.gsd;
if (imageSize > ChunkJobSizeLimit) {
  jobs.push([job.name])
} else {
  chunkJob.push(job.name);
 if (chunkJob.length > MaxChunk) {
jobs.push(chunkJob)
 chunkJob = [];
}
}
if (chunkJob.length > 0) jobs.push(ChunkJob)
return jobs;

}
```

and then unit test the function

blacha (approved):

PR #2218, opened 2022-05-27
feat(server): bundle basemaps-server cli so its easier to install

blacha:


packages/config/src/json/json.config.ts
```diff
@@ -51,6 +51,7 @@ export class ConfigJson {
     for await (const tsPath of fsa.list(fsa.join(basePath, 'tileset'))) {
       if (!tsPath.endsWith('.json')) continue;
       const ts = await cfg.tileSet(await fsa.readJson(tsPath));
+      // console.log('LoadTileSet', )
```
Wentao-Kuang, line 54: Comment out code.

blacha: good point I was going to add logs to this.

packages/shared/src/cli/base.ts
```diff
@@ -0,0 +1,60 @@
...
+      LogConfig.get().level = 'info';
+    }
+
+    const logger = LogConfig.get().child({ id: CliId });
+    logger.info(CliInfo, 'CliStart');
+    LogConfig.set(logger);
+
+    return super.onExecute();
+  }
+  protected onDefineParameters(): void {
```
Wentao-Kuang, line 46: What is this function for?

blacha: its a required function (I think?) this file was just copy pasted from @basemaps/cli to a central location

Wentao-Kuang (approved):

PR #2222, opened 2022-05-31
feat(server): provide a better error when loading configuration bundles

blacha:


packages/config/src/memory/memory.config.ts
```diff
@@ -7,9 +7,11 @@ import { ConfigProcessingJob } from '../config/processing.job.js';
...
 import { ConfigTileSet, TileSetType } from '../config/tile.set.js';
 import { ConfigVectorStyle } from '../config/vector.style.js';
+import { ulid } from 'ulid';
 
 /** bundle the configuration as a single JSON object */
 export interface ConfigBundled {
+  id: string;
   imagery: ConfigImagery[];
   style: ConfigVectorStyle[];
   provider: ConfigProvider[];
```
Wentao-Kuang, line 17: Do we need multiple Providers here?

blacha: There is no way of knowning how many providers are in the config files so yeah we need multiple, we currently hard code one called "linz" I wonder if we should change it to "default" or something?

Wentao-Kuang (approved):

PR #2223, opened 2022-05-31
feat(lambda-tiler): tiff paths should allow trailing slashes

blacha:


packages/lambda-tiler/src/__test__/tile.set.cache.test.ts
```diff
@@ -66,6 +66,7 @@ o.spec('TileSetCache', () => {
 
       const aTiff = subTileSet.getTiffsForTile({ x: 0, y: 0, z: 0 });
       o(aTiff.length).equals(1);
+      console.log(aTiff[0].source);
```
Wentao-Kuang, line 69: console.log

blacha: Good spotting will remove. I wonder if we should have a lint rule for console.logs

Wentao-Kuang (approved):

PR #2227, opened 2022-05-31
test: increase test timeout to 1,000ms as macosx often timesout

blacha:


Wentao-Kuang (approved):

Wentao-Kuang:
I can see some other test also timeout like `packages/cli/build/cog/__test__/cutline.test.js`

blacha:
> I can see some other test also timeout like `packages/cli/build/cog/__test__/cutline.test.js`

I think anything that touches the file system is quite slow in these mac tests, so we will just need to add more timeout duration to them.

Wentao-Kuang (approved):

PR #2228, opened 2022-06-01
feat: bundle server into a function url

blacha:
- test: increase test timeout to 1,000ms as macosx often timesout
- wip
- wip
- fix: url for public url base
- feat: include a hash of the config


Wentao-Kuang:
Is this still relevant? looks like we already done some of this.

blacha:
Server is now deployed as a function url. this is no longer needed.

PR #2236, opened 2022-06-03
feat(config-cli): Provide a cli for creating temporary server and dump screenshots.

Wentao-Kuang:


packages/config-cli/README.md
```diff
@@ -16,3 +16,11 @@ Dump the screenshots from different host and tag
 ./bin/bmc.js screenshot --host HOST --tag PR-TAG
 
 ```
+
+## Usage -- Screenshots-server
+
+Create a temporary server from a config bundle file and dump the screenshots
+
+```bash
+./bin/bmc.js screenshot-server - config s3://..../config.json.gz
```
blacha, line 25: I think this should just be a argument to `screenshot`

eg 

```
./bin/bmc.js screenshot --config s3://.../config.json.gz
```

packages/config-cli/src/cli/screenshot/screenshot.server.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Bundle Config
+    const configJson = await fsa.readJson<ConfigBundled>(config);
+    const mem = ConfigProviderMemory.fromJson(configJson);
+    Config.setConfigProvider(mem);
+
+    // Create a basemaps server.
+    const port = 5000;
+    const ServerUrl = `http://localhost:${port}`;
+    const BasemapsServer = createServer(logger);
+    BasemapsServer.listen(port, '0.0.0.0', () => {
```
blacha, line 54: this is async you should really `await` for the server to start before starting chrome.

granted chrome startup is going to be way slower than the server but it could create issues if the server startup takes too long

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -58,85 +60,91 @@ export class CommandScreenShot extends CommandLineAction {
...
+      throw e;
     }
+    logger.info({ url, expected: fileName }, 'Page:Load:Done');
+    await page.close();
   }
+}
 
-  async getTileSetId(tileSetId: string, tag: string): Promise<string> {
-    if (tag === 'production') return tileSetId;
+async function getTileSetId(tileSetId: string, tag: string): Promise<string> {
```
blacha, line 130: I think we can remove these now.

if we are pointing at a bundled config there is not `@tag` to worry about any more!

packages/config-cli/README.md
Wentao-Kuang: Might be good to separate to two different cli? I think this will make the cli logic not too complex.

packages/config-cli/src/cli/screenshot/screenshot.server.ts
Wentao-Kuang: `createServer()` is not async function? Do you want me to update to async?

packages/config-cli/src/cli/screenshot/screenshot.ts
Wentao-Kuang: Pr tag removed. We can just point a lambda function url as host to take screenshot for `linz/basemaps-config`.

packages/config-cli/src/cli/screenshot/screenshot.server.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (config == null) throw new Error('Please specify a correct config file for the server.');
+    if (tiles == null) throw new Error('Test tiles not defined.');
+
+    // Bundle Config
+    const configJson = await fsa.readJson<ConfigBundled>(config);
+    const mem = ConfigProviderMemory.fromJson(configJson);
+    Config.setConfigProvider(mem);
+
+    // Create a basemaps server.
+    const port = 5000;
```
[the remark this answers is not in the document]
Wentao-Kuang: Using random free port from 10000 to 11000 now.

packages/config-cli/README.md
blacha: should be pretty simiple logic ?

something like 
```
if (this.config.value) 
 const host = await startServer()
 return await screenshot(host);
}

return await screenshot(this.host.value)
```

they are both doing the same thing, screenshotting some host, so to me they are the same command

packages/config-cli/src/cli/screenshot/screenshot.server.ts
blacha: something like this 

```typescript
await new Promise(resolve => BaseampsServer.listen(port, '0.0.0.0', () => resolve()))
```

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +45,88 @@ export class CommandScreenShot extends CommandLineAction {
...
+    const tiles = this.tiles.value ?? DefaultTestTiles;
+
+    let BasemapsServer: FastifyInstance | undefined = undefined;
+    if (config != null) {
+      const port = await getPort({ port: portNumbers(10000, 11000) });
+      host = `http://localhost:${port}`;
+      BasemapsServer = createServer(logger);
+
+      if (BasemapsServer == null) throw new Error('Failed to Create server with the config File');
+      BasemapsServer.listen(port, '0.0.0.0', () => {
```
blacha, line 82: this is not waiting for this to finish loading.

packages/config-cli/README.md
```diff
@@ -10,9 +10,16 @@ Dump the screenshots from basemaps production
...
 ```bash
-./bin/bmc.js screenshot --host HOST --tag PR-TAG
+./bin/bmc.js screenshot --host HOST
+
+```
+
+Dump the screenshots with config file
+
+```bash
+./bin/bmc.js screenshot - config s3://..../config.json.gz
```
blacha, line 23: --config ?

Wentao-Kuang: Good pick.

packages/config-cli/src/cli/screenshot/screenshot.ts
Wentao-Kuang: Sorry, I forget this. Just fixed.

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +45,92 @@ export class CommandScreenShot extends CommandLineAction {
...
     const logger = LogConfig.get();
+    const config = this.config.value;
+    let host = this.host.value ?? DefaultHost;
+    const tiles = this.tiles.value ?? DefaultTestTiles;
+
+    let BasemapsServer: FastifyInstance | undefined = undefined;
+    if (config != null) {
+      const port = await getPort({ port: portNumbers(10000, 11000) });
+      host = `http://localhost:${port}`;
+      const server = createServer(logger);
```
blacha, line 79: NIT: why create a new var if we could just use `BasemapsServer`

another option would be `const BasemapsServer = this.startServer(config)` which returns `null | Server`

blacha (dismissed):

packages/config-cli/src/cli/screenshot/screenshot.ts
Wentao-Kuang: Not sure why if I pass the `BasemapsServer` into the `await new Promise<void>()` if always return type checking errors to say that `BasemapsServer` could be `undefined` even I have checked not null before. So I create a new var to pass it after.

blacha: yeah if you did this as a function that did `Promise<Server | null>` could be optimised out.

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +47,97 @@ export class CommandScreenShot extends CommandLineAction {
...
-  }
+async function startServer(host: string, port: number, config: string, logger: LogType): Promise<FastifyInstance> {
+  // Bundle Config
+  const configJson = await fsa.readJson<ConfigBundled>(config);
+  const mem = ConfigProviderMemory.fromJson(configJson);
+  Config.setConfigProvider(mem);
+
+  // Start server
+  const server = createServer(logger);
+  server.listen(port, '0.0.0.0', () => {
```
blacha, line 139: this lost its promise again?

Wentao-Kuang: Sorry, I copy from the wrong place. Just fixed.

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +46,100 @@ export class CommandScreenShot extends CommandLineAction {
...
-  }
+async function startServer(host: string, port: number, config: string, logger: LogType): Promise<FastifyInstance> {
+  // Bundle Config
+  const configJson = await fsa.readJson<ConfigBundled>(config);
+  const mem = ConfigProviderMemory.fromJson(configJson);
+  Config.setConfigProvider(mem);
+
+  // Start server
+  const server = createServer(logger);
+  await new Promise<void>((resolve) =>
```
blacha, line 138:
```
return new Promise<FastifyInstance>(r => {
r(server);
});
```

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +46,100 @@ export class CommandScreenShot extends CommandLineAction {
...
-    if (styleId == null) return '';
-    if (tag === 'production') return styleId ?? '';
-
-    const styleIdTagId = `${styleId}@${tag}`;
-    const dbId = Config.Style.id(styleIdTagId);
-    const style = await Config.Style.get(dbId);
-    if (style) return styleIdTagId;
-    return styleId;
-  }
+async function startServer(host: string, port: number, config: string, logger: LogType): Promise<FastifyInstance> {
```
blacha, line 130: why is host passed in just to log it?

could generate host inside the function or log outside the function

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -39,104 +46,100 @@ export class CommandScreenShot extends CommandLineAction {
...
 
   async onExecute(): Promise<void> {
     const logger = LogConfig.get();
+    const config = this.config.value;
+    let host = this.host.value ?? DefaultHost;
+    const tiles = this.tiles.value ?? DefaultTestTiles;
+
+    let BasemapsServer: FastifyInstance | undefined = undefined;
+    if (config != null) {
+      const port = await getPort({ port: portNumbers(10000, 11000) });
```
blacha, line 78: just use any port?

blacha (approved):

PR #2240, opened 2022-06-07
feat(tiler-sharp): extract regions before rescaling them when overzooming

blacha:
- test: add tests for very zoomed in tests
- refactor: remove unused tiles
- feat(tiler-sharp): extract regions before rescaling them when overzooming


packages/sprites/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@basemaps/sprites",
-  "version": "6.28.0",
+  "version": "6.27.0",
```
Wentao-Kuang, line 3: Why do we down version here?

blacha: ahh must have been a bad rebase.

Wentao-Kuang (dismissed):

packages/server/package.json
```diff
@@ -47,7 +47,7 @@
     "@basemaps/landing": "^6.28.1"
   },
   "dependencies": {
-    "sharp": "^0.30.2"
+    "sharp": "^0.30.6"
```
Wentao-Kuang, line 50: No "@types/sharp" required here?

blacha: its not actually used here, but its used as a way of forcing the dependency to be installed.

We bundle the code a single js file. which removes all the deps from it

Wentao-Kuang (approved):

PR #2246, opened 2022-06-08
feat(lambda-tiler): serve assets via /v1/sprites and /v1/fonts

blacha:
Using `/v1/fonts` over `/v1/glyphs` as the uri as that is what tileserver-gl uses

### DIFF
--- a/packages/lambda-tiler/src/routes/__tests__/fonts.test.ts
+++ b/packages/lambda-tiler/src/routes/__tests__/fonts.test.ts
@@ -1,0 +1,68 @@
+import { fsa } from '@chunkd/fs';
+import o from 'ospec';
+import { fontGet, fontList, getFonts } from '../fonts.js';
+import { FsMemory } from './memory.fs.js';
+import { mockRequest } from '../../__tests__/xyz.util.js';
+import { Env } from '@basemaps/shared';
+import { handler } from '../../index.js';
+
+o.spec('listFonts', () => {
+  const memory = new FsMemory();
+  o.before(() => {
+    fsa.register('memory://', memory);
+  });
+
+  o.beforeEach(() => {
+    process.env[Env.AssetLocation] = 'memory://';
+  });
+
+  o.afterEach(() => {
+    delete process.env[Env.AssetLocation];
+    memory.files.clear();
+  });
+
+  o('should list font types', async () => {
+    await Promise.all([
+      fsa.write('memory://fonts/Roboto Thin/0-255.pbf', Buffer.from('')),
+      fsa.write('memory://fonts/Roboto Thin/256-512.pbf', Buffer.from('')),
+      fsa.write('memory://fonts/Roboto Black/0-255.pbf', Buffer.from('')),
+    ]);
+
+    const fonts = await getFonts('memory://fonts/');
+    o(fonts).deepEquals(['Roboto Black', 'Roboto Thin']);
+  });
+
+  o('should return empty list if no fonts found', async () => {
+    const res = await fontList();
+    o(res.status).equals(200);
+    o(res.body).equals('[]');
+  });
+
+  o('should return 404 if no assets defined', async () => {
+    delete process.env[Env.AssetLocation];
+    const res = await fontList();
+    o(res.status).equals(404);
+  });
+
+  o('should return a list of fonts found', async () => {
+    await Promise.all([
+      fsa.write('memory://fonts/Roboto Thin/0-255.pbf', Buffer.from('')),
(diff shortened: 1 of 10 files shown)


packages/lambda-tiler/src/routes/fonts.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function fontList(): Promise<LambdaHttpResponse> {
+  const assetLocation = Env.get(Env.AssetLocation);
+  if (assetLocation == null) return new LambdaHttpResponse(404, 'No Found');
+
+  try {
+    const filePath = fsa.join(assetLocation, '/fonts');
+    const fonts = new Set<string>();
+
+    // TODO use {recursive: false}
+    for await (const fontPath of fsa.list(filePath)) {
```
blacha, line 35: chunkd/fs is currently only recursive listing, which makes this kinda hard!

the next version of chunkd/fs supports `fsa.list(filePath, {recusrive: false})`

packages/lambda-tiler/src/routes/fonts.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (assetLocation == null) return new LambdaHttpResponse(404, 'No Found');
+
+  try {
+    const filePath = fsa.join(assetLocation, path.join('fonts', req.params.fontStack, req.params.range)) + '.pbf';
+    const buf = await fsa.read(filePath);
+    const res = new LambdaHttpResponse(200, 'ok');
+    res.buffer(buf, 'application/x-protobuf');
+    return res;
+  } catch (e: any) {
+    if (e.code === 404) return new LambdaHttpResponse(404, 'No Found');
```
Wentao-Kuang, line 21: 'No Found'. -> 'Not Found'
Same as below.

packages/lambda-tiler/src/routes/fonts.ts
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    const filePath = fsa.join(assetLocation, '/fonts');
+    const fonts = new Set<string>();
+
+    // TODO use {recursive: false}
+    for await (const fontPath of fsa.list(filePath)) {
+      if (!fontPath.endsWith('.pbf')) continue;
+      const basePath = fontPath.slice(filePath.length + 1);
+      const dirName = path.dirname(basePath); // TODO this only works for /a/b.pbf and not /a/b/c.pbf
+      if (dirName.includes('/')) continue;
```
Wentao-Kuang, line 39: Logic here is quite confusing. Are we getting all the fonts directory name without `/`? like `s3://linz-basemaps/font/NewTimes`?

packages/shared/src/const.ts
```diff
@@ -18,6 +18,9 @@ export const Env = {
   /** Public URL base that tiles are served from */
   PublicUrlBase: 'BASEMAPS_PUBLIC_URL',
 
+  /** Location of sprites and glyphs */
+  AssetLocation: 'BASEMAPS_ASSEST_LOCATION',
```
Wentao-Kuang, line 22: Do we need to setup new environment variable in the lambda stack?

packages/lambda-tiler/src/routes/fonts.ts
blacha: Yeah this is annoying as its a recursive list. the next `@chunkd` would make this logic redundant.

basically uou get a list looking like

```
s3://linz-basemaps-static/fonts/Roboto Thin/0-255.pbf
s3://linz-basemaps-static/fonts/Roboto Black/0-255.pbf
```

and you want the output to be ```["Roboto Thin", "Roboto Black"]```

blacha: I need to work on some unit tests for this, I am making a `FsMemory` to write tests against. so it should make more sense once we have tests!

packages/shared/src/const.ts
blacha: We will, I am wanting to use this as a way to bundle the sprites and glyphs into the lambda.zip when we deploy a entire package of it.

So some time in the future we should have basemaps serving fonts/glyphs from `/v1/...` rather than the directly from the static bucket.

packages/lambda-tiler/src/routes/__tests__/fonts.test.ts
```diff
@@ -0,0 +1,68 @@
+import { fsa } from '@chunkd/fs';
+import o from 'ospec';
+import { fontGet, fontList, getFonts } from '../fonts.js';
```
Wentao-Kuang, line 3: Looks good now. Just linting failure to fix.

Wentao-Kuang (approved):

PR #2248, opened 2022-06-13
fix(cli): ensure errors are thrown

blacha:


packages/shared/src/cli/base.ts
```diff
@@ -28,6 +28,15 @@ export abstract class BaseCommandLine extends CommandLineParser {
     description: 'Show extra extra logging detail',
   });
 
+  toolName: string;
+  toolDescription: string;
```
Wentao-Kuang, line 32: Are we using this `toolDescription` anywhere?

blacha: Nope figured for completeness we should have it though

Wentao-Kuang (approved):

PR #2252, opened 2022-06-15
feat(config-cli): new command tool to bundle config json from a config path

Wentao-Kuang:


packages/config-cli/README.md
```diff
@@ -23,3 +23,11 @@ Dump the screenshots with config file
 ./bin/bmc.js screenshot --config s3://..../config.json.gz
 
 ```
+
+## Usage -- Bundle
+
+Bundle config files into config bundle json from a given config path.
+
+```bash
+./bin/bmc.js screenshot --config config/ --bundle config/config.json
```
blacha, line 32: `bundle` ? 

bundling into the output directory could be interesting, maybe make the recommendation just `--output config.json` ?

Wentao-Kuang: Updated.

blacha (dismissed):

packages/config-cli/src/cli/screenshot/screenshot.ts
```diff
@@ -75,9 +75,12 @@ export class CommandScreenShot extends CommandLineAction {
 
     let BasemapsServer: FastifyInstance | undefined = undefined;
     if (config != null) {
-      const port = await getPort({ port: portNumbers(10000, 11000) });
+      const port = await getPort();
       host = `http://localhost:${port}`;
-      BasemapsServer = await startServer(host, port, config, logger);
+      // Force a default url base so WMTS requests know their relative url
+      process.env[Env.PublicUrlBase] = host;
```
blacha, line 81: if this is set we need to respect it we should not be overwriting env vars.

Wentao-Kuang: We can set this back after the script running in finally.

blacha: Proab safer to just use it if it exists the same as basemaps-server cli

Wentao-Kuang: Cool, updated.

PR #2253, opened 2022-06-15
fix(lambda-cog): only assume a read-only role if the current role does not have permission

blacha:


packages/lambda-cog/src/import/imagery.find.ts
```diff
@@ -44,8 +49,19 @@ export class RoleRegister {
...
+    const isAbleToRead = await canRead(path);
+    // If we can directly read/write this path we don't need to register a role for it
+    if (isAbleToRead) return;
+
     const roles = await this.loadRoles();
-    return roles.find((f) => path.startsWith(`s3://${f.bucket}`));
+    const targetRole = roles.find((f) => path.startsWith(`s3://${f.bucket}`));
+    if (targetRole == null) throw new Error(`Failed to read ${path}`);
+
+    fsa.register(
```
blacha, line 60: only register the specific role we need to read the target location

packages/lambda-cog/src/routes/import.ts
```diff
@@ -41,30 +46,47 @@ export async function Import(req: LambdaHttpRequest): Promise<LambdaHttpResponse
...
-  if (role == null) return new LambdaHttpResponse(403, 'Unable to Access the s3 bucket');
   const fileInfo = await findImagery(path);
   if (fileInfo.files.length === 0) return new LambdaHttpResponse(404, 'Imagery Not Found');
+
   const files = fileInfo.files;
-  const numberLimit = Number(Env.get(Env.ImportFilesNumberLimit));
-  const sizeLimit = Number(Env.get(Env.ImportFilesSizeLimitGb));
-  const totalSizeInGB = Math.round((fileInfo.totalSize / Math.pow(1024, 3)) * 100) / 100;
-  if (files.length >= numberLimit || totalSizeInGB >= sizeLimit)
+  const numberLimit = Env.getNumber(Env.ImportFilesNumberLimit, 10_000);
```
blacha, line 53: these could have been undefined whic would lead to `NaN` as the `numberLimit`

Wentao-Kuang (approved):

PR #2255, opened 2022-06-17
feat(config-cli): New cli to import configs from a bundle config json into dynamo

Wentao-Kuang:


packages/config-cli/src/cli/bmc/import.config.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (commit) {
+      const res = await fetch(healthEndpoint);
+      if (!res.ok) throw new Error('Cannot update basemaps is unhealthy');
+    }
+
+    logger.info({ config: config }, 'Import:Load');
+    const configJson = await fsa.readJson<ConfigBundled>(config);
+
+    logger.info({ config: config }, 'Import:Style');
+    for await (const config of configJson.style) this.update(UpdaterType.Style, config, commit);
```
blacha, line 70: I think we can greatly simplify this now.

As this configuration has already been "imported" it has already been validated and the record can basically just be inserted.

I don't think we need a "updater" per type, we just need to query the database for the id, then diff and then insert by id.

packages/config-cli/src/cli/bmc/config/config.update.ts
```diff
@@ -0,0 +1,77 @@
...
+   */
+  constructor(config: S, isCommit: boolean) {
+    this.config = config;
+    this.prefix = config.id.slice(0, 2);
+    this.isCommit = isCommit ? isCommit : false;
+    this.logger = LogConfig.get();
+  }
+
+  getDB(): BasemapsConfigObject<ConfigTileSet | ConfigImagery | ConfigProvider | ConfigVectorStyle> {
+    if (this.prefix === 'im') return Config.Imagery;
```
blacha, line 35: would be good to use the ConfigPrefix enum rather than the strings

packages/config-cli/src/cli/bmc/config/config.update.ts
```diff
@@ -0,0 +1,77 @@
...
+  isCommit: boolean;
+  logger: LogType;
+
+  /**
+   * Class to apply an TileSetConfig source to the tile metadata db
+   * @param config a string or TileSetConfig to use
+   */
+  constructor(config: S, isCommit: boolean) {
+    this.config = config;
+    this.prefix = config.id.slice(0, 2);
```
blacha, line 29: I think there is a `Config.getPrefix(config.id)` im not sure we have set that the prefix is only 2 chars.

Wentao-Kuang: All updated. Thanks.

blacha (approved):

PR #2260, opened 2022-06-21
feat(cli): merge all cli from config-cli package and server package into cli package.

Wentao-Kuang:


packages/cli/README.md
```diff
@@ -1,35 +1,83 @@
...
+```bash
+./bin/bmc.js import --config config.json --commit
+```
+
+## Usage -- Server
+
+Basemaps server expects a folder tree full of configuration, with multiple tilesets and styles configuration files.
+
+```bash
+basemaps-server --config path/to/config/
```
blacha, line 36: I think this should still just be `./bin/bmc.js serve --config ...`

packages/cli/README.md
```diff
@@ -1,35 +1,83 @@
...
 ```bash
-cogify -V cog --job ./cogs/01DYREBEEFFXEPBAYBED2TMAFJ/job.json --name 1-2-3 --commit
+./bin/bmc.js -V cog --job ./cogs/01DYREBEEFFXEPBAYBED2TMAFJ/job.json --name 1-2-3 --commit
 ```
 
 Build all the COGs using aws BATCH
 
 ```bash
-cogify -V batch --job ./cogs/01DYREBEEFFXEPBAYBED2TMAFJ/job.json --commit
+./bin/bmc.js -V batch --job ./cogs/01DYREBEEFFXEPBAYBED2TMAFJ/job.json --commit
```
blacha, line 80: does the batch command still exist?

packages/cli/src/cli/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import { CommandBasemapsServer } from './server/action.server.js';
+
+export class BasemapsConfigCommandLine extends BaseCommandLine {
+  constructor() {
+    super({
+      toolFilename: 'bmc',
+      toolDescription: 'Basemaps config command tools',
+    });
+
+    this.addAction(new ActionCogCreate());
```
blacha, line 18: shall we rename theses to command to be consistent?

packages/cli/src/cli/server/action.server.ts
```diff
@@ -22,37 +28,46 @@ const BaseProvider: ConfigProvider = {
...
-  constructor() {
+export class CommandBasemapsServer extends CommandLineAction {
+  config: CommandLineStringParameter;
+  ignoreConfig: CommandLineFlagParameter;
+  bundle: CommandLineStringParameter;
+  assets: CommandLineStringParameter;
+  port: CommandLineIntegerParameter;
+
+  protected onDefineParameters(): void {
+    this.config = this.defineStringParameter({
```
blacha, line 39: I think this is much easier not to put this here, and just to use the the previious `config = this.defineStringParam`

as there is no need to double define the types.

packages/server/package.json
```diff
@@ -25,20 +25,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "bin": {
-    "basemaps-server": "./bin/basemaps-server.cjs"
```
blacha, line 29: The server still needs this to be bundled, otherwise the server wont be able to start?

packages/cli/src/cli/server/action.server.ts
Wentao-Kuang: Previously we are extend `BaseCommandLine` a instead of `CommandLineAction`, as we need combine these actions together I have to convert it to `CommandLineAction`, and onDefineParameters() must be defined in this class.

packages/server/package.json
Wentao-Kuang: I will test the cli and see how it goes.

packages/cli/README.md
Wentao-Kuang: I don't think so.

packages/server/package.json
Wentao-Kuang: The bundle will require the cli.ts which have been removed.

Wentao-Kuang: @blacha I have revert this back, might leave this later as it is not necessary for the story.

PR #2298, opened 2022-07-03
fix(landing): wait for map to be actually loaded before adding a "#map-loaded" div

blacha:


packages/landing/package.json
```diff
@@ -36,7 +36,7 @@
     "@servie/events": "^3.0.0",
     "@types/geojson": "^7946.0.7",
     "@types/proj4": "^2.5.0",
-    "maplibre-gl": "^1.14.1-rc.2",
+    "maplibre-gl": "^2.1.9",
```
Wentao-Kuang, line 39: That's a big jump of version. Hopefully not broken anything.

Wentao-Kuang: Might be support global view now? https://www.mapbox.com/blog/globe-view/?utm_campaign=email%7Cblast%7Cglobe-launch-weather&utm_content=Globe+view+announcement+1&utm_medium=email_action&utm_source=customer.io

blacha: I think it went 1.15 -> 2.0.0 then 2.1.0 then a bunch of bug fixes, mostly looks to be around types the brokenness.

it did fix some of the loading event problems

packages/landing/src/components/map.switcher.tsx
```diff
@@ -33,7 +34,8 @@ export class MapSwitcher extends Component {
       attributionControl: false,
     });
 
-    this.map.on('load', () => {
+    // console.log('MapSwitcher');
```
Wentao-Kuang, line 37: Comment out code not removed.

packages/landing/src/components/map.tsx
```diff
@@ -128,7 +138,7 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
   }
 
   render(): ComponentChild {
-    const isLayerSwitcherEnabled = Config.map.tileMatrix === GoogleTms && !Config.map.isDebug;
+    const isLayerSwitcherEnabled = false; //Config.map.tileMatrix === GoogleTms && !Config.map.isDebug;
```
Wentao-Kuang, line 141: Might be good just remove this variable and remove the check after?

blacha: ahh was testing something, good find.

Wentao-Kuang (approved):

PR #2304, opened 2022-07-05
feat(lambda-tiler): Convert the sprite and glphys url when serve from assets location.

Wentao-Kuang:


packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -18,6 +18,8 @@ export function convertRelativeUrl(url?: string, apiKey?: string): string {
   if (url == null) return '';
   const host = Env.get(Env.PublicUrlBase) ?? '';
   if (!url.startsWith('/')) return url; // Not relative ignore
+  const assetLocation = Env.get(Env.AssetLocation);
```
blacha, line 21: we shouldnt do this here, we should change the styles to point at /v1/sprites.

Wentao-Kuang:
This is wrong, close it.

PR #2310, opened 2022-07-07
feat(lambda-tiler): allow serving assets from a cotar file

blacha:


packages/lambda-tiler/src/routes/fonts.ts
```diff
@@ -2,6 +2,7 @@ import { Env } from '@basemaps/shared';
 import { fsa } from '@chunkd/fs';
 import { LambdaHttpRequest, LambdaHttpResponse } from '@linzjs/lambda';
 import path from 'path';
+import { serveFromCotar } from '../cotar.cache';
```
Wentao-Kuang, line 5: from '../cotar.cache.js';

packages/lambda-tiler/src/routes/fonts.ts
```diff
@@ -43,6 +48,8 @@ export async function fontList(): Promise<LambdaHttpResponse> {
   const assetLocation = Env.get(Env.AssetLocation);
   if (assetLocation == null) return NotFound;
 
+  if (assetLocation.endsWith('.tar.co')) return serveFromCotar(assetLocation, 'fonts.json', 'application/json');
```
Wentao-Kuang, line 51: I like this, then we will not broken the existing ones until the new one working very well.

blacha: it also means we can still use local files for assets :)

Wentao-Kuang (approved):

PR #2311, opened 2022-07-07
feat(cli): New cli to bundle the assets into cotar file.

Wentao-Kuang:


packages/cli/src/cli/config/action.bundle.assets.ts
```diff
@@ -0,0 +1,102 @@
...
+    const outputTar = output.split('.co')[0];
+
+    const tarBuilder = new TarBuilder(outputTar);
+
+    const files = await fsa.toArray(fsa.list(input));
+    // Ensure files are put into the same order
+    files.sort((a, b) => a.localeCompare(b));
+    logger.info({ output: outputTar, files: files.length }, 'Tar:Create');
+
+    for (const file of files) await tarBuilder.write(file, await fsa.read(file));
```
blacha, line 62: this file name might need to be shortened, what does the tar archive look like?

Wentao-Kuang: Yes, do we need to be short as `assets/fonts/...` or just `fonts/...`?

Wentao-Kuang: Updated. By doing this. We also can move the fonts.json to root directory in cotar file.

blacha: I think just leave the json file as is, and we can just update the server to point at the right spot?

packages/cli/package.json
```diff
@@ -45,6 +45,8 @@
     "@basemaps/sprites": "^6.29.0",
     "@chunkd/fs": "^8.4.0",
     "@cogeotiff/core": "^7.0.0",
+    "@cotar/core": "^5.4.0",
```
blacha, line 48: can you check to see if we are using another version of cotar somewhere else? the yarn lock changes seem excessive!

packages/cli/src/cli/config/action.bundle.assets.ts
Wentao-Kuang: It was in `fonts/fonts.json` after build. Then we move it to `fonts.json` in the cotar. Shall we still keep it in the `fonts/font.json`?

packages/cli/src/cli/config/action.bundle.assets.ts
```diff
@@ -0,0 +1,115 @@
...
+    const outputTar = output.split('.co')[0];
+
+    const tarBuilder = new TarBuilder(outputTar);
+
+    const files = await fsa.toArray(fsa.list(input));
+    // Ensure files are put into the same order
+    files.sort((a, b) => a.localeCompare(b));
+    logger.info({ output: outputTar, files: files.length }, 'Tar:Create');
+
+    for (const file of files) {
```
blacha, line 62: I think the best option would be to use something like
```typescript
const basePath = path.resolve(input)
filePath = file.replace(basePath, '')
```

packages/cli/src/cli/config/action.bundle.assets.ts
blacha: yeah this bundling logic should replicate the excat file system we should change it in the filesystem if we want it changed!

packages/cli/src/cli/config/action.bundle.assets.ts
Wentao-Kuang: Fixed. Thanks you.

packages/cli/package.json
Wentao-Kuang: Fixed.

packages/cli/src/cli/config/action.bundle.assets.ts
blacha: I didnt actually check this, does this actually give us the correct tar output?

blacha (approved):

packages/cli/src/cli/config/action.bundle.assets.ts
Wentao-Kuang: Leading / removed now. Thanks.

PR #2312, opened 2022-07-07
feat(lambda-tiler): prefer using route handler for managing routes

blacha:


packages/lambda-tiler/src/index.ts
```diff
@@ -1,59 +1,52 @@
...
-  } finally {
-    if (slowTimer) clearTimeout(slowTimer);
-    slowTimer = null;
   }
-}
+});
 
-export const handler = lf.http(LogConfig.get());
+handler.router.get('/v1/ping', Ping);
+handler.router.get('/v1/health', Health);
```
Wentao-Kuang, line 43: Do we also need to update the cdk code healthCheck which listen to the `'/health'` endpoint?

blacha: hmm yeah I wonder how this works actually, `/v1/ping` works from the website eg https://basemaps.linz.govt.nz/v1/ping but its only registered as `/ping` 🤯

Wentao-Kuang (approved):

PR #2328, opened 2022-07-18
fix(cli): ensure diff is printed for changes

blacha:


packages/cli/src/cli/config/config.update.ts
```diff
@@ -62,6 +62,8 @@ export class Updater<S extends BaseConfig = BaseConfig> {
   async reconcile(): Promise<boolean> {
     const db = this.getDB();
 
+    if (this.config.id !== 'ts_topo50-gridless-2020-4.23m') return false;
```
Wentao-Kuang, line 65: We never update this layer?

blacha: this was my testing layer, it had a bunch of "undefined" keys added to it.

Wentao-Kuang (approved):

Wentao-Kuang (approved):

PR #2330, opened 2022-07-18
fix(landing): do not duplicate the basemap layers in the layer selector

blacha:


Wentao-Kuang:
Looks good, just some linting to fix.

Wentao-Kuang (approved):

PR #2331, opened 2022-07-19
feat(cli): update import cli to optional output backup database 

Wentao-Kuang:


packages/cli/src/cli/config/action.import.ts
```diff
@@ -73,16 +82,29 @@ export class CommandImport extends CommandLineAction {
...
+      const db = updater.getDB();
+
       const hasChanges = await updater.reconcile();
-      if (hasChanges) this.invalidations.push(updater.invalidatePath());
+      if (hasChanges) {
+        this.invalidations.push(updater.invalidatePath());
+        const oldData = await db.get(config.id);
+        if (oldData != null) this.backupConfig.put(oldData); // No need to backup anything if there is new insert
+      } else {
+        this.backupConfig.put(config);
```
blacha, line 103: we only need to store the changed data.

packages/cli/src/cli/config/action.import.ts
```diff
@@ -73,16 +82,29 @@ export class CommandImport extends CommandLineAction {
...
   update(config: BaseConfig, commit: boolean): void {
     const promise = Q(async (): Promise<boolean> => {
       const updater = new Updater(config, commit);
+      const db = updater.getDB();
+
       const hasChanges = await updater.reconcile();
-      if (hasChanges) this.invalidations.push(updater.invalidatePath());
+      if (hasChanges) {
+        this.invalidations.push(updater.invalidatePath());
+        const oldData = await db.get(config.id);
```
blacha, line 100: this should already be fetched insider `updater.reconcile()`? could we have `updater.dbConfig` or similar to not fetch it multiple times?

packages/cli/src/cli/config/action.import.ts
Wentao-Kuang: I am thinking to store the full config data which will same as the config json as this is not very big and also this make import and re-use much easier.

packages/cli/src/cli/config/action.import.ts
Wentao-Kuang: Fixed.

packages/cli/src/cli/config/config.update.ts
```diff
@@ -56,15 +57,20 @@ export class Updater<S extends BaseConfig = BaseConfig> {
     else return `/v1/tiles/${this.config.id.slice(3)}/*`;
   }
 
+  async getOldData(): Promise<ConfigImagery | ConfigTileSet | ConfigProvider | ConfigVectorStyle | null> {
+    if (this.oldData) return this.oldData;
```
blacha, line 61: this will trigger a fetch multiple times if `oldData` is null.

your best to store oldData fetch as a `Promise` 

```typescript
getOldData(): Promise<...> { // no async
  if (this.oldData) return this.oldData;
  this.oldData = this.getDb().get(this.config.id)` // (no await)
  return this.oldData
}
```

then you can just `const oldData = await this.getOldData()`

Wentao-Kuang: Fixed. Thank you.

blacha (approved):

PR #2359, opened 2022-07-25
feat(infra): create a function url for the lambda-tiler

blacha:


packages/_infra/src/serve/index.ts
```diff
@@ -75,6 +75,7 @@ export class ServeStack extends cdk.Stack {
     });
 
     new cdk.CfnOutput(this, 'LambdaXyzAlb', { value: lb.loadBalancerDnsName });
+    new cdk.CfnOutput(this, 'LambdaXyzUrl', { value: lambda.functionUrl.url });
```
Wentao-Kuang, line 78: Why do we need this functionUrl, I think we are using cloud front for the lambda tiler?

Wentao-Kuang (approved):

packages/_infra/src/serve/index.ts
blacha: We are going to eventually move from ALBs to function urls as they seem to be significantly cheaper.

PR #2361, opened 2022-07-25
ci: create lambda for use with function urls BM-531

blacha:


packages/_infra/src/serve/lambda.tiler.ts
```diff
@@ -36,23 +39,39 @@ export class LambdaTiler extends Construct {
...
       timeout: Duration.seconds(60),
       handler: 'index.handler',
-      code: lambda.Code.fromAsset(CODE_PATH),
+      code,
+      environment,
+      logRetention: RetentionDays.ONE_MONTH,
+    });
+
+    /**
+     * While moving to function URLS create two separate lambda functions
```
Wentao-Kuang, line 59: I'm not sure this is safe to expose lambda directly to the internet?

blacha: The lambda's are already directly exposed to the internet via a ALB?

Wentao-Kuang: But I think VPC is the place for us to limit the access.

PR #2374, opened 2022-07-27
feat(lambda-tiler): Assets provider to get assets from any location.

Wentao-Kuang:


.github/workflows/screenshot.yml
```diff
@@ -31,7 +31,7 @@ jobs:
         with:
           aws-region: ap-southeast-2
           mask-aws-account-id: true
-          role-to-assume: ${{ secrets.AWS_ROLE_NON_PROD }}
+          role-to-assume: ${{ secrets.AWS_ROLE_SCREENSHOT }}
```
blacha, line 34: I made this change too! so this may need a rebase

Wentao-Kuang: Should be good now. I forget to force push after rebase and pull all these commit locally again. Now squashed into one commit.

packages/lambda-tiler/src/index.ts
```diff
@@ -19,6 +20,10 @@ export const handler = lf.http(LogConfig.get());
 handler.router.hook('request', (req) => {
   req.set('name', 'LambdaTiler');
 
+  // Set the asset location for asset provider
+  const assetLocation = Env.get(Env.AssetLocation);
+  assetProvider.set(assetLocation);
```
blacha, line 25: you made this async, it doesnt need to be, but if its async you need to `await`

packages/lambda-tiler/src/util/assets.provider.ts
```diff
@@ -0,0 +1,71 @@
...
+ * This will also set two headers
+ * - Content-Encoding if the file starts with gzip magic
+ * - Content-Type from the parameter contentType
+ */
+export async function serveAssets(
+  req: LambdaHttpRequest,
+  file: string,
+  contentType: string,
+): Promise<LambdaHttpResponse> {
+  const buf = await assetProvider.get(file);
```
blacha, line 61: I think we could improve this logic a bit, if the asset is a `tar.co` then we only need the path to the cotar and the path to the file inside as the cotar should be done by hash, eg `assets-kmnkjsadn1jds83f.tar.co` so the internal files will never change unless the asset path changes .

for local files we cant do this though

packages/lambda-tiler/src/util/assets.provider.ts
```diff
@@ -0,0 +1,71 @@
...
+   * /home/blacha/config/build/assets # Local File
+   * /home/blacha/config/build/assets.tar.co # Local Cotar
+   * s3://linz-baesmaps/assets/ # Remote location
+   * s3://linz-basemaps/assets/assets-b4ff211a.tar.co # Remote Cotar
+   */
+
+  /** Path of the assets location */
+  path: string | undefined;
+
+  async set(path?: string): Promise<void> {
```
blacha, line 21: this shouldn't be async

packages/lambda-tiler/src/index.ts
Wentao-Kuang: Good pick.

packages/lambda-tiler/src/util/config.cache.ts
```diff
@@ -0,0 +1,40 @@
...
+
+class LruConfig {
+  value: ConfigProviderMemory;
+
+  constructor(mem: ConfigProviderMemory) {
+    this.value = mem;
+  }
+
+  get size(): number {
+    return JSON.stringify(this.value.toJson()).length;
```
blacha, line 13: this will be super slow, your best bet is to do it once in the constructor. but it would be even faster to just jut `fsa.read()` and get the length of the bufffer rather than the `readJson`

packages/lambda-tiler/src/util/config.cache.ts
```diff
@@ -0,0 +1,40 @@
...
+      return mem;
+    } catch (e: any) {
+      if (e.code === 404) return null;
+      throw e;
+    }
+  }
+}
+
+/** Approx 10MB around 20 configs(less than 500KB each)*/
+export const CachedConfig = new ConfigCache(10 * 1000 * 1000);
```
blacha, line 40: might be a option to just cache a `number` of configs the swapping LRU could just return `size:1` for every thing then we can just cache ~20-50 configs

packages/lambda-tiler/src/util/config.cache.ts
```diff
@@ -0,0 +1,40 @@
...
+export class ConfigCache {
+  cache: SwappingLru<LruConfig>;
+  constructor(maxSize: number) {
+    this.cache = new SwappingLru<LruConfig>(maxSize);
+  }
+
+  async getConfig(location: string): Promise<ConfigProviderMemory | null> {
+    const existing = this.cache.get(location)?.value;
+    if (existing != null) return existing;
+    try {
```
blacha, line 26: the fetch needs to be sync, so multiple things can call this function

eg doing
```typescript
Promise.all([
 getConfig('a.json')
 getConfig('a.json')
 ])
 ```
 should not create two requests to 'a.json'

File: packages/lambda-tiler/src/util/config.cache.ts (hunk 6 above)
Comment (Wentao-Kuang) [reply packages/lambda-tiler/src/util/config.cache.ts:40]: This is a ready good option. So we don't have to calculate the buffer size.

File: packages/lambda-tiler/src/util/assets.provider.ts [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   *  Load a assets from local path or cotar returning the file back as a LambdaResponse
+   *
+   * This will also set two headers
+   * - Content-Encoding if the file starts with gzip magic
+   * - Content-Type from the parameter contentType
+   */
+  async serve(req: LambdaHttpRequest, file: string, contentType: string): Promise<LambdaHttpResponse> {
+    const config = req.query.get('config');
```
blacha, line 55: config can not be handled here, as this would only affect the "serve" routes, this has to be handled at the top level route handler.

packages/lambda-tiler/src/index.ts
```diff
@@ -20,6 +21,10 @@ export const handler = lf.http(LogConfig.get());
 handler.router.hook('request', (req) => {
   req.set('name', 'LambdaTiler');
 
+  // Set the asset location for asset provider
+  const assetLocation = Env.get(Env.AssetLocation);
```
blacha, line 25: this is where the `req.query.get('config')` logic should live.

Wentao-Kuang: Are we just replace the asset location at the moment, or we can replace the whole configProvider?

blacha: at the moment we can just have a single `Asset` provider.

longer term we will make `req.assets` and `req.config` so the request has it all and we no longer be using global statics.

blacha (approved):

PR #2407, opened 2022-08-04
feat(lambda-tiler): Provide get info api and post tileserver api for arcgis BM-78

Wentao-Kuang:


packages/lambda-tiler/src/arcgis/arcgis.info.ts
```diff
@@ -0,0 +1,22 @@
...
+    fullVersion: '10.1',
+    owningSystemUrl: host,
+  };
+
+  const json = JSON.stringify(info);
+  const data = Buffer.from(json);
+
+  const response = new LambdaHttpResponse(200, 'ok');
+  response.header(HttpHeader.CacheControl, 'public, max-age=604800, stale-while-revalidate=86400');
+  response.buffer(data);
```
blacha, line 19: if you respond as a buffer,  you need to set the `, 'application/json'`

packages/lambda-tiler/src/arcgis/arcgis.info.ts
```diff
@@ -0,0 +1,22 @@
...
+    currentVersion: 10.1,
+    fullVersion: '10.1',
+    owningSystemUrl: host,
+  };
+
+  const json = JSON.stringify(info);
+  const data = Buffer.from(json);
+
+  const response = new LambdaHttpResponse(200, 'ok');
+  response.header(HttpHeader.CacheControl, 'public, max-age=604800, stale-while-revalidate=86400');
```
blacha, line 18: would be good to eTag this too?

packages/lambda-tiler/src/index.ts
```diff
@@ -99,4 +100,6 @@ handler.router.get('/v1/tiles/WMTSCapabilities.xml', wmtsCapabilitiesGet);
 
 // Arcgis Vector
 handler.router.get('/v1/arcgis/rest/services/:tileSet/VectorTileServer', arcgisTileServerGet);
+handler.router.post('/v1/arcgis/rest/services/:tileSet/VectorTileServer', OkResponse);
```
blacha, line 107: Did this work with a OkResponse? or shoudl this just respond with the `arcgisTileServerGet`

packages/lambda-tiler/src/util/response.ts
```diff
@@ -3,3 +3,4 @@ import { LambdaHttpResponse } from '@linzjs/lambda';
 export const NotFound = (): LambdaHttpResponse => new LambdaHttpResponse(404, 'Not Found');
 export const NotModified = (): LambdaHttpResponse => new LambdaHttpResponse(304, 'Not modified');
 export const NoContent = (): LambdaHttpResponse => new LambdaHttpResponse(204, 'No Content');
+export const OkResponse = (): LambdaHttpResponse => new LambdaHttpResponse(200, 'ok');
```
blacha, line 6: You should respond with some sort of body, maybe copy the Error body example where it puts the ids and a message as a JSON response.

You should be able to do something like

`LambdaHttpResponse.ok().json({ id: req.id, correladtionId: req.correlationId, message: 'ok' })`

packages/lambda-tiler/src/index.ts
Wentao-Kuang: Yes. I tested just working with ok.

packages/lambda-tiler/src/util/response.ts
Wentao-Kuang: Fixed. Also, I have updated the name as `OkPostResponse`;

blacha: Does it need to be specific to POST? could just be a `OkResponse` as you could use it as a PUT too?

Wentao-Kuang: I think you are correct. Just reverted it back. Should be ready now.

packages/lambda-tiler/src/util/response.ts
```diff
@@ -1,5 +1,7 @@
-import { LambdaHttpResponse } from '@linzjs/lambda';
+import { LambdaHttpRequest, LambdaHttpResponse } from '@linzjs/lambda';
 
 export const NotFound = (): LambdaHttpResponse => new LambdaHttpResponse(404, 'Not Found');
 export const NotModified = (): LambdaHttpResponse => new LambdaHttpResponse(304, 'Not modified');
 export const NoContent = (): LambdaHttpResponse => new LambdaHttpResponse(204, 'No Content');
+export const OkResponse = (req: LambdaHttpRequest): LambdaHttpResponse =>
+  new LambdaHttpResponse(200, 'ok').json({ id: req.id, correladtionId: req.correlationId, message: 'ok' });
```
blacha, line 7: I think you misspelled  correlation

Wentao-Kuang: :1st_place_medal: Good pick.

blacha: I use a code spell checker so it highlights spelling issues like this 
image

Wentao-Kuang: Wait, did you do reviews in vs code?

blacha: ahh Nope, but it would highlight this issue to you before you make a PR :grin:

blacha (approved):

packages/lambda-tiler/src/util/response.ts
Wentao-Kuang: I got the spell checker as well just not working sometimes?
image

PR #2412, opened 2022-08-08
feat(lambda-tiler): Some unit test for the arcgis api. BM-78

Wentao-Kuang:


Wentao-Kuang:
Interesting, we got an attribution difference in the screenshot. Might because merging https://github.com/linz/basemaps-config/pull/609 when running the screenshot?

blacha (approved):

PR #2420, opened 2022-08-10
fix(cli): include sharp in batch container

blacha:


packages/cli/Dockerfile.aws
```diff
@@ -10,6 +10,9 @@ RUN update-ca-certificates
 RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
 RUN apt-get install -y nodejs
 
+# Install sharp TODO update this when we change sharp versions
+RUN npm install sharp@0.30.7 playwright@1.22.2
```
Wentao-Kuang, line 14: Do we need playwright?

blacha: ahh no we dont!

Wentao-Kuang (approved):

PR #2428, opened 2022-08-11
feat(infra): Grant read access to workflow artifact buckets.

Wentao-Kuang:


packages/_infra/src/config.ts
```diff
@@ -48,6 +52,7 @@ export const BaseMapsDevConfig: BaseMapsConfig = {
   CloudFrontDns: ['dev.basemaps.linz.govt.nz', 'tiles.dev.basemaps.linz.govt.nz'],
   PublicUrlBase: 'https://dev.basemaps.linz.govt.nz',
   AwsRoleConfigBucket: 'linz-bucket-config',
+  WorkflowArtifactsBucket: 'linz-workflow-artifacts',
```
blacha, line 55: should just add it to `CogBucket` ?

Wentao-Kuang: Thanks. Updated.

blacha (approved):

PR #2429, opened 2022-08-12
feat(cli): move screenshot tool into linz/basemaps-screenshot

blacha:
- feat(cli): move screenshot tool into linz/basemaps-screenshot
- refactor: remove playwright

PR #2440, opened 2022-08-17
fix(landing): ensure attribution has loaded before taking screenshots

blacha:


Wentao-Kuang:
Attribution still missing?

blacha:
this was included in #2429

PR #2459, opened 2022-08-18
feat(cli): Add a output for written url to argo workflow output.

Wentao-Kuang:


packages/cli/src/cli/config/action.imagery.config.ts
```diff
@@ -120,13 +128,12 @@ export class CommandImageryConfig extends CommandLineAction {
...
-      logger.info(
-        { path: output, url: `https://basemaps.linz.govt.nz/?config=${configPath}&tileMatrix=NZTM2000Quad${location}` },
-        'ImageryConfig:Done',
-      );
+      const outputPath = fsa.join(path, `basemaps-config-${configJson.hash}.json.gz`);
+      await fsa.writeJson(outputPath, configJson);
+      const configPath = base58.encode(Buffer.from(outputPath));
+      const url = `https://basemaps.linz.govt.nz/?config=${configPath}&i=${name}&tileMatrix=NZTM2000Quad${location}`;
+      logger.info({ path: output, url }, 'ImageryConfig:Done');
+      if (output != null) fsa.write(output, url);
```
blacha, line 136: you need to `await` this

Wentao-Kuang: Good pick. Fixed.

blacha: is there anyway in argo to just set a env var or do we always have to use files for outputs?

blacha (approved):

packages/cli/src/cli/config/action.imagery.config.ts
Wentao-Kuang: We should investigate this later. Currently what I see is just parameter output and artifacts output.

PR #2460, opened 2022-08-18
fix(lambda-tiler): send 408 timeout response rather than timing out.

blacha:


packages/lambda-tiler/src/index.ts
```diff
@@ -20,18 +20,17 @@ import { St } from './util/source.tracer.js';
...
+handler.router.timeoutEarlyMs = 1_000;
+
 handler.router.hook('request', (req) => {
   req.set('name', 'LambdaTiler');
 
   // Reset the request tracing before every request
   St.reset();
 });
 
-let totalRequests = 0;
```
blacha, line 30: linzjs/lambda now counts requests.

packages/lambda-tiler/src/index.ts
```diff
@@ -20,18 +20,17 @@ import { St } from './util/source.tracer.js';
 
 export const handler = lf.http(LogConfig.get());
 
+/** If the request takes too long, respond with a 408 timeout when there is approx 1 second remaining */
+handler.router.timeoutEarlyMs = 1_000;
```
Wentao-Kuang, line 24: Is 1 second enough for attribution?

blacha: its `lambda.maxDuration - 1_000` so its 59,000ms since we have a default duration of 1 minute.

Wentao-Kuang: Haha, I see.

Wentao-Kuang (approved):

PR #2465, opened 2022-08-19
feat(landing): expose config objects to debug page.

blacha:


packages/lambda-tiler/src/routes/config.ts
```diff
@@ -0,0 +1,83 @@
...
+  req.timer.end('tileset:load');
+  if (tileSet == null) return NotFound();
+
+  req.timer.start('imagery:load');
+  let imagery = await config.Imagery.get(config.Imagery.id(req.params.imageryId));
+  req.timer.end('imagery:load');
+
+  if (imagery == null) {
+    const imageryLayer = tileSet.layers.find(
+      (f) => f.name === req.params.imageryId || standardizeLayerName(f.name) === req.params.imageryId,
```
Wentao-Kuang, line 66: Still quite confusing here, so if we cannot file the `req.params.imageryId` we will look for the matching name from the provided tileset? So the `req.params.imageryId` can be both an `imageryId` or `imageryName`?

PR #2473, opened 2022-08-24
feat: switch to aws role provider from chunkd

blacha:


Wentao-Kuang (dismissed):

blacha:
This will also remove the cog import url as that is now being handled by argo. Maybe we want to hold off on merging this for a few days?

Wentao-Kuang (approved):

PR #2504, opened 2022-09-15
feat(cli): Allow empty string as null values in the make-cog cli.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.make.cog.ts
```diff
@@ -85,7 +85,15 @@ export class CommandMakeCog extends CommandLineAction {
     if (name == null) name = source.split('/').filter(Boolean).pop();
     if (name == null) throw new Error('Failed to find imagery set name');
 
-    const tileMatrix: string[] = this.tileMatrix.value ? [this.tileMatrix.value] : ['NZTM2000Quad', 'WebMercatorQuad'];
+    const tileMatrix: string[] = [];
+    if (this.tileMatrix.value && this.tileMatrix.value === 'All') {
```
blacha, line 89: how about a comma separatored list here as all could be interesting?

Wentao-Kuang: Do you mean still passing a string like '"['NZTM2000Quad', 'WebMercatorQuad']" from argo?

Wentao-Kuang: I think it's much easier to make this `All`, because in argo we can use enum which makes a drop down list might be easier to use? However, I haven't found a multi-choice option with a separator for this in argo.  

image

Wentao-Kuang: Updated to support comma separated string and leave the argo workflow as default value of 'NZTM2000Quad,WebMercatorQuad'

blacha: sorry, delayed Response how about `NZTM2000Quad/WebMercatorQuad` as there are lot more tile matrixes than just NZTM/WebMercator

Wentao-Kuang: Updated. Thanks.

packages/cli/src/cli/cogify/action.make.cog.ts
```diff
@@ -85,10 +85,14 @@ export class CommandMakeCog extends CommandLineAction {
     if (name == null) name = source.split('/').filter(Boolean).pop();
     if (name == null) throw new Error('Failed to find imagery set name');
 
-    const tileMatrix: string[] = this.tileMatrix.value ? [this.tileMatrix.value] : ['NZTM2000Quad', 'WebMercatorQuad'];
+    let tileMatrixSets: string[] = [];
+    const tileMatrix = this.tileMatrix.value;
+    if (tileMatrix == null) throw new Error('Please provide valid tile set matrix.');
+    if (tileMatrix.includes('/')) tileMatrixSets = tileMatrixSets.concat(tileMatrix.split(','));
```
blacha, line 91: split "," ?

Wentao-Kuang: Good pick. Updated.

blacha (approved):

PR #2506, opened 2022-09-18
fix(geo): Find tile matrix by identifier without case sensitive. 

Wentao-Kuang:


Wentao-Kuang:
Closed. Let's stay with case sensitive.

PR #2509, opened 2022-09-19
feat(cli):Cog cli support string type of array from argo input.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -65,15 +66,29 @@ export class CommandCogCreate extends CommandLineAction {
...
+            { name, names: files.map((r) => r.name).join(', ') },
+            'Name does not exist inside job',
+          );
+        output.add(name);
+      }
+    }
+
+    const namesStr = this.names?.value;
+    if (namesStr == null) throw new LoggerFatalError({ names }, 'Names cannot be null');
+    const namesArr = namesStr.replace('[', '').replace(']', '').split(',');
```
blacha, line 82: `JSON.parse`

should just use `--names` and have a `parseNames(s: string):string[]` which determines if its JSON like

eg `s.startsWith("[")` then `JSON.parse` ?

blacha (approved):

PR #2515, opened 2022-09-22
feat(cli): New cli to create github pr for the import imagery workflow.

Wentao-Kuang:


packages/cli/src/cli/github/github.ts
```diff
@@ -0,0 +1,168 @@
...
+  isOk = (s: number): boolean => s >= 200 && s <= 299;
+
+  /**
+   * Get branch by name if exists
+   *
+   * @returns {ref} github references or the new created branch
+   */
+  async getBranch(branch: string): Promise<string | undefined> {
+    this.logger.info({ branch }, 'GitHub: Get branch');
+    const response = await this.octokit.rest.repos.getBranch({
```
blacha, line 43: pretty much all of these should fit on one line.

packages/cli/src/cli/github/make.cog.pr.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Prepare aerial tileSet config json
+   */
+  async prepareTileSetConfig(
+    layer: ConfigLayer,
+    tileSet: ConfigTileSetRaster,
+  ): Promise<ConfigTileSetRaster | undefined> {
+    // Set layer zoom level and add to latest order
+    if (layer.name.includes('rural')) {
```
blacha, line 82: if the layer names no longer have rural or urban in them, how does this work?

Can we pull this info from the stac somehow?

Wentao-Kuang: Good point, This happens before when importing `manawatu-whanganui`. This is just create an empty draft PR with the correct title. We need manually edit the file and update the PR.

packages/cli/src/cli/github/github.ts
Wentao-Kuang: Good point. All fixed.

packages/cli/src/cli/github/make.cog.pr.ts
Wentao-Kuang: This been fixed. We are adding the new layer at bottom with following configs.
 ```
layer.minZoom = 32;
layer.category = 'New Aerial Photos';
```

blacha (approved):

PR #2525, opened 2022-10-04
feat(landing): Add the zoomToExtent checkbox for layer selector #BM-666

Wentao-Kuang:


packages/landing/src/components/layer.switcher.dropdown.tsx
```diff
@@ -4,13 +4,13 @@ import { LayerInfo, MapConfig } from '../config.map.js';
 
 export interface LayerSwitcherDropdownState {
   layers?: Map<string, LayerInfo>;
-
+  extend: boolean;
```
blacha, line 7: this should be `extent`? or possibly `zoomToExtent`

Wentao-Kuang: Fixed.

blacha (approved):

PR #2540, opened 2022-10-12
feat(cli): New cli to create overview of imagery BM-628

Wentao-Kuang:


packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,122 @@
...
+    await Promise.all(promises);
+    await pool.close();
+    return;
+  }
+
+  async prepareTiles(tiffs: CogTiff[], maxZoom: number): Promise<void> {
+    for (const tiff of tiffs) {
+      const bbox = tiff.getImage(0).bbox;
+      const geom: Polygon = { type: 'Polygon', coordinates: Bounds.fromBbox(bbox).toPolygon() };
+      const tiles = cover.tiles(geom, { min_zoom: 5, max_zoom: maxZoom }); // TODO: What is min zoom here?
```
Wentao-Kuang, line 103: @blacha Do you know what min zoom should we use here?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+import { filterTiff, MaxConcurrencyDefault } from '../../cog/job.factory.js';
+import { Cutline } from '../../cog/cutline.js';
+import { CogTiff } from '@cogeotiff/core';
+import { createHash } from 'crypto';
+
+// Create tiles per worker invocation
+const WorkerTaskSize = 500;
+const workerUrl = new URL('./tile.generator.js', import.meta.url);
+const threadCount = os.cpus().length / 8;
+const pool = new WorkerRpcPool(threadCount, workerUrl);
```
blacha, line 23: Can you add typing to this

```typescript
export type RpcContract = {
   createTiles(x: something): something;
}

const pool = WorkerRpcPool<RpcContract>
```

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+    this.source = this.defineStringParameter({
+      argumentName: 'SOURCE',
+      parameterLongName: '--source',
+      description: 'Path of source imagery files',
+      required: true,
+    });
+    this.maxZoom = this.defineIntegerParameter({
+      argumentName: 'MAX_ZOOM',
+      parameterLongName: '--max-zoom',
+      description: 'Output of the bundle file',
```
blacha, line 53: this description is wrong.

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  addChildren(qk: string, maxZoom: number): void {
+    if (qk.length >= maxZoom) return;
+    for (const child of QuadKey.children(qk)) {
+      this.tiles.add(child);
+      if (child.length < maxZoom) this.addChildren(child, maxZoom);
+    }
+  }
+
+  async prepareSourceFiles(sources: ChunkSource[]): Promise<TileMatrixSet> {
```
blacha, line 118: why are we reading in all files to just get the tile matrix?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+    const index = await CotarIndexBuilder.create(fd, opts);
+    const indexBinary = await CotarIndexBinary.create(new SourceMemory('index', index.buffer));
+    await TarReader.validate(fd, indexBinary);
+    await fs.writeFile(tarIndexPath, index.buffer);
+
+    // Copy the output into s3 location
+    const output = this.output.value;
+    if (output) {
+      logger.info({ output }, 'CreateOverview: UploadOutput');
+      await fsa.write(fsa.join(output, tarFile), await fsa.read(tarFilePath));
```
blacha, line 172: this should be one file, you shouldnt write the index seperatly.

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+      const jobTiles: JobTiles = { path, files: this.files, tileMatrix: tileMatrix.identifier, tiles: todo };
+      promises.push(pool.run('tile', jobTiles));
+    }
+
+    await Promise.all(promises);
+    await pool.close();
+  }
+
+  async createTar(path: string, logger: LogType): Promise<void> {
+    const tarFile = 'overview.co.tar';
```
blacha, line 149: .tar.co

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,176 @@
...
+    const tarIndex = 'overview.tar.index';
+    const tarFilePath = fsa.join(path, tarFile);
+    const tarIndexPath = fsa.join(path, tarIndex);
+
+    // Create tar file
+    const tiles = await fsa.toArray(fsa.list(`${path}/tiles/`));
+    const cwd = fsa.join(process.cwd(), path);
+    const paths = tiles.map((c) => c.replace(`${cwd}/`, ''));
```
blacha, line 157: you can just strip of `${path}`  we should avoid `process.cwd()`

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: This is just additional `--output` parameter to copy the tar file and index into s3.

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: I need to pass `cwd/${path}` into `tar.cwd` which will make sure the no additional path are compressed into tar file.

blacha: Im not sure I follow,  can you just use the exact same tar creation method as cotar ? 

https://github.com/linz/cotar/blob/master/packages/cli/src/commands/tar.ts#L41

Using `process.cwd()` for anything,  can be super unreliable

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: This is mainly used for preparing `this.files: NamedBounds[] = [];`, and just getting the tileMatrix at the same time.
`this.file` will need to pass in to tile-generator for reference the source files and find bounds intersection with tiles.

blacha: the metadata builder also reads in all the files? doesnt it also create a file list?

I think having these functions modify `this.` state makes it somewhat hard to follow, these could be functions that just return the data?

Wentao-Kuang: It is the different files list from metatdata builder, it is the cogs not the source files. I am prepare a new files with the bounds and sources. 

Cool, I can update the function to return a tilematrix and files instead of using this.files.

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: I am not sure I am follow this? Is the `RpcContract` same as the `WorkerRpc` in the `tile.generator.ts`? So we can directly pass it into pool and remove the `workerUrl` things?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,184 @@
...
+    tiles.sort((a, b) => a.localeCompare(b));
+    for (const file of tiles) await tarBuilder.write(file.slice(file.indexOf('tiles/')), await fsa.read(file));
+
+    // Creating tar index
+    const fd = await fs.open(tarFilePath, 'r');
+    const opts: CotarIndexOptions = { packingFactor: 1.25, maxSearch: 50 }; // Default package rule.
+    const index = await CotarIndexBuilder.create(fd, opts);
+    const indexBinary = await CotarIndexBinary.create(new SourceMemory('index', index.buffer));
+    await TarReader.validate(fd, indexBinary);
+    await fs.writeFile(tarIndexPath, index.buffer);
```
blacha, line 175: is this the same as how cotar cli creates a cotar file?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,184 @@
...
+      const jobTiles: JobTiles = { path, files, tileMatrix: tileMatrix.identifier, tiles: todo };
+      promises.push(pool.run('tile', jobTiles));
+    }
+
+    await Promise.all(promises);
+    await pool.close();
+  }
+
+  async createTar(path: string, logger: LogType): Promise<void> {
+    const tarFile = 'overview.tar.co';
```
blacha, line 158: can we call this overview"s".tar.co  ?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (this.tiles.size < 1) throw new Error('Failed to prepare overviews.');
+
+    logger.info({ source, path }, 'CreateOverview: GenerateTiles');
+    await this.generateTiles(path, sourceFiles.tileMatrix, sourceFiles.files);
+
+    logger.info({ source, path }, 'CreateOverview: CreatingTarFile');
+    await this.createTar(path, logger);
+
+    logger.info({ source, path }, 'CreateOverview: Finished');
+    return;
```
blacha, line 100: this return is not doing anything.

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,183 @@
...
+  addChildren(qk: string, maxZoom: number): void {
+    if (qk.length >= maxZoom) return;
+    for (const child of QuadKey.children(qk)) {
+      this.tiles.add(child);
+      if (child.length < maxZoom) this.addChildren(child, maxZoom);
+    }
+  }
+
+  async prepareSourceFiles(sources: ChunkSource[]): Promise<{ tileMatrix: TileMatrixSet; files: NamedBounds[] }> {
+    let tileMatrix;
```
blacha, line 126: the cogbuilder.build already loads all this metadata.

This is also single threaded.

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    // Create tar file
+    const tiles = await fsa.toArray(fsa.list(`${path}/tiles/`));
+    const tarBuilder = new TarBuilder(tarFilePath);
+    tiles.sort((a, b) => a.localeCompare(b));
+    for (const file of tiles) await tarBuilder.write(file.slice(file.indexOf('tiles/')), await fsa.read(file));
+
+    // Creating tar index
+    const fd = await fs.open(tarFilePath, 'r');
+    const opts: CotarIndexOptions = { packingFactor: 1.25, maxSearch: 50 }; // Default package rule.
```
blacha, line 169: do we need to specify these? especially since they are the defaults?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -0,0 +1,183 @@
...
+    const indexBinary = await CotarIndexBinary.create(new SourceMemory('index', index.buffer));
+    await TarReader.validate(fd, indexBinary);
+    await fd.close();
+    await fs.appendFile(tarFilePath, index.buffer);
+
+    // Copy the output into s3 location
+    const output = this.output.value;
+    if (output) {
+      logger.info({ output }, 'CreateOverview: UploadOutput');
+      await fsa.write(fsa.join(output, tarFile), await fsa.read(tarFilePath));
```
blacha, line 180: We should never read a file fully into memory. use `fsa.stream`

packages/cli/src/cli/overview/tile.generator.ts
```diff
@@ -0,0 +1,122 @@
...
+  const res = await TileComposer.compose({
+    layers,
+    format: ImageFormat.Webp,
+    background: DefaultBackground,
+    resizeKernel: DefaultResizeKernel,
+  });
+  if (res.layers === 0) return;
+
+  // Check and skip if the buffer is empty webp
+  if (res.buffer.byteLength < 215) {
```
blacha, line 116: Thinking about this more, we could run a `libvips`  `stats` function to see if there is any non alpha in the image.

I am unsure if we are creating other empty tiles or if all empty tiles will always create this buffer.

blacha: https://sharp.pixelplumbing.com/api-input#stats

packages/shared/package.json
```diff
@@ -32,7 +32,7 @@
     "@linzjs/metrics": "^6.32.1",
     "aws-sdk": "^2.890.0",
     "entities": "^4.3.0",
-    "pino": "^8.3.0",
+    "pino": "^8.6.1",
```
blacha, line 35: you need to upgrade all versions of pino no this specific one, this is making it so we now have multiple versions of pino installed.

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: I think is not just for the `cogbuilder.builder`, I also need find the `tilematrix` and source `files` for `generateTiles` worker.

packages/shared/package.json
Wentao-Kuang: I think that's the only `pino` in Basemaps. I think the other pino are come from different project like `@linzjs/lambda`.

packages/cli/src/cli/overview/tile.generator.ts
Wentao-Kuang: I have tried to skip the ones with `stat.isOpaque`; Seems it still pick the empty ones. I think this might be the empty images come from which already have a background in it. https://github.com/linz/basemaps/blob/c232694ad1f1186b7b56a8af1def56ef165df8fd/packages/tiler-sharp/src/index.ts#L100

Wentao-Kuang: This the stat for empty images. We might need a smart why to check this? 
`{
  channels: [
    {
      min: 0,
      max: 0,
      sum: 0,
      squaresSum: 0,
      mean: 0,
      stdev: 0,
      minX: 0,
      minY: 0,
      maxX: 0,
      maxY: 0
    },
    {
      min: 0,
      max: 0,
      sum: 0,
      squaresSum: 0,
      mean: 0,
      stdev: 0,
      minX: 0,
      minY: 0,
      maxX: 0,
      maxY: 0
    },
    {
      min: 0,
      max: 0,
      sum: 0,
      squaresSum: 0,
      mean: 0,
      stdev: 0,
      minX: 0,
      minY: 0,
      maxX: 0,
      maxY: 0
    },
    {
      min: 0,
      max: 0,
      sum: 0,
      squaresSum: 0,
      mean: 0,
      stdev: 0,
      minX: 0,
      minY: 0,
      maxX: 0,
      maxY: 0
    }
  ],
  isOpaque: false,
  entropy: 0,
  sharpness: 0,
  dominant: { r: 8, g: 8, b: 8 }
}`

Wentao-Kuang: Maybe we check if all the channels.toString equal to `{ min: 0, max: 0, sum: 0, squaresSum: 0, mean: 0, stdev: 0, minX: 0, minY: 0, maxX: 0, maxY: 0 }` and skip it.

blacha: Hmm, the last channel should be the alpha channel, and it should be 255 (?) for full alpha 

so something like  `image[image.bands - 1].min == 255`  should work?

Wentao-Kuang: For this empty ones, it looks like all four channels are the same, they all be `{ min: 0, max: 0, sum: 0, squaresSum: 0, mean: 0, stdev: 0, minX: 0, minY: 0, maxX: 0, maxY: 0 }`

packages/cli/src/cli/overview/action.create.overview.ts
blacha: 0?

Wentao-Kuang: This is not relavent any more. We don need to create geom to find the tiles. We are using the logic from `CogTiff` now.

packages/cli/src/cli/overview/tile.generator.ts
blacha: then maybe alpha of 0? is full alpha? `image[image.bands - 1].max == 0` ?

Wentao-Kuang: I will try to print all `image[image.bands - 1].max == 0` ones and see if they are all empty channels.

Wentao-Kuang: Looks like working. I have updated that.

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: This been fixed, and get the sources from `metadata.bounds`;

packages/shared/package.json
Wentao-Kuang: I tried to run `yarn upgrade pino` seems not changing `yarn.lock` at all.

blacha (approved):

PR #2543, opened 2022-10-12
feat: searchable layer dropdown

blacha:
Massive lui upgrade 7 major versions! Also moves to react v18

### DIFF
--- a/packages/landing/src/components/layer.switcher.dropdown.tsx
+++ b/packages/landing/src/components/layer.switcher.dropdown.tsx
@@ -1,7 +1,7 @@
-import { Component, ReactNode, Fragment, MouseEventHandler, ChangeEventHandler } from 'react';
+import { ChangeEventHandler, Component, ReactNode } from 'react';
+import Select from 'react-select';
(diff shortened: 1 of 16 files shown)


packages/landing/src/components/layer.switcher.dropdown.tsx
```diff
@@ -43,47 +55,44 @@ export class LayerSwitcherDropdown extends Component<unknown, LayerSwitcherDropd
...
     this.setState({ zoomToExtent: target.checked });
   };
 
-  render(): ComponentChild {
+  render(): ReactNode {
+    const ret = this.makeOptions();
+
     return (
-      <div class="LuiDeprecatedForms">
+      <div className="LuiDeprecatedForms">
```
Wentao-Kuang, line 67: Maybe we should stop using these `LuiDeprecatedForms` later as already upgraded to latest lui.

blacha: I was trying to avoid making too many changes, could be lots of other improvements we could do here.

Wentao-Kuang (approved):

PR #2545, opened 2022-10-14
feat: add overview archive to imagery config

blacha:


packages/config/src/json/json.config.ts
```diff
@@ -252,9 +262,60 @@ export class ConfigJson {
       bounds,
       files,
     };
+
+    output.overviews = await ConfigJson.findImageryOverviews(output);
     this.mem.put(output);
     // Ensure there is also a tile set for each imagery set
```
Wentao-Kuang, line 268: These two lines comment out code seems not useful.

packages/tiler-sharp/src/index.ts
```diff
@@ -63,16 +71,28 @@ export class TileMakerSharp implements TileMaker {
     return this.toImage(format, this.createImage(background).composite(layers));
   }
 
+  /** Can we just serve the image directly from the cotar */
+  private isDirectCotar(ctx: TileMakerContext): boolean {
+    if (ctx.layers.length !== 1) return false;
+    const firstLayer = ctx.layers[0];
+    if (firstLayer.type !== 'cotar') return false;
+    if (this.tileSize !== 256) return false; // TODO this should not be hard coded
+    if (ctx.format !== ImageFormat.Webp) return false; // TODO this should not be hard coded
```
Wentao-Kuang, line 80: Are we able to convert the webp to png or jpeg before serving?

packages/config/src/json/json.config.ts
blacha: I didnt want to modify this, could remove it later?

packages/config/src/json/json.config.ts
```diff
@@ -252,9 +262,60 @@ export class ConfigJson {
...
     // this.mem.put(ConfigJson.imageryToTileSet(output));
     return output;
   }
+
+  static async findImageryOverviews(cfg: ConfigImagery): Promise<ConfigImageryOverview | undefined> {
+    if (cfg.overviews) throw new Error('Overviews exist already for config: ' + cfg.id);
```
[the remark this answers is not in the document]
blacha: it prevents us accidentally overwriting the overview config if it exists, wonder if it will ever be hit.

packages/tiler-sharp/src/index.ts
blacha: this is only checking if we can directly serve the image back to the user without actually modifying it.

so using this path the server never even has to uncompress the webp!

packages/config/src/config/imagery.ts
```diff
@@ -21,4 +21,23 @@ export interface ConfigImagery extends BaseConfig {
...
   /** list of file basenames and their bounding box */
   files: NamedBounds[];
+
+  /** Separate overview cache */
+  overviews?: ConfigImageryOverview;
+}
+
+export interface ConfigImageryOverview {
+  /** Path to overview archive */
+  path: string;
```
Wentao-Kuang, line 31: This should be relative path?

Wentao-Kuang: Can we update the comment to say this is a relative path from the imagery.uri?

Wentao-Kuang (approved):
Looks fine, and tested good. Maybe just one comment change. Otherwise looks good to me.

packages/config/src/config/imagery.ts
blacha: interesting thought, maybe we should check if it starts with `./` then join it or allow it to be absolute? `s3://foo/bar/overviews`

Wentao-Kuang: That would be better. I would like to have the option to put overview in different location.

PR #2574, opened 2022-11-09
fix(cli): Remove the experimental feature and create absolute path url for the worker.

Wentao-Kuang:
import.meta.url not support for cjs as experimental feature. So I need to create the URL with file:${absolute path}


Wentao-Kuang:
Close this one. We want to split the tile generator task as argo tasks.

PR #2575, opened 2022-11-11
feat(cli): Refactoring create overview cli into multiple tasks for argo workflow. 

Wentao-Kuang:
Remove the use of @wtrpc/core as `import.meta.url` not support for cjs. 
Not using the argo nodes to spilt the tasks.


Wentao-Kuang:
Closed. We are using single thread for overview cli.

PR #2590, opened 2022-11-16
feat(cli): create WMTSCapabilties.xml for the overviews cotar

blacha:


packages/cli/src/cli/overview/overview.wmts.ts
```diff
@@ -0,0 +1,37 @@
...
+import { ImageFormat, TileMatrixSet } from '@basemaps/geo';
+import { WmtsCapabilities } from '@basemaps/lambda-tiler/build/wmts.capability.js';
+
+export function createOverviewWmtsCapabilities(
+  tileMatrix: TileMatrixSet,
+  maxZoom: number,
+  title = 'cotar-overviews',
+): string {
+  const fakeLayer: ConfigLayer = { [tileMatrix.projection.code]: '', title, name: 'cotar-overviews' };
+  const tileSet: ConfigTileSetRaster = {
```
Wentao-Kuang, line 11: Can we using the `--source` path or an optional id to get this tile set config from dynamoDB?

blacha: most of the time it wont exist inside of dynamodb.

I think we may want a  `--title` added to the CLI at  some stage.

Wentao-Kuang: Cool, this might not need when using for standerised imagery.

Wentao-Kuang (dismissed):

Wentao-Kuang (approved):

PR #2597, opened 2022-11-17
fix(cli): remove the insert config in create-cog cli.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -208,10 +207,6 @@ export class CommandCogCreate extends CommandLineAction {
 
     const cfg = getDefaultConfig();
     if (expectedTiffs.size === 0) {
-      // Insert Imagery and TileSet Config
-      await insertConfigImagery(cfg, job, logger);
-      await insertConfigTileSet(cfg, job, logger);
-
       // Update job status if this is the processing job.
```
blacha, line 210: all of the processingId stuff needs to go too.

Wentao-Kuang: Yes, I will make another pr to remove all the processId.

blacha (approved):

PR #2612, opened 2022-11-30
feat(cli): Create-overview determine max zoom overview from the gsd

Wentao-Kuang:


packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -18,6 +18,7 @@ import { JobTiles, tile } from './tile.generator.js';
 import { SimpleTimer } from './timer.js';
 
 const DefaultMaxZoom = 15;
+const MaxNumberTiles = 250000;
```
blacha, line 21: could we add a sentence explaining what these numbers do?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -115,6 +114,7 @@ export class CommandCreateOverview extends CommandLineAction {
         qk = QuadKey.parent(qk);
       }
     }
+    if (tiles.size > MaxNumberTiles) this.prepareTiles(files, maxZoom - 1);
```
blacha, line 117: return?

blacha: also why is this async?

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -126,8 +126,7 @@ export class CommandCreateOverview extends CommandLineAction {
     }
   }
 
-  async getTileMatrix(sources: ChunkSource[]): Promise<TileMatrixSet> {
-    const tiff = await CogTiff.create(sources[0]);
+  async getTileMatrix(tiff: CogTiff): Promise<TileMatrixSet> {
```
blacha, line 129: this doesnt need to be async the tiff is `tiff.init(true)` before.

(you are initing the tiff twice)

packages/cli/src/cli/overview/action.create.overview.ts
```diff
@@ -55,20 +51,23 @@ export class CommandCreateOverview extends CommandLineAction {
...
     const st = new SimpleTimer();
     logger.debug({ source }, 'CreateOverview:ListTiffs');
     const tiffList = (await fsa.toArray(fsa.list(source))).filter(filterTiff);
     const tiffSource = tiffList.map((path: string) => fsa.source(path));
+
     logger.info({ source, duration: st.tick() }, 'CreateOverview:ListTiffs:Done');
 
     logger.debug({ source }, 'CreateOverview:PrepareSourceFiles');
-    const tileMatrix = await this.getTileMatrix(tiffSource);
+    const tiff = await CogTiff.create(tiffSource[0]);
```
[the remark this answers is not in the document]
Wentao-Kuang: We have a `filterTiff` for the file lists. So I am ok if the script just die here if the not tiff?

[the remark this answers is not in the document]
blacha: we should let it  die here, but with a nice error message, stack traces are not super helpful to users

packages/cli/src/cli/overview/action.create.overview.ts
Wentao-Kuang: I need this to run `await tiff.getImage(0).loadGeoTiffTags();`, before getting the espg. The just tiff.init now working. 
image

blacha: The tiff.init(true) is called in the other function you should pull that out to the same place as the new CogTiff

blacha (approved):

PR #2622, opened 2022-12-08
fix(cli): Create config to parse the title back to imagery name

Wentao-Kuang:


packages/shared/src/util.ts
```diff
@@ -21,6 +21,20 @@ export function titleizeImageryName(name: string): string {
...
 
+/**
+ * Make a tile imagery title as imagery name
+ * @example
+ *  'Tasman rural 2018-19 0.3m' => 'tasman_rural_2018-19_0-3m'
+ */
+export function nameImageryTitle(title: string): string {
+  return title
+    .replace(/ /g, '_')
+    .replace('.', '-')
```
blacha, line 32: this only does the first `.` might want to remove all `.` 's

packages/shared/src/util.ts
```diff
@@ -21,6 +21,20 @@ export function titleizeImageryName(name: string): string {
...
+    .replace(/[^a-zA-Z0-9-_]/g, '')
```
blacha, line 33: I wonder if there is a way to preserve the utf8 macrons here, eg `Manawatū`

Wentao-Kuang: Not sure yet. I will make a test for this and see how to keep it. Might just working.

Wentao-Kuang: Add `À-ȕ` in the filter to keep these macrons.

packages/shared/src/util.ts
Wentao-Kuang: All fixed, thank you.

packages/shared/src/util.ts
blacha: How far does that reach? will this will cover all macrons used in NZ?

Wentao-Kuang: image
I think it covers all the `a e i o u` I think.

blacha (approved):

PR #2625, opened 2022-12-12
feat(landing): Hidden overflow title in the debug page.

Wentao-Kuang:


packages/landing/static/index.css
```diff
@@ -169,6 +169,8 @@ button {
 
 .debug__value {
   display: flex;
+  overflow: hidden;
+  white-space:nowrap;
```
blacha, line 173: nit: should be a ` ` between the : and the word, odd do we not have a linter/formatter for this?

Wentao-Kuang: Good pick.

blacha (approved):

PR #2629, opened 2022-12-15
feat(tiler): Exclude layers from style json. BM-730

Wentao-Kuang:


packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -101,6 +101,7 @@ export async function tileSetToStyle(
 export async function styleJsonGet(req: LambdaHttpRequest<StyleGet>): Promise<LambdaHttpResponse> {
   const apiKey = Validate.apiKey(req);
   const styleName = req.params.styleName;
+  const excludeLayers = req.query.getAll('e');
```
blacha, line 104: could we have something slightly more readable than `e` ? 

`layerExclude=` or maybe even `exclude=`?

packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -112,6 +113,13 @@ export async function styleJsonGet(req: LambdaHttpRequest<StyleGet>): Promise<La
     if (tileSet == null) return NotFound();
     if (tileSet.type !== TileSetType.Raster) return NotFound();
     return tileSetToStyle(req, tileSet, apiKey);
+  } else {
+    const layers = styleConfig.style.layers;
+    styleConfig.style.layers = [];
+    for (const layer of layers) {
+      if (excludeLayers.find((l) => l.toLowerCase() === layer.id.toLowerCase())) continue;
```
blacha, line 120: this is not very efficient, granted the layer array is quite small..

you could however do something like

```javascript
 const excluded =  new Set(excludeLayers)

for (const layer of layers) { 
  if (excluded.has(l.id)) continue; 
  ....
}
```

packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -112,10 +115,20 @@ export async function styleJsonGet(req: LambdaHttpRequest<StyleGet>): Promise<La
...
+    }
   }
 
   // Prepare sources and add linz source
-  const style = convertStyleJson(styleConfig.style, apiKey, ConfigLoader.extract(req));
+  const style = convertStyleJson(
+    styleConfig.style,
+    apiKey,
+    ConfigLoader.extract(req),
+    layers.length > 0 ? layers : undefined,
```
blacha, line 130: if you exclude everything, you get all the layers? but otherwise there will alwys be a layer array with everything in?

how about passing the filter into the function instead? then the return could just be `style.layers.filter(f => !excluded.has(f.id.toLowerCase()))`

Wentao-Kuang: Just fixed.

blacha (approved):

PR #2640, opened 2023-01-12
feat(cli): Add Concurrency for the make cog cli.

Wentao-Kuang:
Changes

- Support concurrency in cog creation cli.
- Make the chunkJobUnit = concurrency*maxChunkUnit to chunk the more jobs with multi concurrency.
- Fix create-cog to support local source files.

### DIFF
--- a/packages/cli/src/cli/cogify/action.cog.ts
+++ b/packages/cli/src/cli/cogify/action.cog.ts
@@ -118,7 +118,7 @@ export class CommandCogCreate extends CommandLineAction {
 
     const tmpFolder = await makeTempFolder(`basemaps-${job.id}-${CliId}`);
 
-    const Q = pLimit(this.concurrency?.value ? this.concurrency.value : DefaultConcurrency);
+    const Q = pLimit(this.concurrency?.value ?? DefaultConcurrency);
 
     try {
       Promise.all(
--- a/packages/cli/src/cli/cogify/action.cog.ts
+++ b/packages/cli/src/cli/cogify/action.cog.ts
@@ -4,6 +4,7 @@ import {
   CommandLineAction,
   CommandLineFlagParameter,
   CommandLineIntegerListParameter,
+  CommandLineIntegerParameter,
   CommandLineStringListParameter,
   CommandLineStringParameter,
 } from '@rushstack/ts-command-line';
@@ -18,12 +19,16 @@ import { Gdal } from '../../gdal/gdal.js';
 import { makeTempFolder, makeTiffFolder } from '../folder.js';
 import path from 'path';
 import { prepareUrl } from '../util.js';
+import pLimit from 'p-limit';
+
+const DefaultConcurrency = 2;
 
 export class CommandCogCreate extends CommandLineAction {
   private job?: CommandLineStringParameter;
   private name?: CommandLineStringListParameter;
   private commit?: CommandLineFlagParameter;
   private cogIndex?: CommandLineIntegerListParameter;
+  private concurrency?: CommandLineIntegerParameter;
 
   public constructor() {
     super({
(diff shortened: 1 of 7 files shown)


packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -113,11 +118,17 @@ export class CommandCogCreate extends CommandLineAction {
 
     const tmpFolder = await makeTempFolder(`basemaps-${job.id}-${CliId}`);
 
+    const Q = pLimit(this.concurrency?.value ? this.concurrency.value : DefaultConcurrency);
```
blacha, line 121: Use `??`

packages/cli/src/gdal/gdal.docker.ts
```diff
@@ -51,7 +51,7 @@ export class GdalDocker extends GdalCommand {
   /** this could contain sensitive info like AWS access keys */
   private async getDockerArgs(): Promise<string[]> {
     const DOCKER_CONTAINER = Env.get(Env.Gdal.DockerContainer) ?? 'osgeo/gdal';
-    const DOCKER_CONTAINER_TAG = Env.get(Env.Gdal.DockerContainerTag) ?? 'ubuntu-small-3.3.0';
+    const DOCKER_CONTAINER_TAG = Env.get(Env.Gdal.DockerContainerTag) ?? 'ubuntu-small-3.6.1';
```
blacha, line 54: This needs to be updated in the docketfile too

packages/cli/src/cli/cogify/action.make.cog.ts
```diff
@@ -86,6 +87,12 @@ export class CommandMakeCog extends CommandLineAction {
       description: 'Aligned level between resolution and cog',
       required: false,
     });
+    this.concurrency = this.defineIntegerParameter({
+      argumentName: 'CONCURRENCY',
+      parameterLongName: '--concurrency',
+      description: 'Number of concurrency create cog jobs',
+      required: false,
+    });
```
ajacombs, line 95: Would it be better to set the default concurrency value here via `defaultValue`, rather than in a constant on line 24? That would mean we wouldn't need to check if it had been supplied every time it is used.

Wentao-Kuang: Very good point. This default value won't be set into `this.concurrency.value` if this parameter not used, the use case is `this.concurrency.defaultValue`, so we still need to check exists and set `this.concurrency.defaultValue` in not exists.

packages/cli/src/gdal/gdal.docker.ts
Wentao-Kuang: Just updated. Thank you.

packages/cli/src/cli/cogify/action.make.cog.ts
blacha: > This default value won't be set into this.concurrency.value if this parameter not used, the use case is this.concurrency.defaultValue

That is just a typing problem, and one of the reasons I think we should change our cli parser.

if no value is provided, the `this.concurrency.value` will be `defaultValue`

packages/cli/src/cli/cogify/action.cog.ts
```diff
@@ -113,11 +118,17 @@ export class CommandCogCreate extends CommandLineAction {
...
     try {
-      for (const name of names) {
-        const tiffJob = await CogStacJob.load(jobLocation);
-        await this.processTiff(tiffJob, name, tmpFolder, isCommit, logger.child({ tiffName: name }));
-      }
+      Promise.all(
+        Array.from(names).map((name) =>
+          Q(async () => {
+            const tiffJob = await CogStacJob.load(jobLocation);
+            this.processTiff(tiffJob, name, tmpFolder, isCommit, logger.child({ tiffName: name }));
```
blacha, line 128: this should be awaited?

Wentao-Kuang: Good point, just fixed.

blacha (approved):

PR #2645, opened 2023-01-16
ci: publish major/minor tags on release

blacha:


.github/workflows/containers.yml
```diff
@@ -34,11 +34,16 @@ jobs:
         id: buildx
         uses: docker/setup-buildx-action@v2
 
-      - name: Set GIT_VERSION
+      - name: Setup tags
         id: version
         run: |
           GIT_VERSION=$(git describe --tags --always --match 'v*')
-          echo "::set-output name=version::${GIT_VERSION}"
```
blacha, line 41: `::set-output` is going to be removed in may! :exploding_head: 

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Wentao-Kuang (approved):

PR #2654, opened 2023-01-18
feat(cli): Shuffle cogs randomly before chunk them.

Wentao-Kuang:


packages/cli/src/cli/cogify/batch.job.ts
```diff
@@ -193,7 +193,10 @@ export class BatchJob {
     const jobs: string[][] = [];
     let chunkJob: string[] = [];
     let chunkUnit = 0; // Calculate the chunkUnit based on the size
-    for (const file of job.output.files) {
+
+    // Shuffle the array randomly before chunk job.
+    const shuffledFiles = job.output.files.slice().sort(() => Math.random() - 0.5);
```
blacha, line 198: If you shuffle outside of this function, then you can leave the tests.

also this is not a very random shuffle,   this is a cool visualization of the lack of randomness https://bost.ocks.org/mike/shuffle/compare.html

blacha:
Lets not shuffle these for now.

PR #2655, opened 2023-01-19
feat(cli): Remove the nztm aligned level offset adjustment.

Wentao-Kuang:


packages/cli/src/cli/cogify/action.make.cog.ts
```diff
@@ -192,7 +192,7 @@ export class CommandMakeCog extends CommandLineAction {
         id,
         projection: Epsg.Nztm2000,
         resampling,
-        alignedLevel: this.adjustAlignedLevel(tileMatrix, alignedLevel),
+        alignedLevel: alignedLevel,
```
blacha, line 195: NIT: you dont need to specify  `: alignedLevel`

Wentao-Kuang: Ahh good point. Updated.

blacha (approved):

PR #2662, opened 2023-01-29
feat: filter layers by date

blacha:
This adds `?date[before]=` and `?date[after]=` to a number of routes to filter the imagery down to a specific date rate.

These dates are inclusive so `?date[before]=2022` its a `less than or equal`.

### DIFF
--- a/packages/lambda-tiler/src/util/config.loader.ts
+++ b/packages/lambda-tiler/src/util/config.loader.ts
@@ -14,9 +14,9 @@ export class ConfigLoader {
   static async getDefaultConfig(): Promise<BasemapsConfigProvider> {
     const config = getDefaultConfig();
     if (config.assets == null) {
-      // const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
-      // if (cb == null) throw new LambdaHttpResponse(400, 'Unable to get lastest config bundle for asset.');
-      // config.assets = cb.assets;
+      const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
+      if (cb == null) throw new LambdaHttpResponse(400, 'Unable to get lastest config bundle for asset.');
+      config.assets = cb.assets;
     }
     return config;
   }
--- a/packages/lambda-tiler/src/util/config.loader.ts
+++ b/packages/lambda-tiler/src/util/config.loader.ts
@@ -14,9 +14,9 @@ export class ConfigLoader {
   static async getDefaultConfig(): Promise<BasemapsConfigProvider> {
     const config = getDefaultConfig();
     if (config.assets == null) {
-      const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
-      if (cb == null) throw new LambdaHttpResponse(400, 'Unable to get lastest config bundle for asset.');
-      config.assets = cb.assets;
+      // const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
+      // if (cb == null) throw new LambdaHttpResponse(400, 'Unable to get lastest config bundle for asset.');
+      // config.assets = cb.assets;
     }
     return config;
   }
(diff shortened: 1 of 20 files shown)


packages/lambda-tiler/src/util/config.loader.ts
```diff
@@ -14,9 +14,9 @@ export class ConfigLoader {
   static async getDefaultConfig(): Promise<BasemapsConfigProvider> {
     const config = getDefaultConfig();
     if (config.assets == null) {
-      const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
-      if (cb == null) throw new LambdaHttpResponse(400, 'Unable to get lastest config bundle for asset.');
-      config.assets = cb.assets;
+      // const cb = await config.ConfigBundle.get(config.ConfigBundle.id('latest'));
```
Wentao-Kuang, line 17: Why do we comment out this? config assets is broken?

packages/lambda-tiler/src/util/filter.ts
```diff
@@ -0,0 +1,51 @@
...
+  const dateBeforeQuery = req.query.get(FilterNames.DateBefore);
+
+  if (dateAfterQuery == null && dateBeforeQuery == null) return layers;
+  const dateAfter = parseDateAsIso(dateAfterQuery);
+  const dateBefore = parseDateAsIso(dateBeforeQuery);
+
+  return layers.filter((l) => {
+    if (l.title == null) return false;
+    const yearRange = extractYearRangeFromTitle(l.title);
+    if (yearRange == null) return false;
```
Wentao-Kuang, line 44: Does the date range slider will ignore all the layers without title or year range? Shall we just make sure all the layers to have a validate title with year in it when we bundle the config and validate it?

packages/lambda-tiler/src/wmts.capability.ts
```diff
@@ -29,7 +29,7 @@ export interface WmtsCapabilitiesParams {
   /** List of tile matrixes to output */
   tileMatrix: TileMatrixSet[];
   /** Should WMTS Layers be created for each imagery set inside this tileSet */
-  isIndividualLayers: boolean;
+  // isIndividualLayers: boolean;
```
Wentao-Kuang, line 32: Comment out code.

packages/lambda-tiler/src/wmts.capability.ts
```diff
@@ -66,17 +68,19 @@ export class WmtsCapabilities {
...
   maxZoom = 32;
+  layers: ConfigLayer[] | null | undefined;
 
   constructor(params: WmtsCapabilitiesParams) {
     this.httpBase = params.httpBase;
     this.provider = params.provider;
     this.tileSet = params.tileSet;
     this.config = params.config;
-    this.isIndividualLayers = params.isIndividualLayers;
+    // this.isIndividualLayers = params.isIndividualLayers;
```
Wentao-Kuang, line 78: comment out code.

packages/shared/src/util.ts
```diff
@@ -46,6 +46,20 @@ export function extractYearRangeFromName(name: string): [number, number] {
...
+  const yearChunk = t.slice(t.lastIndexOf(' ') + 1).trim(); // "(2019)" or "(2019-2020)"
+  if (yearChunk == null) return null;
+  if (!yearChunk.endsWith(')') || !yearChunk.startsWith('(')) return null;
+
+  const chunks = yearChunk
+    .slice(1, yearChunk.length - 1)
+    .split('-')
+    .map(Number);
+  if (chunks.length === 1 && !isNaN(chunks[0])) return chunks as [number];
+  else if (chunks.length === 2 && !isNaN(chunks[0]) && !isNaN(chunks[0])) return chunks as [number, number];
```
Wentao-Kuang, line 59: The second `!isNaN(chunks[0]` should be `!isNaN(chunks[1]`.

packages/lambda-tiler/src/util/config.loader.ts
blacha: running a server without this commented out, causes the server to crash unless you specify a assets and you dont need the assets for most actions of the server.

but this shouldnt be left commented out.

packages/lambda-tiler/src/util/filter.ts
blacha: this is only applied when the `date[before]` or `date[after]` is found

Wentao-Kuang (approved):

Wentao-Kuang (dismissed):

Wentao-Kuang (approved):

PR #2671, opened 2023-02-06
feat(config): Support disable flag in the config layers.

Wentao-Kuang:


packages/config/src/config/tile.set.ts
```diff
@@ -26,6 +26,9 @@ export interface ConfigLayer extends Partial<Record<EpsgCode, string>> {
 
   /** Max zoom to show the layer @default 30 */
   maxZoom?: number;
+
+  /** Disable the layer from the TileSet @default false*/
+  disable?: boolean;
```
blacha, line 31: I think we should use `disabled` not `disable` 

`disabled` is what all the HTML elements use and it has a slightly different dictionary meaning

packages/config/src/json/json.config.ts
```diff
@@ -171,6 +171,7 @@ export class ConfigJson {
     // Map the configuration sources into imagery ids
     for (const l of ts.layers) {
       const layer = { ...l };
+      if (layer.disable) layer.minZoom = 32;
```
blacha, line 174: We should support  `disabled` in the `ConfigImagery` and shouldn't need to convert this back to `minZoom` we then just need a check for `disabled` in the tile rendering pipeline.

Wentao-Kuang: Good point, I thought about this before. However, I believe this `disabled` flag is just for the `ConfigTileSet.ConfigLayer` only. I think the `ConfigImagery` for mainly used for individual imagery and also don't have any zoom level settings? The other one is Virtual `ConfigTileSet` which used for individual as well and always set as minZoom:0 and maxZoom:32.

Wentao-Kuang: I have removed the set `minzoom:32` and filter out the layers have disabled flag in the api.

blacha (dismissed):

packages/lambda-tiler/src/routes/attribution.ts
```diff
@@ -92,6 +92,7 @@ async function tileSetAttribution(
   const host = await config.Provider.get(config.Provider.id('linz'));
 
   for (const layer of filteredLayers) {
+    if (layer.disabled) continue;
```
blacha, line 95: When we make individual tile sets do we turn disabled off?

Wentao-Kuang: I think individual TileSet are created with hard coded values(minzoom:0, maxzoom:32) and without disable flag in it.

packages/config/src/json/parse.tile.set.ts
```diff
@@ -32,6 +32,7 @@ const zLayerConfig = z
     3857: z.string().optional(),
     minZoom: zZoom.optional(),
     maxZoom: zZoom.optional(),
+    disable: z.boolean().optional(),
```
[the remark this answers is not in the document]
Wentao-Kuang: Very good point, I will fix that now and test it with bundle config.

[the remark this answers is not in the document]
Wentao-Kuang: Fixed and tested with bundle and import cli. The individual tilesets are hard coded `(minzoom:0, maxzoom:32)` when creating them.

blacha (approved):

PR #2677, opened 2023-02-09
feat(config): generate configuration from a folder of tiffs

blacha:
This allows runing the server from a local collection of imagery with no configuration


Wentao-Kuang (approved):
Looks good to me.

PR #2678, opened 2023-02-09
feat(lambda-tiler): simple cli to trace the rendering of a tile

blacha:


packages/lambda-tiler/src/cli/render.tile.ts
```diff
@@ -0,0 +1,38 @@
+import { ConfigProviderMemory, getAllImagery } from '@basemaps/config';
+import { initConfigFromPath } from '@basemaps/config/build/json/tiff.config.js';
+import { Epsg, ImageFormat, Nztm2000QuadTms } from '@basemaps/geo';
+import { LogConfig, setDefaultConfig } from '@basemaps/shared';
+import { fsa } from '@chunkd/fs';
+import { LambdaHttpRequest, LambdaUrlRequest, UrlEvent } from '@linzjs/lambda';
+import { Context } from 'aws-lambda';
+import { TileXyzRaster } from '../routes/tile.xyz.raster.js';
+
+const target = `/home/blacha/tmp/basemaps/`;
```
Wentao-Kuang, line 10: Can we covert this as a cli tool? And leave these as parameters?

blacha: I didn't want to include a full cli library into the tiler package, but its on my todo list in a future pr

Wentao-Kuang (approved):

PR #2681, opened 2023-02-09
feat: add proof of concept date range slider

ajacombs:
Rough proof of concept to add a date range slider UI to the Basemaps landing page.

### DIFF
--- a/packages/attribution/src/attribution.ts
+++ b/packages/attribution/src/attribution.ts
@@ -43,15 +43,10 @@ export class AttributionBounds {
    */
   intersects(extent: BBox, zoom: number, dateAfter?: string, dateBefore?: string): boolean {
     if (zoom > this.maxZoom || zoom < this.minZoom) return false;
+    if (dateAfter && this.endDate && dateAfter > this.endDate) return false;
+    if (dateBefore && this.startDate && dateBefore < this.startDate) return false;
     if (!Wgs84.intersects(extent, this.bbox)) return false;
-    const poly = Wgs84.bboxToMultiPolygon(extent);
-    // Is the endDate later than the supplied dateAfter date.
-    // Default to true so as not to exclude if either is null.
-    const isEndAfter = this.endDate == null || dateAfter == null || this.endDate >= dateAfter;
-    // Is the startDate earlier than the supplied dateBefore date.
-    // Default to true so as not to exclude if either is null.
-    const isStartBefore = this.startDate == null || dateBefore == null || this.startDate <= dateBefore;
-    return isEndAfter && isStartBefore && this.intersection(poly);
+    return this.intersection(Wgs84.bboxToMultiPolygon(extent));
   }
 
   /**
--- a/packages/attribution/src/attribution.ts
+++ b/packages/attribution/src/attribution.ts
@@ -31,20 +31,27 @@ export class AttributionBounds {
   }
 
   /**
-   * Does this AttributionCollection intersect with `extent`
+   * Does this AttributionCollection intersect with an extent at a specified
+   * zoom level, optionally within a specified date range.
    *
    * @param extent The extent to test for intersection against
    * @param zoom only test extent if `zoom` is between `minZoom` and `maxZoom`
+   * @param dateAfter ISO 8601 format datetime for the start of the range of
+   *        time to test whether this occurs within.
+   * @param dateBefore ISO 8601 format datetime for the end of the range of
+   *        time to test whether this occurs within.
    */
   intersects(extent: BBox, zoom: number, dateAfter?: string, dateBefore?: string): boolean {
     if (zoom > this.maxZoom || zoom < this.minZoom) return false;
     if (!Wgs84.intersects(extent, this.bbox)) return false;
     const poly = Wgs84.bboxToMultiPolygon(extent);
-    return (
-      (this.endDate == null || dateAfter == null || this.endDate >= dateAfter) &&
-      (this.startDate == null || dateBefore == null || this.startDate <= dateBefore) &&
-      this.intersection(poly)
-    );
+    // Is the endDate later than the supplied dateAfter date.
+    // Default to true so as not to exclude if either is null.
+    const isEndAfter = this.endDate == null || dateAfter == null || this.endDate >= dateAfter;
+    // Is the startDate earlier than the supplied dateBefore date.
+    // Default to true so as not to exclude if either is null.
+    const isStartBefore = this.startDate == null || dateBefore == null || this.startDate <= dateBefore;
+    return isEndAfter && isStartBefore && this.intersection(poly);
   }
 
   /**
--- a/packages/attribution/src/attribution.ts
+++ b/packages/attribution/src/attribution.ts
@@ -17,6 +17,8 @@ export class AttributionBounds {
   boundaries: Ring[][] = [];
   minZoom: number;
   maxZoom: number;
+  startDate?: string;
+  endDate?: string;
 
   constructor(collection: AttributionCollection) {
     this.collection = collection;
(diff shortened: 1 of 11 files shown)


packages/landing/src/components/daterange.tsx
```diff
@@ -0,0 +1,68 @@
+import { Component, ReactNode } from 'react';
+import { Config } from '../config';
+
+const minYear = '1950';
+const maxYear = '2023';
```
Wentao-Kuang, line 5: This should be able to use `new Date().getFullYear();`

packages/landing/src/components/map.tsx
```diff
@@ -80,14 +81,82 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
   updateStyle = (): void => {
     this.ensureGeoControl();
     const tileGrid = getTileGrid(Config.map.tileMatrix.identifier);
-    const style = tileGrid.getStyle(Config.map.layerId, Config.map.style);
+    const style = tileGrid.getStyle(Config.map.layerId, Config.map.style, undefined, Config.map.dateRange);
     this.map.setStyle(style);
 
     if (Config.map.tileMatrix !== GoogleTms) this.map.setMaxBounds([-180, -85.06, 180, 85.06]);
     else this.map.setMaxBounds();
+    // TODO check and only update when Config.map.layer changes.
```
Wentao-Kuang, line 89: I think we can check layers changes from Attribution now, we don't need Config.map.layer at here anymore.

packages/landing/src/attribution.ts
```diff
@@ -114,8 +118,17 @@ export class MapAttribution {
...
+    const dateAfter = Config.map.dateRange.yearAfter
+      ? `${Config.map.dateRange.yearAfter}-01-01T00:00:00.000Z`
+      : undefined;
+    const dateBefore = Config.map.dateRange.yearAfter
+      ? `${Config.map.dateRange.yearBefore}-12-31T23:59:59.999Z`
+      : undefined;
     const bbox = this.mapboxBoundToBbox(this.bounds, Config.map.tileMatrix);
-    const filtered = attr.filter(bbox, this.zoom);
+    const filtered = attr.filter(bbox, this.zoom, dateAfter, dateBefore);
+    const filteredLayerIds = filtered.map((x) => x.id).join('_');
```
Wentao-Kuang, line 129: Can we hash this instead of a very long string with ids.

packages/landing/src/components/daterange.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+    this.scheduleUpdateConfig();
+  };
+
+  render(): ReactNode {
+    return (
+      <div className="date-range">
+        <p>After: {this.state.yearAfter}</p>
+        <input
+          id="after"
```
blacha, line 47: `id` is globally unique on the page this would limit this, could you classify these sliders in a different way?

ajacombs: `date-range-after` & `date-range-before`?

packages/landing/src/url.ts
```diff
@@ -97,10 +98,13 @@ export const WindowUrl = {
     layerId: string,
     style?: string | null,
     config?: string | null,
+    dateRange?: DateRangeState,
   ): string {
     const queryParams = new URLSearchParams();
     if (Config.ApiKey != null && Config.ApiKey !== '') queryParams.set('api', Config.ApiKey);
     if (config != null) queryParams.set('config', ensureBase58(config));
+    if (dateRange != null && dateRange.yearAfter != null) queryParams.set('date[after]', dateRange.yearAfter);
```
blacha, line 106: NIT: could use `? eg:  `if (dateRange?.yearAfter != null) ...`

packages/landing/static/index.css
```diff
@@ -196,4 +196,19 @@ button {
...
   width: 600px;
   height: 40px;
+}
+
+.date-range {
+  background:rgba(255, 255, 255, 0.8);
+  box-shadow: 0 0 0.5rem 0.5rem rgb(0 0 0 / 10%);
+  border: 2px solid rgba(255, 255, 255);
+  border-radius: 0.5rem;
+  padding: 0 1rem 1rem 1rem;
```
blacha, line 206: I dont think we use any other `rem` for measurements, might just be best to use what everything else is using 4/8px increment.s

packages/landing/src/components/daterange.tsx
```diff
@@ -0,0 +1,68 @@
+import { Component, ReactNode } from 'react';
+import { Config } from '../config';
+
+const minYear = '1950';
+const maxYear = '2023';
+
+export interface DateRangeState {
+  yearAfter?: string;
```
blacha, line 8: can these be `dateBefore` / `dateAfter` to be consistent with everything else, you can just set them as full ISO strings.

packages/landing/src/components/layout.header.tsx
```diff
@@ -178,12 +179,26 @@ Your Service/App URL:
 
     const children: ReactNode[] = [];
     if (projections.has(EpsgCode.Nztm2000)) {
-      const nztmTileUrl = Config.map.toTileUrl(MapOptionType.Wmts, Nztm2000QuadTms);
+      const nztmTileUrl = Config.map.toTileUrl(
```
blacha, line 182: NIT: I think this needs to change into some other way of constructing a tileUrl.

Using a object I think would make this a lot more understandable.
`Config.map.toTileUrl({ type: Wmts, tileMatrix: GoogleTms })`

packages/landing/src/attribution.ts
ajacombs: I did wonder about that, but wasn't sure how to create a hash without bringing in a dependency. What would be the cleanest way to do that?

Wentao-Kuang: I normally use crypto in basemaps.

``import { createHash } from 'crypto'; 
const cacheKey = createHash('sha256').update(data).digest('base64');``

blacha: this runs both in the browser and in node so `import crypto` is not a option, using a map join is a simple work around.

packages/landing/src/components/daterange.tsx
blacha: still would be globally unique, I guess the easiest option would be to `onChange={e => this.handleChange(e, 'after')`

packages/attribution/src/attribution.ts
```diff
@@ -32,10 +36,15 @@ export class AttributionBounds {
    * @param extent The extent to test for intersection against
    * @param zoom only test extent if `zoom` is between `minZoom` and `maxZoom`
    */
-  intersects(extent: BBox, zoom: number): boolean {
+  intersects(extent: BBox, zoom: number, dateAfter?: string, dateBefore?: string): boolean {
     if (zoom > this.maxZoom || zoom < this.minZoom) return false;
     if (!Wgs84.intersects(extent, this.bbox)) return false;
-    return this.intersection(Wgs84.bboxToMultiPolygon(extent));
+    const poly = Wgs84.bboxToMultiPolygon(extent);
+    return (
```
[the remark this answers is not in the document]
ajacombs: What do you think of latest commit which separates them out (and adds some comments to clarify logic)?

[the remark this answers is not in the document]
blacha: there are a bunch of previous checks that use `if (..) return false` it would be good to follow that pattern

packages/landing/src/components/daterange.tsx
ajacombs: Ah yup. Have pushed new commit to change to that and use an enum for the two kinds.

blacha (approved):

PR #2682, opened 2023-02-09
fix(server): make --no-config actually load the configuration from tiffs

blacha:
This config option was documented but didn't exist in the code

### DIFF
--- a/packages/config/src/json/tiff.config.ts
+++ b/packages/config/src/json/tiff.config.ts
@@ -78,7 +78,7 @@ async function loadStacFromPath(target: string): Promise<StacCollection | null> 
  *
  * @returns Imagery configuration generated from the path
  */
-export async function imageryFromTiffPath(target: string, Q: pLimit.Limit): Promise<ConfigImagery> {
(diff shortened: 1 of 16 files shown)


packages/config/src/json/tiff.config.ts
```diff
@@ -78,7 +78,7 @@ async function loadStacFromPath(target: string): Promise<StacCollection | null>
  *
  * @returns Imagery configuration generated from the path
  */
-export async function imageryFromTiffPath(target: string, Q: pLimit.Limit): Promise<ConfigImagery> {
+export async function imageryFromTiffPath(target: string | string, Q: pLimit.Limit): Promise<ConfigImagery> {
```
Wentao-Kuang, line 81: What is `string | string` DateType? I never see this before.

blacha: I wonder how that got there! `string | string` is just a `string`!

Wentao-Kuang (dismissed):

Wentao-Kuang (approved):

PR #2691, opened 2023-02-15
feat(landing): update and load date range from window URL

Wentao-Kuang:


packages/landing/src/components/daterange.tsx
```diff
@@ -35,6 +36,8 @@ export class DateRange extends Component {
     Config.map.dateRange.dateAfter = this.state.dateAfter;
     Config.map.dateRange.dateBefore = this.state.dateBefore;
     Config.map.emit('dateRange', this.state);
+    const dateRangeSearch = '?' + MapConfig.toUrl(Config.map);
+    window.history.replaceState(null, '', dateRangeSearch);
```
blacha, line 40: we shouldnt change the window object in here, we can do it from the dateRange event?

packages/landing/src/components/daterange.tsx
```diff
@@ -49,6 +52,12 @@ export class DateRange extends Component {
     this.scheduleUpdateConfig();
   };
 
+  componentDidMount(): void {
+    Config.map.updateFromUrl();
```
blacha, line 56: we should listen for updates so if people update the url this should update.

Wentao-Kuang: What event do you think is good for tracking url updates.

packages/landing/src/config.map.ts
```diff
@@ -107,9 +109,12 @@ export class MapConfig extends Emitter<MapConfigEvents> {
     this.style = style ?? null;
     this.layerId = layerId.startsWith('im_') ? layerId.slice(3) : layerId;
     this.tileMatrix = tileMatrix;
+    this.dateRange.dateBefore = dateBefore ?? undefined;
+    this.dateRange.dateAfter = dateAfter ?? undefined;
```
blacha, line 113: do we want to validate that these dates are valid? `new Date().getTime()` and dateBefore < dateAfter?

Wentao-Kuang: Good point

packages/landing/src/components/daterange.tsx
blacha: I thought we had something like a on PopState or something somewhere?

packages/landing/src/components/daterange.tsx
```diff
@@ -49,6 +50,20 @@ export class DateRange extends Component {
     this.scheduleUpdateConfig();
   };
 
+  componentDidMount(): void {
+    window.addEventListener('popstate', function () {
```
blacha, line 54: This should not be done inside of the this class, you will get into lots of weird issues if you start b inding window events between multiple components. there is nothing stopping this component being added to the page multiple times.

The window URL / Config logic should be handling all of this.

If you ever bind events inside of a didMount, you need to make sure you unbind them too.

Wentao-Kuang: I think this updateFromUrl already been added as event for the whole front-end. We might not need to add another listener for this. 
image

packages/landing/src/components/daterange.tsx
```diff
@@ -1,8 +1,9 @@
 import { Component, ReactNode } from 'react';
-import { Config } from '../config';
+import { Config } from '../config.js';
+import { MapConfig } from '../config.map.js';
 
-const minDate = '1950-01-01T00:00:00.000Z';
-const maxDate = `${new Date().getFullYear().toString()}-12-31T23:59:59.999Z`;
+export const minDate = '1950-01-01T00:00:00.000Z';
```
blacha, line 5: if exported,  use `TitleCase`

blacha (approved):

PR #2699, opened 2023-02-20
feat(cli): Update a preview url for zoom level changes.

Wentao-Kuang:


packages/cli/src/cli/config/action.import.ts
```diff
@@ -176,7 +176,7 @@ export class CommandImport extends CommandLineAction {
           let msg = ' - Zoom level updated.';
           if (layer.minZoom !== existing.minZoom) msg += ` min zoom ${existing.minZoom} -> ${layer.minZoom}`;
           if (layer.maxZoom !== existing.maxZoom) msg += ` max zoom ${existing.maxZoom} -> ${layer.maxZoom}`;
-          change.push(`${msg}\n`);
+          change.push(`${msg} -- [Aerial]((${PublicUrlBase}?config=${this.config.value}&debug))\n`);
```
blacha, line 179: I think this url should be added to the next if statements? we are going to need both NZTM and Webmercator links?

What happens if the zoom changes and the layer id changes? going to get a bunch of links?

Wentao-Kuang: Good point, we shouldn't end up with too many links. I have update this only have a default aerial preview link if only zoom level change and layer id not change.

Wentao-Kuang: I add the links for both available projections now.

image

packages/cli/src/cli/config/action.import.ts
```diff
@@ -171,21 +171,28 @@ export class CommandImport extends CommandLineAction {
...
+        if (layer[2193]) {
           const urls = await this.prepareUrl(layer[2193], mem, Nztm2000QuadTms);
-          change.push(`- Layer update [NZTM2000Quad](${urls.layer}) -- [Aerial](${urls.tag})\n`);
+          if (layer[2193] !== existing[2193])
+            change.push(`- Layer update [NZTM2000Quad](${urls.layer}) -- [Aerial](${urls.tag})\n`);
+          if (zoom) zoom += ` [NZTM2000Quad](${urls.tag})`;
         }
-        if (layer[3857] && layer[3857] !== existing[3857]) {
+        if (layer[3857]) {
           const urls = await this.prepareUrl(layer[3857], mem, GoogleTms);
```
blacha, line 189: need brackets here.

blacha (approved):

PR #2701, opened 2023-02-20
fix: ensure layer loads when dateBefore set early, then moved back late

ajacombs:
When the date before slider was moved to a position earlier than any available imagery (causing the map to be empty), and then moved to a position just after an available imagery layer, the attribution text would update to show the change, but the imagery would not load on the map. On moving the date before slider later again, beyond when a second imagery set was available, that second imagery layer would be displayed.

This was caused by `if (!Config.map.visibleLayers)` being true for both `undefined` (its value before the sliders have been moved), and the empty string, its value when the sliders are set to a position where there is no imagery available.

Related to this bug, there seemed to be some 'glitchiness' around when imagery layers are loaded. This may be due to transient race conditions between the debouncing added in `scheduleUpdateConfig` competing with equivalent debouncing for the attribution rendering. This was added to minimise network requests as previously changing the date sliders requested a new attribution json file. With this refactored out to client side filtering, the debouncing is no longer needed, as the loading of new layers is triggered inside the attribution update functions which are already debounced.


packages/landing/src/components/daterange.tsx
```diff
@@ -46,7 +28,9 @@ export class DateRange extends Component {
         this.setState({ dateBefore: `${event.target.value}-12-31T23:59:59.999Z` });
         break;
     }
-    this.scheduleUpdateConfig();
+    Config.map.dateRange.dateAfter = this.state.dateAfter;
```
blacha, line 31: would be good to check if these have actually changed.

ajacombs: `handleChange` is the callback attached to the `onChange` attribute of the two `<input>` elements, so will only ever be called when a value has changed.

packages/landing/src/components/daterange.tsx
```diff
@@ -45,7 +29,9 @@ export class DateRange extends Component {
         this.setState({ before: `${event.target.value}-12-31T23:59:59.999Z` });
         break;
     }
-    this.scheduleUpdateConfig();
+    Config.map.dateRange.dateAfter = this.state.dateAfter;
```
Wentao-Kuang, line 32: This should be `Config.map.filter.date.after` now.

packages/landing/src/components/daterange.tsx
ajacombs: Further to this, changes made by @Wentao-Kuang to add `Config.map.setFilterDateRange` (which updates the date range values on the config and emits `filter` and `change` events if they have changed) mean that we don't need to emit anything here, and the events will only be emitted if needed.

packages/landing/src/components/map.tsx
```diff
@@ -91,7 +91,7 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
   };
 
   updateVisibleLayers = (newLayers: string): void => {
-    if (!Config.map.visibleLayers) Config.map.visibleLayers = newLayers;
+    if (Config.map.visibleLayers === undefined) Config.map.visibleLayers = newLayers;
```
blacha, line 94: == null is preferred, unless you really need to validate that this is undefined and not null

ajacombs: All good, changed in new commit

packages/landing/src/components/map.tsx
```diff
@@ -91,7 +91,7 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
   };
 
   updateVisibleLayers = (newLayers: string): void => {
-    if (!Config.map.visibleLayers) Config.map.visibleLayers = newLayers;
+    if (Config.map.visibleLayers == null) Config.map.visibleLayers = newLayers;
```
Wentao-Kuang, line 94: I think `visibleLayers` is defined as `string` and not initialised. Might be better to defined as `string|null`. Then you can do this check for initialising here.

blacha (approved):

blacha (approved):

PR #2702, opened 2023-02-20
feat: use transition property for fades, add fade when removing layer

ajacombs:
- Changes the layer fade animation to use the built-in MapLibre transition `raster-opacity-transition` when fading layers rather than repeatedly updating the `raster-opacity` ourselves in JavaScript as before.
- Adds an fade transition when previous layers are being removed, which makes the appearance less jarring.
- Reduces the fade time to 750ms from 1000ms.

The order of operations is now:
- The date range sliders are moved.
- We check if the layers which would be visible in the map bounding box within the date range specified have changed.
- If they have, we add a new layer with the source set to the new date after and date before values.
- The new layer fades in over 750ms.
- If the tiles for the new layer take longer than 750ms to load, they continue to load.
- Once the new layer has finished loading and no further interaction is occurring, the `idle` event is triggered.
- On the idle event, the previous layers opacity is reduced to 0 over 750ms.
- After 750ms, the previous layer and source are removed.

An `as any` typecast is required due to the `*-transition` paint properties not being included in the MapLibre TypeScript typing, which is a bug being tracked at https://github.com/maplibre/maplibre-gl-js/issues/1708.

### DIFF
--- a/packages/landing/src/components/map.tsx
+++ b/packages/landing/src/components/map.tsx
@@ -9,7 +9,7 @@ import { DateRange } from './daterange.js';
 import { Debug } from './debug.js';
 import { MapSwitcher } from './map.switcher.js';
 
-const layerFadeTime = 750;
+const LayerFadeTime = 750;
 
 /**
  * Map loading in maplibre is weird, the on('load') event is different to 'loaded'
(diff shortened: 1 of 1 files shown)


packages/landing/src/components/map.tsx
```diff
@@ -9,6 +9,8 @@ import { DateRange } from './daterange.js';
 import { Debug } from './debug.js';
 import { MapSwitcher } from './map.switcher.js';
 
+const layerFadeTime = 750;
```
blacha, line 12: `TitleCase`

blacha: Some examples of variable names 

```typescript
export const AwsRoleDurationHours = 7
export const CacheExpiresMs = 7500

export const OneDayMs = 60 * 60 * 24 * 1000;
export const AwsRoleDays = OneDayMs * 20;
```

packages/landing/src/components/map.tsx
```diff
@@ -117,32 +119,18 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
...
-          paint: { 'raster-opacity': 0 },
-        });
+          paint: {
+            'raster-opacity': 0,
+            'raster-opacity-transition': {
+              duration: layerFadeTime,
+              delay: 0,
+            },
+          },
+        } as any);
```
blacha, line 129: would be great to remove this `any` 

can we use  `this.map.setPaintProperty(..., 'raster-opacity-transition')`

Wentao-Kuang (approved):

PR #2706, opened 2023-02-23
fix(landing): Stop set state in the render which lead to infinite loop

Wentao-Kuang:


packages/landing/src/attribution.ts
```diff
@@ -119,7 +119,12 @@ export class MapAttribution {
     this.zoom += 1;
 
     const bbox = this.mapboxBoundToBbox(this.bounds, Config.map.tileMatrix);
-    const filtered = attr.filter(bbox, this.zoom, Config.map.dateRange.dateAfter, Config.map.dateRange.dateBefore);
+    const filtered = attr.filter(
+      bbox,
+      this.zoom,
+      Config.map.filter.dateRange.dateAfter,
```
blacha, line 125: could simplify to `date.before` `date.after` Im not sure you need to duplicate the `date` in both parts.

Wentao-Kuang: Good point, just fixed. Thank you.

packages/landing/src/attribution.ts
```diff
@@ -119,7 +119,7 @@ export class MapAttribution {
     this.zoom += 1;
 
     const bbox = this.mapboxBoundToBbox(this.bounds, Config.map.tileMatrix);
-    const filtered = attr.filter(bbox, this.zoom, Config.map.dateRange.dateAfter, Config.map.dateRange.dateBefore);
+    const filtered = attr.filter(bbox, this.zoom, Config.map.filter.date.after, Config.map.filter.date.before);
```
blacha, line 122: NIT: this function is starting to get too big, I think we should move it to a object too, it would be very easy to accidentally put `zoom, before, after` rather than `zoom, after, before` 

But can be done seperately

Wentao-Kuang: Not a big change. I will do this now.

packages/landing/src/components/debug.tsx
```diff
@@ -158,24 +157,22 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
...
-  setWMTSLink(): void {
-    const wmtsLink = WindowUrl.toTileUrl({
+  getWMTSLink(): string {
+    return WindowUrl.toTileUrl({
       urlType: MapOptionType.Wmts,
       tileMatrix: Config.map.tileMatrix,
       layerId: Config.map.layerId,
       config: Config.map.config,
-      dateRange: Config.map.dateRange,
+      date: Config.map.filter.date,
```
blacha, line 166: This would be better taking `filter: Config.map.filter` 

although looking at this, I wonder i f we should have `WindowUrl.toConfigTileUrl(Wmts)` or similar

Wentao-Kuang: That will be much better, which just take a `urlType` and we have all the configs to use directly.

packages/landing/src/attribution.ts
Wentao-Kuang: I have make the change in #2709

packages/landing/src/components/debug.tsx
Wentao-Kuang: I have made the change in #2710

packages/landing/src/config.map.ts
```diff
@@ -162,6 +166,13 @@ export class MapConfig extends Emitter<MapConfigEvents> {
     this.emit('change');
   }
 
+  setFilterDateRange(after: string | undefined, before: string | undefined): void {
+    this.filter.date.after = after;
```
blacha, line 170: this needs to check to see if it actually changed, before actually firing the events.

blacha (dismissed):

blacha (approved):

PR #2736, opened 2023-03-15
feat(sprites): support non svg sprites

blacha:
adds support for `--extension .png` to load png (or any other format that libvips supports) as sprites


ajacombs (approved):

jleedev:
This is a breaking API change but was published with a minor version number.

blacha:
@jleedev you are correct, this was not taged as a breaking change (missing a `!` in the commit title.

We have a couple options, depending on the impact for you:

1. We do nothing,
2. We roll back this change, release a new patch version, then create a major version bump for the api change?

jleedev:
Based on github code search, openstreetmap-americana might be the only code out there that is actually importing this api. so doing nothing is probably fine.

PR #2744, opened 2023-03-19
fix(cli): multithread gdal warp in COG creation

blacha:
https://github.com/OSGeo/gdal/issues/7478


packages/cli/src/gdal/gdal.cog.ts
```diff
@@ -92,6 +92,10 @@ export class GdalCogBuilder {
       // Max CPU POWER
       '-co',
       'NUM_THREADS=ALL_CPUS',
+      // in GDAL 3.7.x NUM_THREADS will also set GDAL_NUM_THREADS
```
Wentao-Kuang, line 95: This looks good to me. Do we also need to upgrade the GDAL version?

Wentao-Kuang (approved):

PR #2746, opened 2023-03-20
feat(lambda-tiler): Include the disabled layers in the attribution with minZoom of 32.

Wentao-Kuang:


packages/geo/src/stac/stac.attribution.ts
```diff
@@ -7,6 +7,7 @@ export type AttributionCollection = StacCollection<{
   'linz:category'?: string;
   'linz:zoom': { min: number; max: number };
   'linz:priority': [number];
+  'linz:disabled'?: boolean;
```
blacha, line 10: this is now always set. no need for `?`

blacha (dismissed):

blacha (approved):

PR #2749, opened 2023-03-20
fix(sprites): correct imports for node12

blacha:


packages/sprites/src/fs.ts
```diff
@@ -1,11 +1,11 @@
-import { readdir, readFile } from 'node:fs/promises';
+import { promises as fs } from 'node:fs';
```
Wentao-Kuang, line 1: Just wondering what is the difference between `import { promises as fs } from 'node:fs';` and `import { promises as fs } from 'fs';` ?

Wentao-Kuang (approved):

PR #2751, opened 2023-03-20
wip: basic chart and stats from imagery

blacha:


blacha:
We need to revist the way we handle historical layers, so this is not valid anymore.

PR #2759, opened 2023-04-02
feat(cogify): retile imagery into COGS aligned to a tile matrix

blacha:


packages/cogify/src/cogify/cli/cli.cog.ts
```diff
@@ -0,0 +1,266 @@
...
+}
+
+function isPathLocal(path: string): boolean {
+  return path.startsWith('./') || path.startsWith('/');
+}
+
+export const BasemapsCogifyCreateCommand = command({
+  name: 'cogify-create',
+  version: CliInfo.version,
+  description: 'Create a covering configuration from a collection from source imagery',
```
Wentao-Kuang, line 49: Description need to update.

packages/cogify/src/cogify/cli/cli.cover.ts
```diff
@@ -0,0 +1,83 @@
...
+import { CutlineOptimizer } from '../../cutline.js';
+import { getLogger, logArguments } from '../../log.js';
+import { createTileCover, TileCoverContext } from '../../tile.cover.js';
+
+const SupportedTileMatrix = [GoogleTms, Nztm2000QuadTms];
+
+export const BasemapsCogifyCoverCommand = command({
+  name: 'cogify-cover',
+  version: CliInfo.version,
+  description: 'Create a COG from a covering configuration',
```
Wentao-Kuang, line 17: I think this description is for cogify-create ?

packages/cogify/src/cogify/covering.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      if (areaPercent < minCoveragePercent && zDiff < maxZoomDifference) {
+        // Not enough coverage was found with this tile, use a more zoomed in tile and try again
+        todo.push(...QuadKey.children(tileQk).map((qk) => QuadKey.toTile(qk)));
+      } else {
+        zoomDifference = Math.max(zDiff, zoomDifference);
+        targetTiles.add(tileQk);
+      }
+
+      if (zDiff === 0) addSurrounding(visited, todo, currentTile, ctx.targetZoom);
```
Wentao-Kuang, line 82: Why do we need to add surrounding for targetZoom?

packages/cogify/src/tile.cover.ts
```diff
@@ -0,0 +1,201 @@
...
+export async function createTileCover(ctx: TileCoverContext): Promise<TileCoverResult> {
+  // Find the zoom level that is at least as good as the source imagery
+  const targetBaseZoom = Projection.getTiffResZoom(ctx.tileMatrix, ctx.imagery.gsd);
+  const targetPixelScale = ctx.tileMatrix.pixelScale(targetBaseZoom);
+
+  // The base zoom is 256x256 pixels at its resolution, we are trying to find a image that is <32k pixels wide/high
+  // zooming out 7 levels converts a 256x256 image into 32k x 32k image
+  // 256 * 2 ** 7 = 32,768 - 256x256 tile
+  // 512 * 2 ** 6 = 32,768 - 512x512 tile
+  const optimalCoveringZoom = targetBaseZoom - 7; // z12 from z19
```
Wentao-Kuang, line 44: Is this possible become less then 0? Like satellite imagery with a small `targetBaseZoom`. Do we want to constraint this number to be larger than 10 or some other number that not too large?

blacha: Yes we do, should limit to `Math.min(1, ...)` good :eyes:

packages/cogify/src/cogify/cli/cli.cover.ts
blacha: You are correct, this is the wrong description, ill fix it

packages/cogify/src/cogify/covering.ts
blacha: since we traverse down into sub tiles this line could be called with a sub tile, we dont want to check sub tiles of nearby tiles as their parents could already be in the todo queue.

Wentao-Kuang (approved):

PR #2764, opened 2023-04-05
feat(landing): Update the daterange slider to years button.

Wentao-Kuang:


packages/landing/src/components/debug.tsx
```diff
@@ -65,7 +65,7 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
       if (Config.map.debug['debug.screenshot']) {
         map.once('idle', async () => {
           // Ensure all the attribution data has loaded
-          await Promise.all([...Attributions.values()]);
+          await Promise.all([...MapAttrState._attrs.values()]); // Unsure if needed
```
blacha, line 68: why are you unsure if this is needed? we could move this to `getCurrentAttribution` as we need to wait for the attribution to load.

Wentao-Kuang: I think that comment is from Andrew. I have updated that to `await MapAttrState.getCurrentAttribution();` to make sure the current attribution is resolved.

blacha (approved):

PR #2771, opened 2023-04-26
feat(cogify): Update the cmd in the cogify README.

Wentao-Kuang:


blacha:
The docs are correct and  is the desired behavior, I have made #2785 to fix the code to be consistent with the docs.

PR #2784, opened 2023-05-10
fix(cli): Fix remove the ref from the repository and get the repo name from it.

Wentao-Kuang:


packages/cli/src/cli/github/github.ts
```diff
@@ -4,10 +4,12 @@ import { execFileSync } from 'child_process';
 export class Github {
   repo: string;
   logger: LogType;
+  repoName: string;
 
   constructor(repo: string, logger: LogType) {
     this.repo = repo;
     this.logger = logger;
+    this.repoName = repo.split('/').slice(-1)[0];
```
blacha, line 12: This is a interesting way to do this.

How about something like

```typescript
const [org, repoName] = repo.split('/')
if (org == null || repoName == null) throw 'Badly formatted repo name'
```

blacha (approved):

PR #2785, opened 2023-05-10
fix(cogify): cogify create should create from stac item json

blacha:


packages/cogify/src/cogify/cli/cli.cog.ts
```diff
@@ -49,150 +64,108 @@ export const BasemapsCogifyCreateCommand = command({
...
-          await fsa.write(targetFile, fsa.stream(f));
-          const duration = performance.now() - startTime;
-          logger.debug({ source: f, target: targetFile, duration }, 'Cog:Source:Download:Done');
-          sourceFileMap.set(f, targetFile);
-        }),
-      );
-
-      for (const item of toCreate) {
+      // TODO should COG creation be run concurrently?
+      for (const { itemPath, item } of filtered) {
```
Wentao-Kuang, line 111: I think we already allow to running concurrent gdal commands previously.

blacha: It is pretty easy to configure into concurrent execution, but unsure if its worth the extra complexity in this pull request.

I could look to add it in after some initial performance tests in Argo?

Wentao-Kuang (approved):

PR #2794, opened 2023-05-20
feat(config): Create an all tileset from imagery configs. BM-805

Wentao-Kuang:


packages/config/src/memory/memory.config.ts
```diff
@@ -105,6 +105,7 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
 
   /** Find all imagery inside this configuration and create a virtual tile set for it */
   createVirtualTileSets(): void {
+    const allLayers: ConfigLayer[] = [];
```
blacha, line 108: We need to in the future handle the removal of layers from the `aerial` tileset, so when we do that all children virutal tilesets will dissapear.

we need to retain `aerial:antipodes-islands-satellite-2019-2020-0.5m` even if it doesnt exist inside of `aerial:` 

Also how often is `aerial:...` actually used? we should get some metrics.

packages/config/src/memory/memory.config.ts
```diff
@@ -115,9 +116,34 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
...
     }
+    // Create an all tileset contains all raster layers
+    if (allLayers.length) this.createVirtualAllTileSet(allLayers);
+  }
+
+  createVirtualAllTileSet(layers: ConfigLayer[]): void {
+    const layerByName = new Map<string, ConfigLayer>();
+    // Set all layers as minZoom:32
+    for (const l of layers) {
+      delete l.maxZoom;
```
blacha, line 131: cannot modify an existing layer as you break all other layers in memory.

packages/config/src/memory/memory.config.ts
```diff
@@ -115,9 +116,34 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
...
+    if (allLayers.length) this.createVirtualAllTileSet(allLayers);
+  }
+
+  createVirtualAllTileSet(layers: ConfigLayer[]): void {
+    const layerByName = new Map<string, ConfigLayer>();
+    // Set all layers as minZoom:32
+    for (const l of layers) {
+      const newLayer = { ...l, maxZoom: undefined, minZoom: 32 };
+      // TODO: This might overwrite the layer id for duplicated configImagery
+      if (layerByName.has(newLayer.name))
```
blacha, line 133: It would be good to compare timestamps to see which imagery set is newer when overwriting it.

blacha: also use `{` anytime you need a if over two lines.

packages/config/src/memory/memory.config.ts
blacha: Turns out it was basically just OSM using the `aerial:` prefix, we have migrated OSM to the newer direct imagery names so now there is pretty much 0 access to `aerial:` and it is safe to remove.

packages/config/src/memory/memory.config.ts
```diff
@@ -113,11 +114,47 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
...
+    const ulidA = ConfigId.unprefix(ConfigPrefix.Imagery, idA);
+    const ulidB = ConfigId.unprefix(ConfigPrefix.Imagery, idB);
+    try {
+      const timeA = decodeTime(ulidA);
+      const timeB = decodeTime(ulidB);
+      if (timeA >= timeB) return idA;
+      else return idB;
+    } finally {
+      //If not ulid return the idB
+      return idB;
```
blacha, line 137: would be good to localeCompare idA vs idB so then its still alphabetical.

so then `findLatestId("a", "b")` is the same output as `findLatestId("b", "a")`

This function does not use `this` so maybe just export it as a standalone function.

packages/config/src/memory/memory.config.ts
```diff
@@ -140,8 +177,14 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
       removeUndefined(existing);
       this.put(existing);
     }
-    // TODO this overwrites existing layers
-    existing.layers[0][i.projection] = i.id;
+    // The latest imagery overwrite the earlier ones.
+    const existingImageryId = existing.layers[0][i.projection];
+    if (existingImageryId) {
+      existing.layers[0][i.projection] = this.findLatestId(existingImageryId, i.id);
```
blacha, line 183: do we have a logger and can we log that this is being overwritten?

If its not loggable maybe we could store these as `this.duplicateImagery = [];`

packages/config/src/memory/memory.config.ts
```diff
@@ -176,11 +181,13 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
       } as ConfigTileSet;
       removeUndefined(existing);
       this.put(existing);
+    } else {
+      this.duplicateImagery.push(existing);
```
blacha, line 184: I think this is in the wrong spot, should it not be on line 190?

Wentao-Kuang: Do you just want to catch the ones that have id replaced?
I thought we want to catch all the ones have duplicated imagery names in configs with different ids.

blacha: The tileset will always be duplicated as this function will be called, for 2193 imagery AND 3857 imagery

so this check needs to move down into the projection logic.

packages/config/src/memory/memory.config.ts
```diff
@@ -31,6 +32,20 @@ function isConfigTileSet(i: BaseConfig): i is ConfigTileSet {
...
+function findLatestId(idA: string, idB: string): string {
+  const ulidA = ConfigId.unprefix(ConfigPrefix.Imagery, idA);
+  const ulidB = ConfigId.unprefix(ConfigPrefix.Imagery, idB);
+  try {
+    const timeA = decodeTime(ulidA);
+    const timeB = decodeTime(ulidB);
+    if (timeA >= timeB) return idA;
+  } finally {
+    //If not ulid return the return id alphabetically.
+    return idA.localeCompare(idB) ? idA : idB;
```
blacha, line 45: I wonder if it would be cleaner to write `idA.localeCompare(idB) > 0 ? idA : idB`

packages/config/src/memory/__tests__/memory.config.test.ts
```diff
@@ -94,28 +104,36 @@ o.spec('MemoryConfig', () => {
...
     o(target?.layers.length).equals(1);
-    o(target?.layers[0][3857]).equals('im_Image234');
+    o(target?.layers[0][3857]).equals(newImId);
     o(target?.layers[0][2193]).equals(undefined);
     o(target?.name).equals('ōtorohanga-urban-2021-0.1m');
   });
 
   o('virtual tilesets should be created with `:`', async () => {
+    const idA = ulid();
+    const idB = ulid();
```
blacha, line 136: need to wait between ulid generations
```await timers.setTimeout(5)```

File: packages/config/src/memory/__tests__/memory.config.test.ts [hunk 9]
```diff
@@ -176,7 +175,9 @@ o.spec('MemoryConfig', () => {
   });
 
   o('The latest imagery should overwrite the old ones', async () => {
+    await new Promise((resolve) => setTimeout(resolve, 5));
```
blacha, line 178: you shouldn't need this await.

packages/config/src/memory/__tests__/memory.config.test.ts
```diff
@@ -176,7 +175,9 @@ o.spec('MemoryConfig', () => {
...
     const idLater = ulid();
+    await new Promise((resolve) => setTimeout(resolve, 5));
```
blacha, line 180: There are async implementations of `setTimeout` in node https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options

```typescript 
import timers from 'node:timers/promises'
await timers.setTimeout(5)
```

blacha (approved):

packages/config/src/memory/memory.config.ts
```diff
@@ -140,8 +181,15 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
       removeUndefined(existing);
       this.put(existing);
     }
-    // TODO this overwrites existing layers
-    existing.layers[0][i.projection] = i.id;
+    // The latest imagery overwrite the earlier ones.
+    const existingImageryId = existing.layers[0][i.projection];
+    if (existingImageryId) {
+      existing.layers[0][i.projection] = findLatestId(i.id, existingImageryId);
+      this.duplicateImagery.push(existing);
```
blacha, line 188: NIT: this wont actually give you useful information, it will overwrite the old ID so you only get that there is a duplicate not what the IDs actually are.

Wentao-Kuang: Do you think which id should be useful to catch. Just the old ids that been overwrited?
I will merge this first and fix in https://github.com/linz/basemaps/pull/2812.

PR #2800, opened 2023-05-24
feat(cogify): improve cogify

blacha:


packages/cogify/src/cogify/parsers.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Parse a input parameter as a URL,
+ * if it looks like a file path convert it using `pathToFileURL`
+ **/
+export const Url: Type<string, URL> = {
+  async from(str) {
+    try {
+      return new URL(str);
+    } catch (e) {
+      // Possibly already a URL
+      if (str.includes(':')) throw e;
```
blacha, line 14: remove this, just use `pathToFileURL`

ajacombs (approved):

PR #2809, opened 2023-06-01
feat(landing): Update the layer selector to use all imagery tileset. BM-807

Wentao-Kuang:


blacha (approved):
:LGTM: but needs to wait for #2794

PR #2812, opened 2023-06-02
feat(config): Remove the chirld aerial: tileset from configs. BM-825

Wentao-Kuang:


packages/config/src/memory/memory.config.ts
```diff
@@ -77,7 +74,7 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
   assets: string;
 
   /** Catch configs with the same imagery that using the different imagery ids. */
-  duplicateImagery: ConfigTileSet[] = [];
+  duplicateImagery: string[] = [];
```
blacha, line 77: I think we need more than just the ids how about `{ name: string, id: string, projection: number}[]` ?

Wentao-Kuang: Looks good to me.

Wentao-Kuang: Updated.

blacha (approved):

PR #2828, opened 2023-06-22
fix: truncate the api key when logging

blacha:
Prevents some logs having a full valid api key logged to them while preserving some ability to lookup the apikey to figure our who the request was triggered by.


packages/config/src/api.truncate.ts
```diff
@@ -0,0 +1,7 @@
+/** Truncate a API key to the last 6 digits */
+export function truncateApiKey(x: unknown): string | undefined {
+  if (typeof x !== 'string') return 'invalid';
+  if (x.startsWith('c')) return 'c' + x.slice(x.length - 6);
+  if (x.startsWith('d')) return 'd' + x.slice(x.length - 6);
+  return 'invalid';
```
ajacombs, line 6: Given we have `isValidApiKey` which does validation, do we want to redo the validation to check it starts with c or d here? That would then embed the business logic for what a valid API key (must start with c or d) in a function which doesn't outwardly seem like it would be responsible for that. I.e. could this be simplified to:
```typescript
export function truncateApiKey(x: string): string {
  return x[0] + x.slice(x.length - 6);
}
```
Or, if we do want to retain validation, could it be:
```typescript
export function truncateApiKey(x: unknown): string | undefined {
  if (typeof x !== 'string') return 'invalid';
  if (x.startsWith('c') || x.startsWith('d')) return x[0] + x.slice(x.length - 6);
  return 'invalid';
}
```

blacha: I threw this logic here, primarily because I had this logic inside the analytics aggregation, so to de-dupe the same logic from multiple places.

I am very much on the fence about putting the logic here, im not sure if we need api logic inside of the config?

ajacombs (approved):

PR #2843, opened 2023-07-12
fix(cogify): use a URL in the cogify STAC documents

blacha:


packages/cogify/src/tile.cover.ts
```diff
@@ -158,7 +158,7 @@ export async function createTileCover(ctx: TileCoverContext): Promise<TileCoverR
     // Add the source imagery as a STAC Link
     for (const src of source) {
       const srcLink: CogifyLinkSource = {
-        href: src.name,
+        href: new URL(src.name, ctx.imagery.url).href,
```
Wentao-Kuang, line 161: What is the `ctx.imagery.url` looks like? Is it something like base path for the imagery? like `s3://linz-imager/foo/bar` or `file:imagery/foo/bar/` ?

blacha: it has to be a full URL so no more `/foo/bar` its `file://`

In the future we should remove `uri` and make everything be actual URLs which makes handling filesystem vs s3 vs http a lot easier

Wentao-Kuang (approved):

PR #2876, opened 2023-08-07
ci: run smoke tests on prod after deploy

blacha:


.github/workflows/push.yml
```diff
@@ -126,3 +126,8 @@ jobs:
         env:
           NODE_ENV: 'production'
           GOOGLE_ANALYTICS: ${{secrets.GOOGLE_ANALYTICS_PROD}}
+
+      - name: (Prod) After Deploy Smoke Test
+        run: |
+            cd packages/smoke
+            BASEMAPS_HOST=https://basemaps.linz.govt.nz node --test
```
Wentao-Kuang, line 133: Can we use the `environment.url` for the host?

blacha: unsure how to reference it, and then its a lookup so adds a level of indirection for anyone reading the pipeline file.

Wentao-Kuang (approved):

PR #2880, opened 2023-08-08
ci: move to latest linz/action-typescript

blacha:


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
Wentao-Kuang, line 1: What is this used for?

blacha: This is to prevent windows changing `\n` into `\r\n` its why [a4999e6](https://github.com/linz/basemaps/pull/2880/commits/a4999e65e8d19e10ca62fd3d345f2dfdc9f4889e) failed to build.

Wentao-Kuang (approved):

PR #2882, opened 2023-08-09
fix(cli): Add missing layer titles from the imagery to tileset.

Wentao-Kuang:


packages/config/src/memory/memory.config.ts
```diff
@@ -174,7 +174,7 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
         category: i.category,
         name: targetName,
         format: ImageFormat.Webp,
-        layers: [{ name: targetName, minZoom: 0, maxZoom: 32 }],
+        layers: [{ name: targetName, title: i.title, minZoom: 0, maxZoom: 32 }],
```
blacha, line 177: The All tileset should not be able to be viewed as a big layer, we could set the minZoom: 32?

packages/config/src/memory/memory.config.ts
```diff
@@ -152,7 +152,7 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
     const allTileset: ConfigTileSet = {
       type: TileSetType.Raster,
       id: 'ts_all',
-      name: 'all_imagery',
+      name: 'all',
       title: 'All Imagery Basemaps',
```
blacha, line 156: Everything is a basemap? could this just be "All Imagery"

packages/config/src/memory/memory.config.ts
Wentao-Kuang: Yes, this already been reset to zoom 32-32 in line 149 when creating all tileset.
This is creating individual tileset so it is zoom 0-32.

blacha: ahh the github diff showed it as the same block :face_exhaling:

blacha (approved):

PR #2886, opened 2023-08-13
feat(landing): Ignore layers in the layer drop down selector.

Wentao-Kuang:


packages/landing/src/components/layer.switcher.dropdown.tsx
```diff
@@ -17,6 +17,9 @@ export interface LayerSwitcherDropdownState {
   zoomToExtent: boolean;
   currentLayer: string;
 }
+
+const ignoredLayers = ['all'];
```
blacha, line 21: `new Set(['all'])` granted it doenst matter where `n=1` but if this ever gets to be bigger than 5 or so items it could become problematic.

blacha (approved):

PR #2900, opened 2023-08-21
fix(cli): Run npm install in the basemaps-config to install the @linz/style

Wentao-Kuang:
### Why?
When we make changes to config file in the basemaps-config repo, we need to format it before commit. Current way of using prettier is broken because we need to install @linzjs/style into the cli docker container. A better solution is just run prettier in the config directory before commit the changes.

### Description of Changes:
- Remove all the prettier stuff from cli
- Run npx prettier in the chirld process in config repo.

#### Author Checklist:
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in PR title

[BM-810]: https://toitutewhenua.atlassian.net/browse/BM-810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/cli/src/cli/github/make.cog.pr.ts
+++ b/packages/cli/src/cli/github/make.cog.pr.ts
@@ -74,9 +74,9 @@ export class MakeCogGithub extends Github {
 
     // Commit and push the changes
     const message = `config(raster): Add imagery ${this.imagery} to ${filename} config file.`;
-    // this.commit(message);
-    // this.push();
-    // await this.createPullRequests(branch, message, false);
+    this.commit(message);
+    this.push();
+    await this.createPullRequests(branch, message, false);
   }
 
   /**
--- a/packages/cli/src/cli/github/make.cog.pr.ts
+++ b/packages/cli/src/cli/github/make.cog.pr.ts
@@ -14,6 +14,7 @@ import { execFileSync } from 'child_process';
 
 export class MakeCogGithub extends Github {
   imagery: string;
+  _formatInstalled = false;
   constructor(imagery: string, repo: string, logger: LogType) {
     super(repo, logger);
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli/github/make.cog.pr.ts
```diff
@@ -23,6 +23,8 @@ export class MakeCogGithub extends Github {
    * Format the config files by prettier
    */
   formatConfigFile(path = './config/'): void {
+    this.logger.info({ repository: this.repo }, 'GitHub: Npm Install');
+    execFileSync('npm', ['install', '--include=dev'], { cwd: this.repoName });
```
blacha, line 27: this should only happen once, not everytime this is run.

Wentao-Kuang: We will cal this `formatConfigFile` function only once every time we run this cli. But I can split this into a seperate function after clone the basemaps-confg repo.

blacha: that would leave a gotcha in this code path, you could cache it with a simple promise

```typescript
if (this._formatInstall) await this._formatInstall
else this._formatInstall = Promise.resolve().then(f => execFileSync('npm', ['install', '--include=dev'], { cwd: this.repoName })
```

although this is all sync code, so even a `this._formatInstalled = true` would do the trick

Wentao-Kuang: Got this updated to use `this._formatInstalled = true`. Thanks.

blacha (approved):

PR #2903, opened 2023-08-21
feat(cogify): ensure cogify path-like args have trailing slashes. BM-858

ajacombs:
#### Description
*What does this change aim to achieve?*

Updates the cogify CLI commands to ensure that arguments representing folder paths have trailing slashes. Adds a `UrlFolder` parser for cmd-ts which parses the input string as a URL, and appends a trailing slash if it's missing. Changing the type from `string` to `URL` (via the `Url` or new `UrlFolder` parsers) involved changing places that use these values to construct other paths. Where these are passed to `fsa` functions, these are converted back to strings using `urlToString`.

#### Intention
*Why is this change being made? What implications or other considerations are there?*

Parts of the cogify commands were using strings for paths, while other parts were using URLs. This led to a situation where a string path was created by `fsa.joinAll()` which did not have a trailing slash, which was then passed to the URL constructor, resulting in a URL one directory higher than intended being created, which prevented finding the STAC collection.json file.

The calls to `urlToString` will be unnecessary once we upgrade to chunkd v11 which swaps to receiving URLs directly, however that is a much larger piece of work so for the time being we'll continue to gradually migrate to using URLs and use `urlToString` as needed when passing values to `fsa` functions.

#### Checklist
- [ ] Tests updated, N/A: no existing tests to update.
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.cover.ts
+++ b/packages/cogify/src/cogify/cli/cli.cover.ts
@@ -7,11 +7,12 @@ import { Metrics } from '@linzjs/metrics';
 import { command, number, oneOf, option, optional, restPositionals, string } from 'cmd-ts';
 import { isArgo } from '../../argo.js';
 import { CutlineOptimizer } from '../../cutline.js';
+import { urlToString } from '../../download.js';
 import { getLogger, logArguments } from '../../log.js';
-import { TileCoverContext, createTileCover } from '../../tile.cover.js';
-import { createFileStats } from '../stac.js';
-import { Url } from '../parsers.js';
 import { Presets } from '../../preset.js';
+import { TileCoverContext, createTileCover } from '../../tile.cover.js';
+import { Url, UrlFolder } from '../parsers.js';
+import { createFileStats } from '../stac.js';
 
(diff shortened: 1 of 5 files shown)


packages/cogify/src/cogify/cli/cli.cover.ts
```diff
@@ -75,15 +76,15 @@ export const BasemapsCogifyCoverCommand = command({
 
     const res = await createTileCover(ctx);
 
-    const targetPath = fsa.joinAll(args.target, String(tms.projection.code), im.name, CliId);
+    const targetPath = new URL(`${tms.projection.code}/${im.name}/${CliId}/`, args.target);
```
Wentao-Kuang, line 79: Does this need to be a `UrlFolder` as you defined in the args, or you just need the args.target to be end with '/'

ajacombs: `UrlFolder` isn't it's own type, it's a parser which takes the string from the CLI and turns it into a `URL` while ensuring it ends with a slash. So the type of `args.target` here is `URL`.

Wentao-Kuang (approved):

PR #2906, opened 2023-08-23
feat(landing): Hide the topo grid layers from the layer drop down list.

Wentao-Kuang:
#### Description
*After imagery configuration refactoring, the topo grid layers been bundle into layer drop down, which is not related. We should hide it.*

#### Intention
*We don't want to provide public access for the topo grid layers in the drop down list.*

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


blacha:
I am a little hesitant to have these hidden here, this is quite specific for our implementation, Can we hide them in the linz/basemaps-config ?

blacha:
layers have been removed already.

PR #2911, opened 2023-08-24
feat(cogify): add --concurrency to allow concurrent gdal_translates

blacha:
#### Description

Adds `--concurrency` to all multiple `gdal_translate` to run at the same time.

#### Intention
In COG Creation `gdal_translate` often has peaks of CPU usage across all cores and then goes down to 1 core for long periods, by running multiple `gdal_translates` on the same machine we can increase the usage of the CPU

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.cog.ts
+++ b/packages/cogify/src/cogify/cli/cli.cog.ts
@@ -124,7 +124,7 @@ export const BasemapsCogifyCreateCommand = command({
 
     try {
       await mkdir(tmpFolder, { recursive: true });
-      // TODO should COG creation be run concurrently?
+
       const promises = filtered.map(async (f) => {
         const { item, url } = f;
         const cutlineLink = getCutline(item.links);
--- a/packages/cogify/src/cogify/cli/cli.cog.ts
+++ b/packages/cogify/src/cogify/cli/cli.cog.ts
@@ -3,9 +3,10 @@ import { LogType, fsa } from '@basemaps/shared';
 import { CliId, CliInfo } from '@basemaps/shared/build/cli/info.js';
 import { CogTiff, TiffTag } from '@cogeotiff/core';
 import { Metrics } from '@linzjs/metrics';
-import { command, flag, option, optional, restPositionals } from 'cmd-ts';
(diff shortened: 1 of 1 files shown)


packages/cogify/src/cogify/cli/cli.cog.ts
```diff
@@ -63,6 +64,13 @@ export const BasemapsCogifyCreateCommand = command({
     ...logArguments,
     path: restPositionals({ type: Url, displayName: 'path', description: 'Path to covering configuration' }),
     force: flag({ long: 'force', description: 'Overwrite existing tiff files' }),
+    concurrency: option({
+      type: number,
+      long: 'concurrency',
+      description: 'Number of COGS to process at the same type',
+      defaultValue: () => 4,
```
ajacombs, line 71: Would setting `defaultValue: () => os.cpus().length` be a worthwhile default? Or is that likely to do funky things in cloud nodes where "the number of CPU cores available" may be a more nebulous concept?

blacha: The way we currently run this is we put 4+ containers on the same host, the host will have 100+ cores, I don't think we can run 400+ `gdal_translates` on the same host.

We generally only batch these containers into 10-20 tiffs per container so having a cpu().length would mean just run them all at the same time all the time!

ajacombs: Ah yup, that makes sense.

ajacombs (approved):

Wentao-Kuang:
I am not sure this is going to work with gdal setting `NUM_THREADS=ALL_CPUS`. Did you tested it?

blacha:
> I am not sure this is going to work with gdal setting `NUM_THREADS=ALL_CPUS`. Did you tested it?

`gdal_tranlsate` has peaks of using all cpus, but there is till a lot of single threaded logic, so we get spikes on all cpus then down to nothing, runnning a few gdal commands at the same time helps with the peaks.

It also means `gdal_translate` can be running while we are uploading or downloading data. it is exactly how the old cog command worked.

PR #2915, opened 2023-08-24
feat(config): detect empty images and ignore them

blacha:
#### Description

Sometimes when tiffs are create they are created outside the bounds of where the source tiff actually has data. 

`gdal_translate` will then create a completly empty tiff, as the tiffs are created with `sparse_ok=true` the internal tiff tag `TileOffsets`  which contains the file offsets for all the internal tiles will contain offsets of `0` with byte length of `0`.

If all the offsets are `0` then there is no data inside the tiff.

#### Intention

We are seeing a large number of empty tiffs being imported into our configuration and until we can clean out the source locations it is easier to check the tiff to see if it contains data, when we are creating the configuration.

I have tested this with the current [linz/basemaps-config](https://github.com/linz/basemaps-config) configs and it found 4,548 empty tiff files ranging in size from 1KB to 250KB.

#### Checklist
- [x] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/config/src/json/json.config.ts
+++ b/packages/config/src/json/json.config.ts
@@ -59,7 +59,7 @@ const SmallTiffSizeBytes = 256 * 1024;
 async function isEmptyTiff(path: string): Promise<boolean> {
   const tiff = await CogTiff.create(fsa.source(path));
 
-  // Starting the smallest tiff overview greatly reduces teh amount of data needing to be read
+  // Starting the smallest tiff overview greatly reduces the amount of data needing to be read
   // if the tiff contains data.
   for (const img of tiff.images.reverse()) {
     const tileOffsets = img.tileOffset;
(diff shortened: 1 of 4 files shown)


packages/config/src/json/json.config.ts
```diff
@@ -41,6 +43,37 @@ export function guessIdFromUri(uri: string): string | null {
...
+ *
+ * This looks at the first few overview levels to see if any internal tiff tiles contain data
+ *
+ * @param path path to check
+ * @returns true if the tiff is empty, false otherwise
+ */
+async function isEmptyTiff(path: string): Promise<boolean> {
+  const tiff = await CogTiff.create(fsa.source(path));
+
+  // Starting the smallest tiff overview greatly reduces teh amount of data needing to be read
```
ajacombs, line 62:
```suggestion
  // Starting the smallest tiff overview greatly reduces the amount of data needing to be read
```
Typo

ajacombs (approved):

ajacombs (approved):

PR #2921, opened 2023-08-28
feat(lambda-tiler): create preview images for og:image BM-264

blacha:
#### Description

Adds a new endpoint `/v1/preview` which generates `og:image` previews for a specific location.

```
/v1/preview/:tileSet/:tileMatrixSet/:z/:lon/:lat
```

example:

```
/v1/preview/aerial/WebMercatorQuad/177.3998405/-39.0852555
```

#### Intention

When people share links to basemaps they are currently presented with the same image of Lyttelton, which while really nice looking is not representative of what they are actually get a link to.

This is the first step in providing an actual image of the area people are linked to, I will add a screenshot tester of this too.

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.config.ts
+++ b/packages/cogify/src/cogify/cli/cli.config.ts
@@ -51,13 +51,18 @@ export const BasemapsCogifyConfigCommand = command({
 
     const configPath = base58.encode(Buffer.from(outputPath));
     const location = getImageryCenterZoom(im);
-    const locationHash = `#@${location.lat.toFixed(7)},${location.lon.toFixed(7)},z${location.zoom}`;
+    const lat = location.lat.toFixed(7);
+    const lon = location.lon.toFixed(7);
+    const locationHash = `#@${lat},${lon},z${location.zoom}`;
     const url = `https://basemaps.linz.govt.nz/?config=${configPath}&i=${im.name}&tileMatrix=${im.tileMatrix}&debug${locationHash}`;
+    const preview = `https://basemaps.linz.govt.nz/v1/preview/${im.name}/${im.tileMatrix}/${location.zoom}/${lon}/${lat}.webp?config=${configPath}`;
+
     logger.info(
       {
         imageryId: im.id,
         path: outputPath,
         url,
+        urlPreview: preview,
         config: configPath,
         title: im.title,
         tileMatrix: im.tileMatrix,
--- a/packages/lambda-tiler/src/routes/preview.ts
+++ b/packages/lambda-tiler/src/routes/preview.ts
@@ -13,7 +13,6 @@ export interface PreviewGet {
   Params: {
     tileSet: string;
     tileMatrix: string;
-    imageFormat: string;
     lat: string;
     lon: string;
     z: string;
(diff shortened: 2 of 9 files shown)


packages/lambda-tiler/src/routes/preview.ts
```diff
@@ -0,0 +1,157 @@
...
+  Params: {
+    tileSet: string;
+    tileMatrix: string;
+    lat: string;
+    lon: string;
+    z: string;
+  };
+}
+
+const PreviewSize = { width: 1200, height: 630 };
```
Wentao-Kuang, line 22: How do we know this is the size we need?

blacha: Google says its the size of open graph images https://www.google.com/search?q=og%3Aimage+size

We could look at letting it be a customized size in the future?

packages/cogify/src/cogify/cli/cli.config.ts
```diff
@@ -51,13 +51,18 @@ export const BasemapsCogifyConfigCommand = command({
 
     const configPath = base58.encode(Buffer.from(outputPath));
     const location = getImageryCenterZoom(im);
-    const locationHash = `#@${location.lat.toFixed(7)},${location.lon.toFixed(7)},z${location.zoom}`;
+    const lat = location.lat.toFixed(7);
+    const lon = location.lon.toFixed(7);
+    const locationHash = `#@${lat},${lon},z${location.zoom}`;
     const url = `https://basemaps.linz.govt.nz/?config=${configPath}&i=${im.name}&tileMatrix=${im.tileMatrix}&debug${locationHash}`;
+    const preview = `https://basemaps.linz.govt.nz/v1/preview/${im.name}/${im.tileMatrix}/${location.zoom}/${lon}/${lat}.webp?config=${configPath}`;
```
Wentao-Kuang, line 58: I wonder why do we want to change this as preview, don't we want a basemaps link that we can browse around?

blacha: It spits out both, but means the slack alerts can come with a image until we add the direct link support

Wentao-Kuang: I see, this is only for the logs.

blacha: Its also a additive change, so the old URL is still there.

packages/lambda-tiler/src/routes/preview.ts
```diff
@@ -0,0 +1,157 @@
...
+const OutputFormat = ImageFormat.Webp;
+
+/**
+ * Serve a preview of a imagery set
+ *
+ * /v1/preview/:tileSet/:tileMatrixSet/:z/:lon/:lat
+ *
+ * @example
+ * Raster Tile `/v1/preview/aerial/WebMercatorQuad/177.3998405/-39.0852555`
```
Wentao-Kuang, line 31: missing a zoom in here?

packages/lambda-tiler/src/routes/preview.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  req.set('layersUsed', overlays.length);
+  req.set('bytes', buf.byteLength);
+  const response = new LambdaHttpResponse(200, 'ok');
+  response.header(HttpHeader.ETag, cacheKey);
+  response.header(HttpHeader.CacheControl, 'public, max-age=604800, stale-while-revalidate=86400');
+  response.buffer(buf, 'image/' + ctx.outputFormat);
+
+  const shortLocation = [ctx.location.lon.toFixed(7), ctx.location.lat.toFixed(7)].join('_');
+  const suggestedFileName = `preview_${ctx.tileSet.name}_z${ctx.z}_${shortLocation}.webp`;
```
Wentao-Kuang, line 153: `.webp` -> `.ctx.outputFormat`

PR #2922, opened 2023-08-29
feat(lambda-tiler): suggest filenames when tiles are downloaded

blacha:
#### Description

Adds a `Content-Disposition` for all xyz tile URLS to provide nicer filenames

#### Intention

When debugging tiles I often download a tile but the filename is just the `:y.png` this adds a suggested filename to all xyz raster tiles in the format `:tileSet_:z_:x_:y.:ext` to make it easier to filter through downloaded tiles.

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/lambda-tiler/src/routes/__tests__/xyz.test.ts
+++ b/packages/lambda-tiler/src/routes/__tests__/xyz.test.ts
@@ -74,6 +74,7 @@ o.spec('/v1/tiles', () => {
       o(res.header('content-type')).equals(`image/${fmt}`);
       o(res.header('etag')).notEquals(undefined);
       o(res.header('cache-control')).equals('public, max-age=604800, stale-while-revalidate=86400');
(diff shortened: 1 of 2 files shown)


packages/lambda-tiler/src/routes/__tests__/xyz.test.ts
```diff
@@ -110,6 +111,7 @@ o.spec('/v1/tiles', () => {
     o(res.header('content-type')).equals('image/png');
     o(res.header('etag')).notEquals(undefined);
     o(res.header('cache-control')).equals('public, max-age=604800, stale-while-revalidate=86400');
+    o(res.header('content-disposition')).equals('inline; filename="🦄 🌈_0_0_0.png"');
```
blacha, line 114: TODO: I am not sure we are allowed to just jam UTF8 characters into filenames, maybe we need to encode them first.

blacha:
UTF8 is not really supported in content-disposition, this logic needs to be updated.

PR #2925, opened 2023-08-30
feat: add og:image preview to all basemaps links BM-264

blacha:
#### Description

Opengraph allows third party websites/applications to embed useful information about your page into their page.

Currently basemaps serves a default image to all of these applications  https://basemaps.linz.govt.nz/basemaps-card.jpeg which is a nice picture of Lyttleon rather than what is actually viewed.

Adding a picture that is a preview of what the link displays is very useful with tools like slack, where often I take a screen shot of the page and then paste a link with the screenshot.

Example Links

#### Slack
image

#### LinkedIn 

image

#### Intention

Because the location of the page is stored only in a hash `#@...` this is never sent to the server so the server cannot update the `index.html`'s `og:image` location.

This change modifies the default URL from `#@...` to `/@`  which converts it from a hash to a path. The path is then sent to the server which processes it updates the latest index.html and serves it back to the client.

The current lambda function does not have direct access to the `index.html` so it has to fetch it from `s3://` if this fetch fails, the user is just redirected back to the landing page. This could be improved in the future by letting the lambda contain a `index.html` but it breaks the deployment separation of the landing/tile server.

#### Checklist
- [x] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title


packages/_infra/src/edge/index.ts
```diff
@@ -101,7 +102,26 @@ export class EdgeStack extends cdk.Stack {
...
+        {
+          pathPattern: '/@*',
+          allowedMethods: cf.CloudFrontAllowedMethods.ALL,
+          forwardedValues: {
+            /** Forward all query strings but do not use them for caching */
+            queryString: true,
+            queryStringCacheKeys: [
+              'config',
+              'exclude',
+              'tileSet',
```
blacha, line 118:
```suggestion
              'tileMatrix',
```

packages/lambda-tiler/src/routes/__tests__/preview.index.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  let lastLocation: string | undefined;
+  o.beforeEach(() => {
+    fsa.register('memory://', fsMem);
+    lastLocation = process.env[Env.StaticAssetLocation];
+  });
+  o.afterEach(() => {
+    if (lastLocation == null) delete process.env[Env.StaticAssetLocation];
+    else process.env[Env.StaticAssetLocation] = lastLocation;
+  });
+  //   const sandbox = sinon.createSandbox();
```
blacha, line 30:
```suggestion
```

packages/geo/src/url.ts
```diff
@@ -0,0 +1,174 @@
...
+   * TileMatrixSet identifier
+   *
+   * @example
+   * "WebMercatorQuad"
+   * "NZTM2000Quad"
+   */
+  tileMatrix: string;
+}
+
+export const LocationUrl = {
```
blacha, line 58: LocationSlug ?

packages/landing/static/index.html
```diff
@@ -7,11 +7,9 @@
     <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
 
     <meta property="og:site_name" content="LINZ Basemaps" />
-
-    <meta name="twitter:card" content="summary_large_image" /> 
-    <meta name="twitter:title" property="og:title" content="LINZ Basemaps" /> 
-    <meta name="twitter:image" property="og:image" content="/basemaps-card.jpeg" />
-    <meta name="twitter:description" property="og:description" content="NZ’s authoritative and open digital basemap service for LINZ and the public." /> 
+    <meta property="og:title" content="LINZ Basemaps" />
```
blacha, line 10: Question, is `twitter:card` actually needed?

ajacombs (approved):

PR #2936, opened 2023-09-05
Revert "feat(lambda-tiler): try arm based lambdas (#2910)"

blacha:
This reverts commit e5bd68df37f487c4cc543c69fbedc66f13c5fdc0.


blacha:
ARM lambda functions seem to be working pretty well, so we don't need to revert them!

PR #2947, opened 2023-09-12
Fix(cli): Import config to get target zoom from covering.json

Wentao-Kuang:
#### Description
The new cogify workflow deprecate the old job.json and use covering.json to prepare jobs. We need fix the import to be able to support both job.json and covering.json. 

#### Intention
We try to get gsd from job.json first, if not exist we loop for covering.json. Also, I have moved the prepare url inside the if condition to stop these s3 read and only try to get job.json if there is a change in config. 

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cli/src/cli/config/action.import.ts
+++ b/packages/cli/src/cli/config/action.import.ts
@@ -261,9 +261,11 @@ export class CommandImport extends CommandLineAction {
     const aerialId = 'ts_aerial';
     const newData = await mem.TileSet.get(aerialId);
     const oldData = await cfg.TileSet.get(aerialId);
+    const aerialLayers: Set<string> = new Set<string>();
     if (newData == null || oldData == null) throw new Error('Failed to fetch aerial config data.');
     for (const layer of newData.layers) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli/config/action.import.ts
```diff
@@ -307,14 +317,25 @@ export class CommandImport extends CommandLineAction {
   }
 
   _jobs: Map<string, CogStacJob> = new Map<string, CogStacJob>();
-  async _loadJob(path: string): Promise<CogStacJob> {
+  async _loadJob(path: string): Promise<CogStacJob | undefined> {
     const existing = this._jobs.get(path);
     if (existing) return existing;
+    const exist = await fsa.exists(path);
```
blacha, line 323: this does two requests, and in theory the time between the exists and read the state could change (eg file deleleted or moved).

As a general rule you should attempt to read the file and look for a `404` if a error happens.

Wentao-Kuang: Good point, I have remove the check exist and move read into try. Return null if failed to get the job.json.

blacha (approved):

PR #2969, opened 2023-09-24
fix(lambda-tiler): Catch the error code while reading the config file in api. BM-898

Wentao-Kuang:
#### Description
When failed to get config file from s3, return correct error code in logs and remove the 500 error in #proj-basemaps-alerts

#### Intention
Catch the error from `fsa.readJson(config)`, and throw correct error code and message.

#### Checklist
- [X] Tests updated
- [ ] Docs updated
- [X] Issue linked in Title

### DIFF
--- a/packages/lambda-tiler/src/util/config.loader.ts
+++ b/packages/lambda-tiler/src/util/config.loader.ts
@@ -46,10 +46,14 @@ export class ConfigLoader {
 
     req.set('config', configLocation);
     req.timer.start('config:load');
-    return CachedConfig.get(configLocation).then((f) => {
-      req.timer.end('config:load');
-      if (f == null) throw new LambdaHttpResponse(404, `Config not found at ${configLocation}`);
-      return f;
-    });
+    return CachedConfig.get(configLocation)
+      .then((f) => {
+        req.timer.end('config:load');
+        if (f == null) throw new LambdaHttpResponse(404, `Config not found at ${configLocation}`);
+        return f;
+      })
+      .catch((e) => {
+        throw new LambdaHttpResponse(e.status, e.statusDescription);
+      });
   }
 }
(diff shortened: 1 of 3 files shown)


packages/lambda-tiler/src/util/config.loader.ts
```diff
@@ -46,10 +46,14 @@ export class ConfigLoader {
...
     req.timer.start('config:load');
-    return CachedConfig.get(configLocation).then((f) => {
-      req.timer.end('config:load');
-      if (f == null) throw new LambdaHttpResponse(404, `Config not found at ${configLocation}`);
-      return f;
-    });
+    return CachedConfig.get(configLocation)
+      .then((f) => {
+        req.timer.end('config:load');
+        if (f == null) throw new LambdaHttpResponse(404, `Config not found at ${configLocation}`);
```
blacha, line 52: Would `400` be more appropriate ? as the parameter is invalid?

packages/lambda-tiler/src/util/config.loader.ts
```diff
@@ -46,10 +46,14 @@ export class ConfigLoader {
...
+        return f;
+      })
+      .catch((e) => {
+        throw new LambdaHttpResponse(e.status, e.statusDescription);
```
blacha, line 56: this seems risky as `e` could be anything... if it was a string this response would be undefined, undefined which may break other things?

Wentao-Kuang: Shall we both throw the same error. 400 with `Invalid config location at ${configLocation}`

blacha: I think this catch is risky, we should handle the permission denied when we are actually reading the file

blacha (approved):

PR #2976, opened 2023-10-10
feat(cli): Allow bundle to guess the imagery ID from the url with slash ending.

Wentao-Kuang:
#### Motivation
We are trying to add slash at the end of all the urls in the basemaps-configs. 
https://github.com/linz/basemaps-config/pull/809

#### Modification

This will fix the bundle to find the correct imagery id from the both new and old url.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/config/src/json/json.config.ts
+++ b/packages/config/src/json/json.config.ts
@@ -33,7 +33,8 @@ const Q = PLimit(10);
 
 export function guessIdFromUri(uri: string): string | null {
   const parts = uri.split('/');
-  const id = parts.pop();
+  let id = parts.pop();
+  if (uri.endsWith('/')) id = parts.pop();
 
   if (id == null) return null;
   try {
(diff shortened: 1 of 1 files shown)


packages/config/src/json/json.config.ts
```diff
@@ -33,7 +33,8 @@ const Q = PLimit(10);
 
 export function guessIdFromUri(uri: string): string | null {
   const parts = uri.split('/');
-  const id = parts.pop();
+  let id = parts.pop();
+  if (uri.endsWith('/')) id = parts.pop();
```
blacha, line 37:
```suggestion
  const id = uri.endsWith('/') ? parts.at(-2) : parts.at(-1)
```

Avoids the double pop.

blacha (approved):

PR #2982, opened 2023-10-29
feat: allow reads from linz "-scratch" buckets TDE-906

blacha:
#### Motivation

LINZ is starting to suffix s3 buckets with a rough indication to the quality of data inside of the bucket a "scratch" bucket is a temporary storage location and should have a data lifecycle rule set to delete data after 90/180 days.

#### Modification

Adds Workflows and Basemaps scratch buckets to allowed locations

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/_infra/src/config.ts
+++ b/packages/_infra/src/config.ts
@@ -15,7 +15,14 @@ export interface BaseMapsConfig {
 }
 
 export const BaseMapsProdConfig: BaseMapsConfig = {
-  CogBucket: ['linz-basemaps', 'linz-basemaps-vector', `linz-basemaps-staging`, 'linz-workflow-artifacts'],
+  CogBucket: [
+    'linz-basemaps',
+    'linz-basemaps-vector',
+    `linz-basemaps-staging`,
+    `linz-basemaps-scratch`,
+    'linz-workflow-artifacts',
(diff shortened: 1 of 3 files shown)


packages/_infra/src/config.ts
```diff
@@ -15,7 +15,14 @@ export interface BaseMapsConfig {
...
 
 export const BaseMapsProdConfig: BaseMapsConfig = {
-  CogBucket: ['linz-basemaps', 'linz-basemaps-vector', `linz-basemaps-staging`, 'linz-workflow-artifacts'],
+  CogBucket: [
+    'linz-basemaps',
+    'linz-basemaps-vector',
+    `linz-basemaps-staging`,
+    `linz-basemaps-scratch`,
+    'linz-workflow-artifacts',
+    'linz-workflows-scratch',
```
Wentao-Kuang, line 24: Why the scratch workflow bucket have `s` after `workflows`, does it not stay the same with artifacts bucket?

blacha: We renamed the cluster to have a trailing "s" so we updated the bucket name at the same time.

Wentao-Kuang (approved):

PR #2984, opened 2023-11-01
feat(cli): Relocate dgal and cog from cli package. BM-879

Wentao-Kuang:
#### Motivation
dgal and cog are using for bathymetry package, we don't need in the cli package. So can relocate them.

#### Modification
- Move dgal into bathymetry package.
- Move types of the cog job in to shared.
- Remove the git from container as the pull request logic already inside argo-tasks
- Remove the some tests.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


packages/cli/src/cli/config/action.import.ts
```diff
@@ -17,7 +17,7 @@ import fetch from 'node-fetch';
 import { invalidateCache } from '../util.js';
 import { Q, Updater } from './config.update.js';
 import { FeatureCollection } from 'geojson';
-import { CogJobJson } from '../../cog/types.js';
+import { CogJobJson } from '@basemaps/shared/build/cog/types.js';
```
blacha, line 20: We should export this from `@basemaps/shared` we should avoid the imports from /build unless we have a really good reason for them

blacha (approved):

PR #2990, opened 2023-11-03
feat: Add proof of concept docs site BM-917

ajacombs:
#### Motivation

This adds an initial proof of concept of a documentation site. The documentation is generated from a combination of static Markdown documents; currently lightly modified from the existing files, plus a few placeholders.

API documentation is generated from the typescript source code using Typedoc. This is rendered to Markdown files using typedoc-plugin-markdown, which is combined with the static Markdown files to render the documentation site using MkDocs.

#### Modification

This is a rough proof of concept implementation, to be fleshed out in subsequent PRs which will:
- Add more documentation.
- Improve the readability of the generated API docs (fix typographic hierarchy and character escaping issues).
- Update the theme to match LINZ branding.

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


.github/workflows/push.yml
```diff
@@ -25,6 +25,17 @@ jobs:
         with:
           package-manager: yarn
 
+      - name: Build docs
+        run: |
+          npx typedoc
+          docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build
```
blacha, line 31: could we pin this to a version `squidfunk/mkdocs-material:v12345` a specific hash may actually be better.

docs/User guide/README.md
```diff
@@ -0,0 +1,7 @@
+# User guide
```
blacha, line 1: What options do we have for not having paths with capitals and spaces in them? as the folder structure then varies from the resto of the repo.

docs/Operator guide/Empty tiles.md
```diff
@@ -10,7 +10,7 @@ Further investigation should also be done on `OATiles-hint: empty` response head
 
 LINZ has imagery datasets which are very sparse, with large empty areas between smaller sections of valid data, for example [Gisborne Urban 2022-2023](https://basemaps.linz.govt.nz/?i=gisborne-2022-2023-0.1m).
 
-[![Gisborne 2023](./static/2023-06-26-gisborne-2023.png)](https://basemaps.linz.govt.nz/?i=gisborne-2022-2023-0.1m)
+[![Gisborne 2023](./images/2023-06-26-gisborne-2023.png)](https://basemaps.linz.govt.nz/?i=gisborne-2022-2023-0.1m)
```
blacha, line 13: I would prefer this to be left as static to be consisetent with landing and also if we want to put other static items into it, eg example tar or config files.

.github/workflows/push.yml
ajacombs: https://hub.docker.com/r/squidfunk/mkdocs-material/tags

Looks like it uses semver, so we could pin it to `9`, `9.4`, `9.4.7`, or the latest hash `91c415cc8084`. Pinning to the major version seems like the best option?

docs/User guide/README.md
ajacombs: Have updated the naming to kebab case to match the rest of the repo.

The labels in the sidebar nav for pages are pulled from the H1 given at the start of the page (if present), and for directories it seems to be converted to sentence case.

I kept the `API` dir as uppercase, as otherwise it was converted to `Api` in the sidebar.

It also has the benefit of using the file names as the URL, so for example `empty-tiles.md` becomes `http://0.0.0.0:8000/operator-guide/empty-tiles/` with a label of `Sparse Datasets and empty tiles` in the sidebar

docs/Operator guide/Empty tiles.md
ajacombs: Agreed, have renamed to `static`.

blacha (approved):

.github/workflows/push.yml
ajacombs: Pinned to 9.4 following slack discussion

PR #3010, opened 2023-11-15
docs: operator quick start config guide

blacha:
#### Motivation

Expand the quick start guide to include how to bundle configuration

#### Modification

- Applies prettier to all of /docs
- adds quick start guide for bundling config

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


docs/operator-guide/quick-start.md
```diff
@@ -104,3 +104,51 @@ It is not recommended to run `:latest` for any prolonged period of time, all of
...
+{
+  "type": "raster",
+  "id": "ts_basic",
+  "title": "Basic Imagery Basemap",
+  "layers": [
+    {
+      // Source location for EPSG:2193 Imagery, has to be absolute
+      "2193": "/home/user/basemaps-quick-start/imagery/otago_2017-2019_0.3m/",
+      // url/slug friendly name of the imagery set
+      "name": "otago_2017-2019_0.3m",
```
Wentao-Kuang, line 128: I think we not using only hyphen for the image names to be url friendly

Wentao-Kuang (approved):

docs/operator-guide/quick-start.md
blacha: I think both `name` and `title` sholuld be optional, and fetched from the STAC, they should only be used if they are a override.

this specific name is only there because its "required"

PR #3016, opened 2023-11-27
feat(cli): compare config bundles using --target config.json

blacha:
#### Motivation

It is useful to compare configuration bundles, but they contain too many ids and values that could rotate. The import command can compare two configurations while ignoring the changes in ids.

#### Modification

compare config bundles using `--target config.json`

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cli/src/cli/config/action.bundle.ts
+++ b/packages/cli/src/cli/config/action.bundle.ts
@@ -41,9 +41,8 @@ export class CommandBundle extends CommandLineAction {
     const config = this.config.value ?? DefaultConfig;
     const bundle = this.output.value ?? DefaultOutput;
     const mem = await ConfigJson.fromPath(config, logger);
+    if (this.assets.value) mem.assets = this.assets.value;
     const configJson = mem.toJson();
-    const assets = this.assets.value;
-    if (assets) configJson.assets = assets;
     await fsa.writeJson(bundle, configJson);
     logger.info({ path: bundle }, 'ConfigBundled');
     return;
--- a/packages/cli/src/cli/config/action.import.ts
+++ b/packages/cli/src/cli/config/action.import.ts
@@ -72,7 +72,7 @@ export class CommandImport extends CommandLineAction {
     });
   }
 
-  async getConfig(logger:LogType): Promise<BasemapsConfigProvider> {
+  async getConfig(logger: LogType): Promise<BasemapsConfigProvider> {
     if (this.target.value) {
       logger.info({ config: this.target.value }, 'Import:Target:Load');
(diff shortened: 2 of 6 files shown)


packages/cli/src/cli/config/action.bundle.ts
```diff
@@ -41,9 +41,8 @@ export class CommandBundle extends CommandLineAction {
     const config = this.config.value ?? DefaultConfig;
     const bundle = this.output.value ?? DefaultOutput;
     const mem = await ConfigJson.fromPath(config, logger);
+    if (this.assets.value) mem.assets = this.assets.value;
```
blacha, line 44: assets locations were not being hashed as part of the config, this changes that.

packages/cli/src/cli/config/config.diff.ts
```diff
@@ -32,7 +32,7 @@ export class ConfigDiff {
   }
 
   static showDiff<T extends { id: string }>(type: string, oldData: T, newData: T, logger: LogType): boolean {
-    const changes = diff.diff(oldData, newData, (_path: string[], key: string) => IgnoredProperties.indexOf(key) >= 0);
+    const changes = diff.diff(oldData, newData, (_path: string[], key: string) => IgnoredProperties.has(key));
```
blacha, line 35: Set was slightly faster in my local testing,

packages/cli/src/cli/config/action.import.ts
```diff
@@ -65,6 +66,11 @@ export class CommandImport extends CommandLineAction {
       parameterLongName: '--output',
       description: 'Output a markdown file with the config changes',
     });
+    this.target = this.defineStringParameter({
+      argumentName: 'TARGET',
+      parameterLongName: '--target',
+      description: 'import location',
```
Wentao-Kuang, line 72:
```suggestion
      description: 'Target config file to compare and replace',
```

blacha: I wonder if target is the wrong word here.

Since its not replaced unless its dynamodb

Wentao-Kuang: Yeah, `target` is for the future if we deprecate dyanmodb.

Wentao-Kuang: I think we just remove the word replace for now.
```suggestion
      description: 'Target config file to compare',
```

Wentao-Kuang (dismissed):

Wentao-Kuang (approved):

PR #3018, opened 2023-11-28
fix(cogify): prevent empty tiffs from being stored

blacha:
#### Motivation

As cogify does not  have the exact bounding polygon of the data inside the tiff, there are occasions where a tiff is created outside the extent of the data. this tiff is empty and gdal will create a empty sparse tiff.

These tiffs are useless and should be ignored/deleted.

#### Modification

Check the output tiff to see if its empty before uploading it.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.cog.ts
+++ b/packages/cogify/src/cogify/cli/cli.cog.ts
@@ -212,33 +212,33 @@ export const BasemapsCogifyCreateCommand = command({
 
         const cogStat = await fsa.head(outputTiffPath);
         if (await isEmptyTiff(outputTiffPath)) {
-          // Empty tiff created, removing it from the stac
-          logger.error({ tileId, tiffPath }, 'Cog:Empty');
+          // Empty tiff created, remove it from the stac
+          logger.warn({ tileId, tiffPath }, 'Cog:Empty');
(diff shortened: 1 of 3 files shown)


packages/cogify/src/cogify/cli/cli.cog.ts
```diff
@@ -200,12 +210,19 @@ export const BasemapsCogifyCreateCommand = command({
         item.properties['linz_basemaps:generated']['gdal'] = gdalVersion.stdout;
 
         const cogStat = await fsa.head(outputTiffPath);
+        if (await isEmptyTiff(outputTiffPath)) {
+          // Empty tiff created, removing it from the stac
+          logger.error({ tileId, tiffPath }, 'Cog:Empty');
+          delete item.assets['cog'];
+          item.properties['linz_basemaps:generated'].invalid = 'empty';
+        }
```
blacha, line 218:
```suggestion
          return
        }
```

blacha: need to fix more than this, write the .json

Wentao-Kuang (approved):

PR #3055, opened 2023-12-14
Improve performance by utilizing hardware resources with `Promise.all` 🚀

sanjaiyan-dev:
I noticed that by leveraging the power of `Promise.all`, we can enhance the performance of our application by executing certain functions in parallel. This pull request aims to implement this optimization technique, which will lead to improved efficiency and reduced execution time.

**Changes Made:**
- Introduced the usage of `Promise.all` to execute multiple functions concurrently.

Please let me know if there's anything else you would like to include or modify in the pull request.


blacha:
This has a similar problem to https://github.com/linz/cotar/pull/501 where it doesnt actually increase the performance as most of the time is not spent reading/writing the files.

PR #3058, opened 2023-12-15
fix: Fix the path of the deploy script to use the URL path

Wentao-Kuang:
#### Motivation

The latest fsa using URL instead of string path. We need fix the deploy script as well.

#### Modification

Fix the filePath as URL type.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


Wentao-Kuang:
Oops, making the same fix at the same time. https://github.com/linz/basemaps/pull/3057

PR #3104, opened 2024-01-29
fix!: remove ImageFormat and VectorFormat due to linting errors

blacha:
#### Motivation

Enum types causes issues with future eslint rules we want to introduce its easier to just remove the types.

#### Modification

Removes ImageFormat and VectorFormat enums and replaces them with string types.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/config-loader/src/json/json.config.ts
+++ b/packages/config-loader/src/json/json.config.ts
@@ -117,7 +117,7 @@ export class ConfigJson {
 
     const todo = files.map(async (filePath) => {
       if (!filePath.pathname.endsWith('.json')) return;
-      const bc: BaseConfig = (await fsa.readJson(filePath)) as BaseConfig;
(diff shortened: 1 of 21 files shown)


packages/config-loader/src/json/json.config.ts
```diff
@@ -117,7 +117,7 @@ export class ConfigJson {
 
     const todo = files.map(async (filePath) => {
       if (!filePath.pathname.endsWith('.json')) return;
-      const bc: BaseConfig = (await fsa.readJson(filePath)) as BaseConfig;
+      const bc: BaseConfig = await fsa.readJson(filePath);
```
Wentao-Kuang, line 120:
```suggestion
      const bc: BaseConfig = await fsa.readJson<BaseConfig>(filePath);
```

Wentao-Kuang (approved):

PR #3161, opened 2024-03-03
feat(landing): Enable elevation preview in the basemaps debug page

Wentao-Kuang:
#### Motivation
We want to update the debug page to be able to preview the elevation config.

#### Modification
- Update the tiler stylejson to set the `raster-dem` type source if there is a `terrain-rgb` output pipeline in the config.
- Update debug page to get the terrain source and add terrain control
- Update debug page to add elevation source for the aerial style

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -84,11 +85,32 @@ export async function tileSetToStyle(
...
-    sources: { [styleId]: { type: 'raster', tiles: [tileUrl], tileSize: 256 } },
-    layers: [{ id: styleId, type: 'raster', source: styleId }],
-  };
+  let style;
+
+  if (tsElevation) {
+    const elevationUrl =
+    (Env.get(Env.PublicUrlBase) ?? '') +
+    `/v1/tiles/${tsElevation.name}/${tileMatrix.identifier}/{z}/{x}/{y}.${tileFormat}${query}&pipeline=terrain-rgb`;
+    style = {
```
blacha, line 94: We should use `tileSet.outputs` to determine the sources to and how they are layered

we can hard code terrain-rgb to raster-dem and color-ramp to raster.

packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -84,11 +85,32 @@ export async function tileSetToStyle(
...
+        [styleId + '-terrain']: { type: 'raster-dem', tiles: [elevationUrl], tileSize: 256 },
+      },
+      layers: [{ id: styleId, type: 'raster', source: styleId }],
+      terrain: {
+        source: styleId + '-terrain',
+        exaggeration: 1,
+      },
+    };
+  } else {
+    style = {
```
blacha, line 107: if `?pipeline` is set just throw a error?

packages/landing/src/components/map.tsx
```diff
@@ -204,6 +213,13 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
         Config.map.on('visibleLayers', this.updateVisibleLayers),
       );
 
+      const terrain = this.map.getStyle().terrain;
+      console.log(terrain);
+      if (terrain != null) {
```
blacha, line 218: this should be similar to `ensureGeoControl` where it adds and removes based on the current layer.

packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -102,6 +102,83 @@ export async function tileSetToStyle(
...
+  const query = toQueryString({ config: configLocation, api: apiKey, ...getFilters(req) });
+
+  const styleId = `basemaps-${tileSet.name}`;
+  const sources: Sources = {};
+  const layers: Layer[] = [];
+
+  if (tileSet.outputs) {
+    //for loop output.
+    for (const output of tileSet.outputs) {
+      const format = output.format?.length ? output.format[0] : 'webp';
```
blacha, line 123:
```suggestion
      const format = output.format?.[0] ?? 'webp';
```

packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -102,6 +102,83 @@ export async function tileSetToStyle(
...
+          layout: { visibility: 'none' },
+        });
+      } else {
+        layers.push({
+          id: `${styleId}-${output.name}`,
+          type: 'raster',
+          source: `${styleId}-${output.name}`,
+          layout: { visibility: 'visible' },
+        });
+      }
```
blacha, line 163:
```suggestion
      layers.push({
        id: `${styleId}-${output.name}`,
        type: 'raster',
        source: `${styleId}-${output.name}`,
        layout: { visibility: layers.length === 0 ?  'visible' : 'none' },
      });
    
```

packages/landing/src/components/debug.tsx
```diff
@@ -277,6 +291,22 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
     );
   }
 
+  renderOutputsDropdown(): ReactNode | null {
+    const layers = this.props.map.getStyle().layers;
```
blacha, line 295: should this be a dropdown if there is only one layer? what does it look like with the vector tiles enabled? that would have 100 layers?

Wentao-Kuang: You are right, we should only allow this for raster map. And maybe disable the dropdown if only on later available.

blacha (approved):

PR #3174, opened 2024-03-08
feat(landing): Add Config Debug for screenshot elevation data.

Wentao-Kuang:
#### Motivation

As we want to screenshot some elevation map for smoke tests. We need enable the debug urls for it.

#### Modification

- Add `debug.layer` for screenshot different output.
- Add dropdown for `debug.layer`
- Add `debug.terrain` for screenshot map with terrain on.
- Add dropdown for `debug.terrain`
- Add bearing and pitch in the location url hash
- Comment out the `updateVisibleLayers`, date range slider is not been used atm. This is cause problematic by automatically adding aerial layers.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


packages/landing/src/components/map.tsx
```diff
@@ -41,6 +40,8 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
     const location = Config.map.location;
     this.map.setZoom(location.zoom);
     this.map.setCenter([location.lon, location.lat]);
+    if (location.bearing) this.map.setBearing(location.bearing);
```
blacha, line 43: Does it matter if this doesn't get set? its much safer to do `!= null`

Wentao-Kuang: Yeah, we should just set it when there is a value. It was never been set before this changes. I guess not set just not impact anything and remain the existing bearing of the map. The `setBearing` called when we change the bearing in url.

Wentao-Kuang: Updated with `!= null`

blacha (approved):

PR #3188, opened 2024-03-14
feat(geo): Truncate bearing and pitch values in url.

Wentao-Kuang:
#### Motivation

We need to truncate the bearing and pitch in url which stay the same as maplibre. 1dp for bearing and 0dp for pitch

#### Modification

Update the `fromSlug` function to include the truncate bearing and pitch output in string.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/geo/src/slug.ts
+++ b/packages/geo/src/slug.ts
@@ -50,6 +50,12 @@ export const LocationSlug = {
   /** Number of decimal places to fix a location zoom too */
   ZoomFixed: 2,
 
+  /** Number of decimal places to fix a bearing */
+  BearingFixed: 1,
+
+  /** Number of decimal places to fix a pitch */
+  PitchFixed: 0,
+
   /**
    * Truncate a lat lon based on the zoom level
    *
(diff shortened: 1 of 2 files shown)


packages/geo/src/slug.ts
```diff
@@ -63,7 +69,20 @@ export const LocationSlug = {
...
+   * Truncate a bearing and pitch
+   *
+   * @param loc location to truncate
+   */
+  truncateBearingPitch(loc: LonLatZoom): Partial<LonLatZoom> {
+    let bearing;
+    let pitch;
+    if (loc.bearing) bearing = Number(loc.bearing.toFixed(LocationSlug.BearingFixed));
+    if (loc.pitch) pitch = Number(loc.pitch.toFixed(LocationSlug.PitchFixed));
+    return { bearing, pitch };
```
blacha, line 82: Could avoid the `let` with something like:

```suggestion
return {
 bearing: loc.bearing ? Number(loc.bearing.toFixed(LocationSlug.BearingFixed)) : undefined,
 pitch: loc.pitch ? Number(loc.pitch.toFixed(LocationSlug.PitchFixed)) : undefined
 }
```

blacha (approved):

PR #3199, opened 2024-03-20
feat(landing): exaggerate the terrain when viewing in NZTM BM-983

blacha:
#### Motivation

When viewing terrain in NZTM the terrain is too flat, this is caused by the NZTM projection hacks to allow Maplibre to render NZTM, it offsets the zoom level by 2 which means everything is approx 4x smaller.

#### Modification

When viewing in NZTM set the terrain to be approx 4x higher than what we use in 3857.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/landing/src/components/debug.tsx
+++ b/packages/landing/src/components/debug.tsx
@@ -284,7 +284,7 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
 
     const target = getTerrainForSource(sourceId, Config.map.tileMatrix.projection);
     // no changes
-    if ((currentTerrain?.source === sourceId, currentTerrain?.exaggeration === target.exaggeration)) return;
+    if (currentTerrain?.source === sourceId && currentTerrain?.exaggeration === target.exaggeration) return;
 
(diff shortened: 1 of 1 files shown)


packages/landing/src/components/debug.tsx
```diff
@@ -282,7 +282,9 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
       return;
     }
 
-    if (currentTerrain?.source === sourceId) return;
+    const target = getTerrainForSource(sourceId, Config.map.tileMatrix.projection);
+    // no changes
+    if ((currentTerrain?.source === sourceId, currentTerrain?.exaggeration === target.exaggeration)) return;
```
Wentao-Kuang, line 287: That's a new conditional check format that I never seen before. I suppose it works same as `&&`?

blacha: very weird, wonder how that happened!

Wentao-Kuang (approved):

PR #3202, opened 2024-03-21
feat(landing): add dropdown to configure a hillshade

blacha:
#### Motivation

Hillshades add another level of context to a DEM layer which can be useful when inspecting DEM/DSM sources

#### Modification

Adds a dropdown to the debug page to allow users to select a hillshade layer 

image

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/landing/src/components/debug.tsx
+++ b/packages/landing/src/components/debug.tsx
@@ -23,6 +23,37 @@ export interface DebugState {
   isCog?: boolean;
 }
 
+const HillshadeLayerId = 'debug-hillshade';
+
+interface DropDownContext {
+  /** Label for the drop down */
+  label: string;
+  /** callback for when the dropdown changes */
+  onChange: ChangeEventHandler<HTMLSelectElement>;
+  /** Current value */
+  value: string;
+  /** List of options */
+  options: string[];
+}
+function debugSourceDropdown(ctx: DropDownContext): ReactNode {
+  return (
+    <div className="debug__info">
+      <label className="debug__label">{ctx.label}</label>
+      <div className="debug__value">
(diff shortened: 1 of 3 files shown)


packages/landing/src/components/debug.tsx
```diff
@@ -23,6 +23,40 @@ export interface DebugState {
   isCog?: boolean;
 }
 
+/** Layer Id for the hillshade laer in the debug map */
```
Wentao-Kuang, line 26:
```suggestion
/** Layer Id for the hillshade layer in the debug map */
```

packages/landing/src/components/debug.tsx
```diff
@@ -270,6 +309,48 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
...
+    const existingSource = map.getSource(hillShadeSourceId);
+    if (existingSource == null) {
+      const source = map.getSource(sourceId);
+      if (source?.type !== 'raster-dem') {
+        // Source cannot be found, config is invalid
+        Config.map.setDebug('debug.hillshade', null);
+        return;
+      }
+
+      map.addSource(hillShadeSourceId, { ...source.serialize(), type: 'raster-dem', id: undefined });
```
Wentao-Kuang, line 342: Why do we got a `id: undefined`?

Wentao-Kuang (approved):

packages/landing/src/components/debug.tsx
blacha: we need to force the id to be undefined.

PR #3210, opened 2024-03-27
fix(cli): Fix config import to prepare changes for individual vector config.

Wentao-Kuang:
#### Motivation

Out config import are broken when prepare changes for individual vector tileset.

#### Modification

Remove the hard coded topographic logic and update to support all vector tilset inside changes.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cli/src/cli/config/action.import.ts
+++ b/packages/cli/src/cli/config/action.import.ts
@@ -171,7 +171,7 @@ export class CommandImport extends CommandLineAction {
 
       const hasChanges = await updater.reconcile();
       if (hasChanges) {
-        this.changes.push(config as BaseConfig);
+        this.changes.push(config);
         this.invalidations.push(updater.invalidatePath());
         const oldData = await updater.getOldData();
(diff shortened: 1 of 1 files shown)


packages/cli/src/cli/config/action.import.ts
```diff
@@ -330,15 +332,16 @@ export class CommandImport extends CommandLineAction {
     const vectorUpdate = [];
     const styleUpdate = [];
     for (const change of this.changes) {
-      if (change === 'ts_topographic') {
+      if (change.id.startsWith(ConfigPrefix.TileSet) && (change as ConfigTileSet).type === TileSetType.Vector) {
```
blacha, line 335: I think this could be `mem.TileSet.is(change)` which would do the type casting for us.

Wentao-Kuang: Good point, just updated.

blacha (approved):

PR #3214, opened 2024-04-02
feat(cli): Output the feature changes for the vector update

Wentao-Kuang:
#### Motivation

Check and Compare the `collections.json` between old and new vector data, and output the feature changes for the Pull request comments.

#### Modification
Generate addition feature changes from `import` cli which can be added into config pull requests comments.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cli/src/cli/config/config.diff.ts
+++ b/packages/cli/src/cli/config/config.diff.ts
@@ -1,8 +1,17 @@
-import { LogType } from '@basemaps/shared';
+import { ConfigTileSet } from '@basemaps/config';
+import { Epsg, StacCollection, StacLink } from '@basemaps/geo';
+import { fsa, LogType } from '@basemaps/shared';
 import c from 'ansi-colors';
 import diff from 'deep-diff';
 
 export const IgnoredProperties = new Set(['id', 'createdAt', 'updatedAt', 'year', 'resolution']);
+
+interface StacLinkLds extends StacLink {
+  'lds:id': string;
+  'lds:name': string;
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli/config/config.diff.ts
```diff
@@ -51,3 +60,100 @@ export class ConfigDiff {
...
+ * Prepare and create pull request for the aerial tileset config
+ */
+export async function diffVectorUpdate(
+  tileSet: ConfigTileSet,
+  existingTileSet: ConfigTileSet | null,
+): Promise<string | undefined> {
+  // Vector layer only support for 3857 and only contain on layer inside
+  const changes: string[] = [];
+  const layer = tileSet.layers[0];
+  if (layer[Epsg.Google.code] == null) return;
```
blacha, line 110: could layer be null?

```suggestion
  if (layer?.[Epsg.Google.code] == null) return;
```

Wentao-Kuang: I was thinking we might want to error here? Because we don't support other projections in vector map and we shouldn't have empty layers in a tileset

blacha: If we want the error we should throw, having "Cannot read property of undefined" is not super helpful

Wentao-Kuang: Just update this to thrown error.

blacha (approved):

PR #3241, opened 2024-04-21
fix(landing): ensure the layer dropdown always showing for all outputs. BM-1001

Wentao-Kuang:
#### Motivation

Outputs layers not showing up when using the `i` query parameter to viewing elevations data. Cause the `Config.map.styleId` will return the empty styleId. 

#### Modification

- Using layerId when `style` query parameter not specified. 
- Showing 'off' when there is no selectedSrouce for the layer dropdown.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/packages/landing/src/components/debug.tsx
+++ b/packages/landing/src/components/debug.tsx
@@ -425,8 +425,7 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
     const sourceIds = this.getSourcesIds('raster');
     if (sourceIds.length <= 1) return;
     // Get default source
-    let selectedSource = this.props.map.getLayer(Config.map.styleId)?.source;
-    if (selectedSource == null) selectedSource = 'off';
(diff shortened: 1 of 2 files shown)


packages/landing/src/components/debug.tsx
```diff
@@ -425,8 +425,8 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
     const sourceIds = this.getSourcesIds('raster');
     if (sourceIds.length <= 1) return;
     // Get default source
-    const selectedSource = this.props.map.getLayer(Config.map.styleId)?.source;
-    if (selectedSource == null) return;
+    let selectedSource = this.props.map.getLayer(Config.map.styleId)?.source;
+    if (selectedSource == null) selectedSource = 'off';
```
blacha, line 429:
```suggestion
```suggestion
    const selectedSource = this.props.map.getLayer(Config.map.styleId)?.source ?? 'off'; 
```
```

blacha (approved):

PR #3277, opened 2024-06-05
fix(cogify): Fix the elevation target path to include dem/dsm. BM-1040

Wentao-Kuang:
#### Motivation

The nz-elevation stac imagery name doesn't include the `dem_1m` prefix. We need to add it into basemaps imagery name and s3 path

#### Modification

Find the prefix from the nz-elevation source, and update the imageryName

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.cover.ts
+++ b/packages/cogify/src/cogify/cli/cli.cover.ts
@@ -16,6 +16,13 @@ import { createFileStats } from '../stac.js';
 
 const SupportedTileMatrix = [GoogleTms, Nztm2000QuadTms];
 
+// Round gsd to 3 decimals without trailing zero, or integer if larger than 1
+function gsdToMeter(gsd: number): number {
+  if (gsd > 1) return Math.round(gsd);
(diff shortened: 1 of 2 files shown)


packages/cogify/src/cogify/cli/cli.cover.ts
```diff
@@ -76,6 +76,14 @@ export const BasemapsCogifyCoverCommand = command({
 
     const res = await createTileCover(ctx);
 
+    // Find the dem/dsm prefix for the nz-elevation bucket source and update imagery name to include prefix
+    if (im.collection != null) {
+      const geoCategory = im.collection['linz:geospatial_category'];
+      if (geoCategory === 'dem' || geoCategory === 'dsm') {
+        im.name = `${im.name}_${geoCategory}_${im.gsd}m`;
```
blacha, line 83: `im.gsd` is a floating point number so you may see something like `1.00004` it would be good to trim it to `1m` or `0.1m` `0.75m`

blacha (approved):

PR #3292, opened 2024-06-19
feat(landing): Add terrain parameter in the url

Wentao-Kuang:
### Motivation

We want a terrain request parameter that shareable for map with 3d terrain enabled. This will need new aerial style config to added after merge. https://github.com/linz/basemaps-config/pull/916

### Modifications

- Add terrain setting in `getStyle` api when `?terrain=LINZ-Terrain`
- Update landing page to lisen the terrain event to set terrain parameter.

image

### Verification

- Unit tests added for tileset to style and style to style in getStyle Api
- Landing page tested with terrain button to update the request parameter.

### DIFF
--- a/packages/landing/src/components/map.tsx
+++ b/packages/landing/src/components/map.tsx
@@ -47,8 +47,11 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
 
   updateTerrainFromEvent = (): void => {
     const terrain = this.map.getTerrain();
-    if (terrain?.source) Config.map.setTerrain(terrain.source);
-    Config.map.setTerrain(terrain?.source ?? null);
+    if (terrain) {
+      Config.map.terrain = terrain.source;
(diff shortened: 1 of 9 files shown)


packages/landing/src/components/map.tsx
```diff
@@ -44,6 +45,16 @@ export class Basemaps extends Component<unknown, { isLayerSwitcherEnabled: boole
...
   };
 
+  updateTerrainFromEvent = (): void => {
+    const terrain = this.map.getTerrain();
+    if (terrain) {
+      Config.map.terrain = terrain.source;
+    } else {
+      Config.map.terrain = null;
+    }
+    window.history.pushState(null, '', `?${MapConfig.toUrl(Config.map)}`);
```
blacha, line 55: do we need to validate that this is actually a change? `Config.map.terrain != terrain.source`

blacha (approved):

PR #3299, opened 2024-06-26
feat(lambda-tiler): Ensure terrain source for all style json configs.

Wentao-Kuang:
### Motivation

We need to ensure a terrain source in all style json configs before the production release. Then we could do another config release for add terrain source into style config. https://github.com/linz/basemaps-config/pull/916

### Modifications

Check and add terrain source for get style config api.

### Verification

image

### DIFF
--- a/packages/lambda-tiler/src/routes/tile.style.json.ts
+++ b/packages/lambda-tiler/src/routes/tile.style.json.ts
@@ -268,6 +268,9 @@ export async function styleJsonGet(req: LambdaHttpRequest<StyleGet>): Promise<La
     styleConfig.style.layers.filter((f) => !excluded.has(f.id.toLowerCase())),
   );
 
+  // Ensure elevation for style json config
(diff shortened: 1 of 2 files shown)


packages/lambda-tiler/src/routes/tile.style.json.ts
```diff
@@ -268,6 +268,9 @@ export async function styleJsonGet(req: LambdaHttpRequest<StyleGet>): Promise<La
     styleConfig.style.layers.filter((f) => !excluded.has(f.id.toLowerCase())),
   );
 
+  // Ensure elevation for style json config
```
blacha, line 271: could we add a TODO to remove this and relate it to the configuration ticket that fixes it?

Wentao-Kuang: Fixed. Thank you.

blacha (approved):

PR #3302, opened 2024-07-01
fix(lambda-tiler): Fix the missing LINZ-terrain source for the elevation layer.

Wentao-Kuang:
### Motivation

Ensure terrain for the tileset that have outputs

### Modifications

Add terrain source for `tileSetOutputToStyle`.

### Verification

image


Wentao-Kuang:
I think we might need an additional check here, to see if this is elevation tileset, We just update the source name `basemaps-elevation-terrain-rgb` to `LINZ-Terrain` ?

blacha (approved):
I think we need to change the way we configure this, but for now this should fix the problem

PR #3307, opened 2024-07-02
fix(landing): Support Terrain for NZTM and add default LINZ-Terrain into debug

Wentao-Kuang:
### Motivation

We have default LINZ-Terrain source been added for all styleJson, this could be added into terrain drop down as well.
Also, we should support terrain for NZTM2000Quad tileMatrix

### Modifications
- Filter LINZ-Terrain into the terrain and hillshade dropdown
- Enable terrain controle and set terrain exaggeration with 4.4 for NZTM

### Verification
image

### DIFF
--- a/.github/workflows/screenshot.yml
+++ b/.github/workflows/screenshot.yml
@@ -40,7 +40,7 @@ jobs:
           # Wait for the server to start
           timeout 30 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:5000/v1/version)" !=  "200" ]]; do sleep 0.5; done' || false
 
-          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
+          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1.5.0 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
 
       - name: Save snapshots
         uses: getsentry/action-visual-snapshot@v2
--- a/.github/workflows/screenshot.yml
+++ b/.github/workflows/screenshot.yml
@@ -40,7 +40,7 @@ jobs:
           # Wait for the server to start
           timeout 30 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:5000/v1/version)" !=  "200" ]]; do sleep 0.5; done' || false
 
-          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1.5.0 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
+          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
 
       - name: Save snapshots
         uses: getsentry/action-visual-snapshot@v2
--- a/packages/config/src/config/vector.style.ts
+++ b/packages/config/src/config/vector.style.ts
@@ -4,6 +4,7 @@ import { ConfigBase } from './base.js';
 
 /**
  * Default Terrain exaggeration settings for different projection
+ * NZTM terrain is too flat, it offsets the zoom level by 2 which means everything is approx 4x smaller than Google.
  */
 export const DefaultExaggeration = {
   [Nztm2000QuadTms.identifier]: 4.4,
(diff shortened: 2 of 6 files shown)


packages/config/src/config/vector.style.ts
```diff
@@ -1,5 +1,15 @@
+import { GoogleTms, Nztm2000QuadTms } from '@basemaps/geo';
+
 import { ConfigBase } from './base.js';
 
+/**
+ * Default Terrain exaggeration settings for different projection
```
blacha, line 6: could we expand this to why they need the different numbers.

NZTM2000Quad is offset from WebMercatorQuad by 2 zoom levels so it needs to be approx 4x as large.

packages/landing/src/components/debug.tsx
```diff
@@ -423,7 +424,9 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
 
   getSourcesIds(type: string): string[] {
     const style = this.props.map.getStyle();
-    return Object.keys(style.sources).filter((id) => id.startsWith('basemaps') && style.sources[id].type === type);
+    return Object.keys(style.sources).filter(
+      (id) => (id.startsWith('basemaps') || id.startsWith('LINZ')) && style.sources[id].type === type,
```
blacha, line 428: I think we should do a exclude of the `__hillshade` rather than the include? as its only the duplicated hillshade layer we are trying to remove.

Wentao-Kuang: This is only filter the useful ones which might startwith `basemaps-` or `LINZ-`

Wentao-Kuang: We don't want others to be added to layer dropdown like `oms` source

blacha: what source is oms ? I could see us in the future removing the forced `basemaps-` prefix.

Wentao-Kuang: image
The layer slider will add open street map(`osm`) source and aerial source into the style, which might added to the layer dropdown if we don't filter them out.

packages/config/src/config/vector.style.ts
Wentao-Kuang: Fixed

blacha: This doesnt look like it has changed?

packages/landing/src/components/debug.tsx
blacha: These wouldn't be `raster-dem` sources though would they? I wonder if we should type the `type` ? or doc this function more

Wentao-Kuang: This function been used for Layer drop down as well. So the osm and aerial will showing up in the Layer drop down.

packages/config/src/config/vector.style.ts
Wentao-Kuang: My previous push failed, just pushed the changes.

packages/landing/src/components/debug.tsx
Wentao-Kuang: Hope that explained the problem for not filtering `basemaps-` and `LINZ-`
image

.github/workflows/screenshot.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           # Wait for the server to start
           timeout 30 bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:5000/v1/version)" !=  "200" ]]; do sleep 0.5; done' || false
 
-          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1.5.0 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
+          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
```
Wentao-Kuang, line 43:
```suggestion
          docker run --rm --network="host" -v $PWD:$PWD ghcr.io/linz/basemaps-screenshot/cli:v1.5.0 --url http://localhost:5000 --output $PWD/.artifacts/visual-snapshots
```

packages/landing/src/components/debug.tsx
```diff
@@ -455,7 +456,13 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
 
   getSourcesIds(type: string): string[] {
```
blacha, line 457: NIT: would be good to document what `type` is for the caller, eg `raster-dem` or `raster`

blacha (approved):

packages/landing/src/components/debug.tsx
Wentao-Kuang: Might just leave it, as the function is very simple to read .

blacha: One point of comments is so you don't have to read the function to understand what the argument is, especially with generic variable names like "type".

I also wonder about typing it as "raster | raster-dem" as no other "type" makes any sense.

Wentao-Kuang: typing it as "raster | raster-dem" sounds better option. I will hack that in next PR.

PR #3308, opened 2024-07-04
fix(cogify): Fix the dependencies for cogify cli.

Wentao-Kuang:
### Motivation

Cogify is bundled and installed in the @basemaps/cli container. They way calling cogify is from `node_modules`, like `node /app/node_modules/.bin/cogify`. We will require the the dependencies to be installed as well to make it work.

### Modifications
Revert back all the dependencies from devdependencies


Wentao-Kuang:
We use dist/index.cjs instead

PR #3314, opened 2024-07-10
feat(config): Update the config to support sky in the style json. BM-1052

Wentao-Kuang:
### Motivation

Update the config style parser and loader to include sky in the stylejson.

### Modifications

Add sky type into stylejson

### Verification

image
image

### DIFF
--- a/packages/config/src/config/vector.style.ts
+++ b/packages/config/src/config/vector.style.ts
@@ -40,6 +40,24 @@ export interface Layer {
   'source-layer'?: string;
 }
 
+/** Sky MapLibre Style Spec, all valuables support interpolate expressions, as unknown format */
+export interface Sky {
+  /** The base color for the sky. Optional Defaults to #88C6FC */
+  'sky-color'?: unknown;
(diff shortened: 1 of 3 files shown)


packages/config/src/config/vector.style.ts
```diff
@@ -40,6 +40,24 @@ export interface Layer {
   'source-layer'?: string;
 }
 
+/** Sky MapLibre Style Spec, all valuables support interpolate expressions, as unknown format */
+export interface Sky {
+  /** The base color for the sky. Optional Defaults to #88C6FC */
+  'sky-color'?: unknown;
```
blacha, line 46: since we know what these are meant to be could we type them properly? `string` `number` etc?

Wentao-Kuang: These all support interpolate expressions, we can use `string | unknown` or just `unknown`

blacha: `string | unknown[]` ?

Wentao-Kuang: Updated

blacha (approved):

PR #3317, opened 2024-07-11
ci: Enable 3d screenshots in the basemaps. BM-1054

Wentao-Kuang:
### Motivation

We want to fix the 3d screenshots in PR.

### Modifications
- When setting 3d in screenshots, we want to force load config from url and wait for 2s.
- Updated to the latest screenshot cli which removes bearing and pitch to save computing.
- Only do 5 tests on 3d map screenshots to keep the workflow running around 4 mins.
- Stop showing `debug.terrain` and `debug.terrain` in url when it is `off`

### DIFF
--- a/packages/landing/src/components/debug.tsx
+++ b/packages/landing/src/components/debug.tsx
@@ -121,8 +121,10 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
           loadedDiv.style.height = '1px';
           document.body.appendChild(loadedDiv);
         }
-        void map.on('idle', () => {
-          void addLoadedDiv();
+        void map.on('sourcedata', (e) => {
+          if (e.isSourceLoaded) {
+            void addLoadedDiv();
+          }
         });
       }
     });
--- a/packages/landing/src/components/debug.tsx
+++ b/packages/landing/src/components/debug.tsx
@@ -122,7 +122,7 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
           document.body.appendChild(loadedDiv);
         }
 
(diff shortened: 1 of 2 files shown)


packages/landing/src/components/debug.tsx
```diff
@@ -117,7 +121,22 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
...
+        }
+
+        void map.on('data', () => {
+          // Remove the map-loaded div if exists
+          void removeLoadedDiv();
+        });
+
+        void map.on('idle', () => {
+          // Ensure hillshade and terrain source is loaded
+          this.updateFromConfig();
```
Wentao-Kuang, line 139: Do and force update and wait for 2s only for terrain screenshots.

packages/landing/src/components/debug.tsx
```diff
@@ -383,10 +390,13 @@ export class Debug extends Component<{ map: maplibregl.Map }, DebugState> {
     const isTurnOff = sourceId === 'off' || sourceId == null;
     const currentTerrain = map.getTerrain();
     if (isTurnOff) {
+      Config.map.setDebug('debug.terrain', null);
```
blacha, line 393: should L387 be removed if this is done here

Wentao-Kuang: I think we should keep this, it doesn't make sense to` debug.terrain=off` when selecting `off` in dropdown, we just set it as `null` should same meaning.
Also, if we don't set null here, line 124 also need to check `null` and `off`.

blacha (approved):

PR #3326, opened 2024-08-14
fix(landing): Fix the map switcher that setting bearing and pitch to 0. BM-1063

Wentao-Kuang:
### Motivation
Landing page map switcher on the bottom left doesn't set the map location when bearing or pitch is 0.

### Modifications
The map switcher sets the bearing and pitch by the update location from config. But it is skip setting when pitch or bearing is null. We should also set to 0 if null.

### DIFF
--- a/packages/landing/src/components/map.switcher.tsx
+++ b/packages/landing/src/components/map.switcher.tsx
@@ -78,8 +78,8 @@ export class MapSwitcher extends Component {
 
     this.map.setZoom(Math.max(location.zoom - 4, 0));
     this.map.setCenter([location.lon, location.lat]);
-    if (Config.map.location.bearing != null) this.map.setBearing(Config.map.location.bearing);
-    if (Config.map.location.pitch != null) this.map.setPitch(Config.map.location.pitch);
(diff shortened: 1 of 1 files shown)


packages/landing/src/components/map.switcher.tsx
```diff
@@ -78,8 +78,8 @@ export class MapSwitcher extends Component {
 
     this.map.setZoom(Math.max(location.zoom - 4, 0));
     this.map.setCenter([location.lon, location.lat]);
-    if (Config.map.location.bearing != null) this.map.setBearing(Config.map.location.bearing);
-    if (Config.map.location.pitch != null) this.map.setPitch(Config.map.location.pitch);
+    Config.map.location.bearing == null ? this.map.setBearing(0) : this.map.setBearing(Config.map.location.bearing);
```
blacha, line 81:
```suggestion
    this.map.setBearing(Config.map.location.bearing ?? 0);
```

Wentao-Kuang: Haha, nice coding Minimalism

blacha (approved):

PR #3329, opened 2024-08-21
fix(lambda-tiler): prevent unhandled promise rejections when the rejection is handled BM-1067

blacha:
### Motivation

Fetching data from remote services can and will error on occasion, the promises to fetch data from tiffs has two `.then()` one of the `.then()` has a associated `.catch()` the other does not, this leads to a `UnhandledPromiseRejection` 

### Modifications

Do not use a `.then()` without `.catch()`

### Verification

Unit test addded.
also checked the code base for more usages of `void ....then()` without .catch and they only occur in the landing page.


Wentao-Kuang:
I didn't see any unit tests added based on your PR description. Might forgot to upload?

blacha:
> I didn't see any unit tests added based on your PR description. Might forgot to upload?

Your quite right, I missed the test file.

Wentao-Kuang (approved):

PR #3332, opened 2024-08-26
fix(infra): reduce cache keys below 10 to prevent deployment errors

blacha:
### Motivation

Deployments are currently failing due to a AWS limit of 10 cache keys per behaviour

### Modifications

Remove deprecated cache keys

### Verification

Will see if it fixes the deployment

### DIFF
--- a/packages/_infra/src/edge/index.ts
+++ b/packages/_infra/src/edge/index.ts
@@ -97,36 +97,30 @@ export class EdgeStack extends cdk.Stack {
         originProtocolPolicy: cf.OriginProtocolPolicy.HTTPS_ONLY,
       },
       behaviors: [
+        // Configuration for all lambda requests
+        // Only the tiles themselves are cached so only things affecting tile rendering should be a cacheKey
         {
           pathPattern: '/v1*',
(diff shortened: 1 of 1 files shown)


packages/_infra/src/edge/index.ts
```diff
@@ -97,36 +97,30 @@ export class EdgeStack extends cdk.Stack {
...
-              'pipeline',
-              'terrain',
-              'labels',
-              // Deprecated single character query params for style and projection
-              's',
-              'p',
-              'i', // ?i=:imageryId is deprecated and should be removed at some point
-              't',
-            ].map(encodeURIComponent),
+            // queryStringCacheKeys are limited to a max of 10
```
Wentao-Kuang, line 120: We don't want `lables` anymore?

blacha: its a cache key, so its only for things that get cached, anything with a API key (eg style.json) has a `cache: no-store` directive.

Wentao-Kuang: If so, probably we want to remove `terrain` as well. As it is only for style.json

Wentao-Kuang (approved):

PR #3341, opened 2024-09-16
fix(landing): hide label button on debug page as its broken BM-20

blacha:
### Motivation

the labels view requires the "labels" style to exist, this generally does not exist inside of custom configs.

### Modifications

Hide the label button when custom configs on the debug page.

### Verification

Manually tested with current debug page.

### DIFF
--- a/packages/landing/src/components/map.label.tsx
+++ b/packages/landing/src/components/map.label.tsx
@@ -47,12 +47,22 @@ export class MapLabelControl implements IControl {
     Config.map.setLabels(labelState);
   };
 
+  /** Is the label button hidden from view */
+  isDisabled(): boolean {
+    // Most vector styles have overlaps with the labels and make them useless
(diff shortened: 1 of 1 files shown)


packages/landing/src/components/map.label.tsx
```diff
@@ -47,12 +47,22 @@ export class MapLabelControl implements IControl {
     Config.map.setLabels(labelState);
   };
 
+  /** Is the label button hidden from view */
+  isDisabled(): boolean {
+    // Most vector styles have overlaps with the labels and make them useless
+    if (Config.map.style && LabelsDisabledLayers.has(Config.map.style)) return true;
+    // Labels use the merge style feature and need the production configuration to work
+    if (Config.map.isDebug && Config.map.config) return true;
```
Wentao-Kuang, line 55: NIL, `Config.map.config` this could point to production config as well. But it's ok.

Wentao-Kuang (approved):

PR #3346, opened 2024-09-23
feat(landing): store the maps bounds to provide a better bounding box intersection

blacha:
### Motivation

The filter layer intersection logic is currently guessing the extent that the user is viewing

### Modifications

Store the extent that the user is looking at when the map pans

### Verification

tested locally

### DIFF
--- a/packages/landing/src/components/layer.switcher.dropdown.tsx
+++ b/packages/landing/src/components/layer.switcher.dropdown.tsx
@@ -1,4 +1,3 @@
-import { Bounds, GoogleTms, Projection } from '@basemaps/geo';
 import { intersection, MultiPolygon, truncate, Wgs84 } from '@linzjs/geojson';
 import { ChangeEventHandler, Component, ReactNode } from 'react';
 import Select from 'react-select';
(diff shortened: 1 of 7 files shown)


packages/landing/src/components/layer.switcher.dropdown.tsx
```diff
@@ -228,25 +230,33 @@ export class LayerSwitcherDropdown extends Component<unknown, LayerSwitcherDropd
...
+
+  const inter = intersection(bounds, poly);
+  if (inter == null || inter.length === 0) return false;
+
+  // No attribution state loaded, assume it intersects
+  const attrs = MapAttrState.attrsSync.get('all');
+  if (attrs == null) return true;
 
-  const layerBounds = Bounds.fromBbox([ul3857[0], ul3857[1], lr3857[0], lr3857[1]]);
+  const attrLayer = attrs.attributions.filter((f) => f.collection.title === layer.title);
```
blacha, line 257: I am not a huge fan of using the title to find a match between the attribution.json and the WMTSCapabilities.xml

I would prefer we add the `layer.name` to attribution.json and/or the `layer.id` to the WMTS

I am leaning towards layer.name to attribution.json, but as its a public API I would like to do that as a seperate PR

Wentao-Kuang:
Looks all good to me. Just wondering could we use capture-aerial multiploygon for intersection instead of the extent bounding box?

Wentao-Kuang (approved):

PR #3348, opened 2024-09-26
fix(server): Add lerc external into server package.

Wentao-Kuang:
### Motivation
It looks like we are missing lerc in the server cli. 
https://github.com/linz/basemaps-config/actions/runs/11061552896/job/30734619452?pr=970
image

### Modifications
Add the `lerc` into server and cli package

### Verification

Local build server container with lerc uninstall vs installed.

#### UNINSTALL
image

#### INSTALLED
image

#### Test Screenshot with fixed container
image

### DIFF
--- a/packages/cli/Dockerfile
+++ b/packages/cli/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update
 RUN apt-get install -y nodejs
 
 # Install sharp TODO update this when we change sharp versions
-RUN npm install sharp@0.33.0 lerc@^4.0.4
+RUN npm install sharp@0.33.0
 
 # Install the landing assets
 COPY ./basemaps-landing*.tgz /app/
--- a/packages/cli/Dockerfile
+++ b/packages/cli/Dockerfile
@@ -15,7 +15,7 @@ RUN apt-get update
 RUN apt-get install -y nodejs
 
 # Install sharp TODO update this when we change sharp versions
-RUN npm install sharp@0.33.0
+RUN npm install sharp@0.33.0 lerc@^4.0.4
 
 # Install the landing assets
 COPY ./basemaps-landing*.tgz /app/
--- a/packages/cli/src/cli/server/action.serve.ts
+++ b/packages/cli/src/cli/server/action.serve.ts
@@ -9,7 +9,7 @@ import {
 
 const DefaultPort = 5000;
 
-//TO DEPRECATE: We should use the server container directly now, and deprecate this later.
+/** @deprecated 2024-09 Use the basemaps/server container directly now, and deprecate this later. */
 export class CommandServe extends CommandLineAction {
   config!: CommandLineStringParameter;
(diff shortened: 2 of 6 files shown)


packages/server/Dockerfile
```diff
@@ -12,6 +12,8 @@ WORKDIR /app/basemaps
 COPY ./basemaps-landing*.tgz /app/basemaps/
 COPY ./basemaps-server*.tgz /app/basemaps/
 
+RUN npm install lerc@^4.0.4
```
blacha, line 15: why is this needed here? if we were missing binary deps then sharp should be missing too?

packages/cli/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update
 RUN apt-get install -y nodejs
 
 # Install sharp TODO update this when we change sharp versions
-RUN npm install sharp@0.33.0
+RUN npm install sharp@0.33.0 lerc@^4.0.4
```
blacha, line 18: where in the CLI is this needed?

Wentao-Kuang: We use this cli container for the server in [basemaps-config](https://github.com/linz/basemaps-config/blob/master/scripts/bms.sh). So we need to install lerc this into cli container.

packages/server/Dockerfile
Wentao-Kuang: I didn't see an error related sharp when I test the server container with screenshots. Might be the deps for `cogify` in the `cli` container? But we definitely need to install `lerc` in here to make it work.

packages/cli/Dockerfile
blacha: if the server is being installed it should come with `lerc` so I think the server is just missing the lerc dep? how is `sharp` defined in basemaps/server ?

Wentao-Kuang: `sharp` is installed as dependency already in the [basemaps/server](https://github.com/linz/basemaps/blob/master/packages/server/package.json#L50)

But we are not installing `sharp` in the basemaps/cli, so we installed into container directly.

Wentao-Kuang: I don't think server package is installed in the container, it is just a [function call](https://github.com/linz/basemaps/blob/master/packages/cli/src/cli/server/action.serve.ts) to the `package/server`, so we still need to install `lerc` in the cli container.

packages/cli/src/cli/server/action.serve.ts
```diff
@@ -9,6 +9,7 @@ import {
 
 const DefaultPort = 5000;
 
+//TO DEPRECATE: We should use the server container directly now, and deprecate this later.
```
blacha, line 12:
```suggestion
/** @deprecated 2024-09 Use the basemaps/server container directly now, and deprecate this later. */
```

blacha (approved):

PR #3358, opened 2024-10-16
feat(doc): Add exclude query parameter for get style API into the documentation. BM-1093

Wentao-Kuang:
### Motivation

exclude query parameter in the getStyle API should expose to the customer facing. So we should add this into basemaps doc as an example

### Modifications

Add exclude parameter into Swagger API example.

### Verification
image


Wentao-Kuang:
This is not required.

PR #3359, opened 2024-10-17
fix(lambda-tiler): correctly log fetch requests

blacha:
### Motivation

Requests tracing has been broken for a while (not entire sure when it broke)

It is very very helpful to have statistics on the number of s3 requests required to make a tile

### Modifications

- `requestCount` is reserved for the number of lambda invocations so move to `fetchCount` for number of fetches
- include the fetchIds to trace hot fetches
- increase the percent of requests with tracing enabled
- Use async local storage to gain access to the logger for the request that triggers the log message, this fixes a bug where the request may be in `trace` but the fetch is using the base logger which is set to `info` so no logs are generated.

### Verification

Ran locally and verified logs are now appearing.

### DIFF
--- a/packages/lambda-tiler/src/index.ts
+++ b/packages/lambda-tiler/src/index.ts
@@ -1,4 +1,4 @@
-import { FsaCache, FsaLog, LogConfig } from '@basemaps/shared';
+import { FsaCache, FsaLog, LogConfig, LogStorage } from '@basemaps/shared';
 import { LambdaHttpRequest, LambdaHttpResponse, lf } from '@linzjs/lambda';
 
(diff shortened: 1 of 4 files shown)


packages/lambda-tiler/src/index.ts
```diff
@@ -35,12 +35,13 @@ function randomTrace(req: LambdaHttpRequest): void {
   // 1% trace
   if (rand < 0.01) req.log.level = 'trace';
   // 5% debug
```
Wentao-Kuang, line 37:
```suggestion
  // 25% debug
```

Wentao-Kuang (approved):

Wentao-Kuang (approved):

PR #3377, opened 2024-12-05
feat(config): Add resizeKernel into the config tileset then disable fastShrinkOnLoad. BM-1146

Wentao-Kuang:
### Motivation
As raster Topo map got edge effects on low zoom levels usually below zoom 6, like following. 
image

We have tests that with output resize kernel of `mitchell` and disable the fastShrinkOnLoad for lower scales levels will fix this problem. 
image

### Modifications

- Add the resiseKernel setting in the config bundle, so that we can bundle it into server.
- disable `fastShrinkOnLoad` for the `resize.scale`less than 0.125.

- Not fix yet: The virtual tileset for `topo-raster-50k` and `topo-raster-250k` won't fix as we can't propagate the resize kernel from combined tileset -> imagery config -> virtual tileset. Unless we add optional resizeKernal property for imagery config. 

### Verification
Local test.
image

### DIFF
--- a/packages/tiler-sharp/src/index.ts
+++ b/packages/tiler-sharp/src/index.ts
@@ -145,6 +145,7 @@ export class TileMakerSharp implements TileMaker {
     if (resize) {
       const resizeOptions = { fit: Sharp.fit.cover, kernel: resize.scaleX > 1 ? resizeKernel.in : resizeKernel.out };
       if (resize.scale <= 0.1251) {
+        // If the scale is less than 0.125, we need to disable fastShrinkOnLoad to prevent edge artifacts for topo raster map
         sharp.resize(resize.width, resize.height, { ...resizeOptions, fastShrinkOnLoad: false });
       } else {
         sharp.resize(resize.width, resize.height, resizeOptions);
--- a/packages/tiler-sharp/src/index.ts
+++ b/packages/tiler-sharp/src/index.ts
@@ -144,7 +144,11 @@ export class TileMakerSharp implements TileMaker {
 
     if (resize) {
       const resizeOptions = { fit: Sharp.fit.cover, kernel: resize.scaleX > 1 ? resizeKernel.in : resizeKernel.out };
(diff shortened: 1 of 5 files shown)


packages/tiler-sharp/src/index.ts
```diff
@@ -144,7 +144,12 @@ export class TileMakerSharp implements TileMaker {
 
     if (resize) {
       const resizeOptions = { fit: Sharp.fit.cover, kernel: resize.scaleX > 1 ? resizeKernel.in : resizeKernel.out };
-      sharp.resize(resize.width, resize.height, resizeOptions);
+      if (resize.scale <= 0.1251) {
```
blacha, line 147: I think we need to link to the ticket for this, and maybe also expand on why the `.1` is used and why `0.125` (aka 1/8th)

packages/tiler/src/raster.ts
```diff
@@ -7,7 +7,7 @@ export interface TileMaker {
   compose(ctx: TileMakerContext): Promise<{ buffer: Buffer; metrics: Metrics }>;
 }
 
-export type ResizeKernelType = 'nearest' | 'lanczos3' | 'lanczos2';
+export type ResizeKernelType = 'nearest' | 'mitchell' | 'lanczos3' | 'lanczos2';
```
blacha, line 10: we have this type in about 4 places any chance we could merge them somehow?

Wentao-Kuang: Yeah, we was question about this a lot when doing the change. Might just use the one in config package

Wentao-Kuang: Turns out we don't install config package in the tiler package. :rofl:

blacha: lets just do this later.

packages/tiler-sharp/src/index.ts
blacha: we can do this later.

PR #3378, opened 2024-12-10
feat(cli): Build test pr container for cli package.

Wentao-Kuang:


Wentao-Kuang:
Close this, it is not safe to create pr container in public repo.

Issue #3365, opened 2024-10-29
Request: raster format support for XYZX topographic tile services


mwtoews:
### User Story

Currently the XYZ topographic tile map service supports vector tiles with `.pbf` (protobuf) format. This service does not work for raster-based XYZ services, e.g. [xyzservices](https://xyzservices.readthedocs.io/).

This request is to support raster format support, similar to other services like aerial.

This service could be implemented by specifying an extension:

    https://basemaps.linz.govt.nz/v1/tiles/topographic/{crs}/{z}/{x}/{y}.{ext}?api={api_key}

where `{ext}` could be `pbf` for currently supported vector protobuf vector tiles, or one of `png`, `jpeg` or `webp` for a possible raster format.

This feature would require server-side rendering of the tiles.

JBLINZ:
Hey.... Thanks for the user story. 

We will likely establish a raster-based topo tile service in the next few months. The service will use the topo 1:50 and 1:250 raster maps rather than the vector tiles being rasterised. 

We may look at vector tiles being rasterised in the future.

mwtoews:
Personally, I would prefer the published-looking 1:50k and 1:250k raster topo map services. (I've been using the WMTS versions of these from data.linz.govt.nz and they work flawlessly!)

Thus rasterized versions of the vector tiles would take a lower priority, as this would be a more complex feature to entertain.

JBLINZ:
Excellent. We will let you know when a test service is available.

mwtoews:
Fantastic! And, for the record, this is somewhat already done with Koordinates via ([e.g.](https://data.linz.govt.nz/layer/50767-nz-topo50-maps/services/tiles/)):

    https://tiles-cdn.koordinates.com/services;key=[api_token]/tiles/v4/layer=50767/EPSG:3857/{z}/{x}/{y}.png

JBLINZ:
Understood. We see value in running the service via basemaps.linz.govt.nz too. We aim to publish 2 individual layers (50 & 250), and a combined service switching between the 2 at appropriate zoom levels.

JBLINZ:
[...]

PR #3388, opened 2025-01-08
feat(cogify): Update cogify to support the topo raster processes. BM-1116

Wentao-Kuang:
### Motivation

As a Basemaps user, I want to consume the NZTopo 50 & 250 Maps [as a tile service].

### Modifications

This work comprises a series of additions and modifications to the `basemaps/cogify` package.

These changes allow us to generate [STAC] files for an NZTopo Map Series imagery collection. They also allow us to generate standardised and cloud-optimised GeoTIFFs from the map sheet files of such collections. We have adjusted the `basemaps/cogify` package to handle the following collections stored in the AWS S3 [TopoReleaseArchive] directory:

```
s3://topographic-upload/TopoReleaseArchive/NZTopo50_GeoTif_Gridless/
s3://topographic-upload/TopoReleaseArchive/NZTopo250_GeoTif_Gridless/
``` 

There is also a [new Argo workflow in development][pr-argo-workflow] that depends on this work as part of its process flow. The workflow's purpose is to automate the standardisation of an NZTopo Map Series imagery collection.

### Generating STAC files

This works defines a new process for generating a nested directory structure of STAC files for an NZTopo Map Series imagery collection. The process operates as follows:

1. Loads the collection of GeoTiff images into memory
6. Extracts the map code, version, and metadata (e.g. EPSG) from each image
7. Identifies the latest version of each map sheet by code
8. Groups the images by EPSG
9. Generates a StacItem file for each image, and a StacCollection file for each EPSG grouping
10. Structures the groups of StacItem and StacCollection files into a directory tree
11. Saves the files as they are structured to a target location

#### Outputs

The command groups the images by EPSG and then structures the generated StacItem and StacCollection files, as illustrated:

| Schema | Example |
| - | - |
| ![][schema] | ![][example] |

The command then saves the generated tree of folders and files into the target location directory.

### Processing Map Sheet files

This work extends the `create` CLI command of the `basemaps/cogify` package. This extension allows the command to recognise and process the map sheet files of an NZTopo Map Series imagery collection by way of particular properties added to the accompanying STAC files.

[as a tile service]: https://github.com/linz/basemaps/issues/3365#issuecomment-2445444420
[STAC]: https://stacspec.org/en
[TopoReleaseArchive]: https://linz-topographic-upload.s3.ap-southeast-2.amazonaws.com/topographic/TopoReleaseArchive/

[pr-argo-workflow]: https://github.com/linz/topo-workflows/pull/932

[schema]: https://github.com/user-attachments/assets/2f4eddd7-eb2c-4796-8ab7-b3d9eca9140f
[example]: https://github.com/user-attachments/assets/8f111ba6-befd-44b5-9566-db07a07089ed


packages/cogify/src/cogify/cli/cli.cover.ts
```diff
@@ -84,6 +84,10 @@ export const BasemapsCogifyCoverCommand = command({
       throw new Error(`No collection.json found with imagery: ${im.url.href}`);
     }
 
+    if (im.collection && im.collection['linz:slug']) {
+      im.name = im.collection['linz:slug'] as string;
+    }
```
blacha, line 89:
```suggestion
const slug = im.collection?.['linz:slug']
if (slug != null) name = slug
```

packages/cogify/src/cogify/gdal.command.ts
```diff
@@ -53,45 +86,38 @@ export function gdalBuildVrtWarp(
...
-  const targetResolution = tileMatrix.pixelScale(cfg.zoomLevel);
+  const targetOpts = getTargetOptions(cfg);
 
   return {
     command: 'gdal_translate',
     output: targetTiff,
     args: [
       ['-of', 'COG'],
+      cfg.srcwin ? ['-srcwin', cfg.srcwin[0], cfg.srcwin[1], cfg.srcwin[2], cfg.srcwin[3]] : undefined,
+      cfg.bigTIFF ? ['-co', `BIGTIFF=${cfg.bigTIFF}`] : ['-co', 'BIGTIFF=IF_NEEDED'], // BigTiff is somewhat slower and most (All?) of the COGS should be well below 4GB
```
blacha, line 98: why do we need bigtiffs? everything should be well below 4gb?

Wentao-Kuang: We set big tiff as `no` in the `topo-raster`, and `If_needed` for all others atm.

packages/cogify/src/cogify/cli/cli.topo.ts
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * @param target: Destination directory URL into which to save the STAC collection and item JSON files
+ * @example TODO
+ *
+ * @param title: The title of the collection
+ * @example "New Zealand Topo50 Map Series (Gridless)"
+ *
+ * @returns an array of StacItem objects
+ */
+async function loadTiffsToCreateStacs(
```
blacha, line 118: this is quite a huge function and would be very easy to mix the ordering of args it would be best to use a context object here.

packages/cogify/src/cogify/cli/cli.topo.ts
```diff
@@ -0,0 +1,207 @@
...
+      await fsa.write(new URL('brokenTiffs.json', targetURL), JSON.stringify(brokenTiffs, null, 2));
+    }
+
+    logger.info({ duration: performance.now() - startTime }, 'ListJobs:Done');
+  },
+});
+
+/**
+ * @param source: Source directory URL from which to load tiff files
+ * @example TODO
```
blacha, line 108: TODO? would be best to just remove this comment?

packages/cogify/src/cogify/cli/cli.topo.ts
```diff
@@ -0,0 +1,207 @@
...
+      args.resolution,
+      args.forceOutput,
+      logger,
+    );
+
+    if (epsgDirectoryPaths.length === 0 || stacItemPaths.length === 0) throw new Error('No Stac items created');
+
+    // write stac items into an JSON array
+    if (args.forceOutput || isArgo()) {
+      const targetURL = isArgo() ? fsa.toUrl('/tmp/topo-stac-creation/') : args.target;
```
blacha, line 91: generally its best to not upper case variables `targetURL -> targetUrl` 

As you get into weird screaming words when you join two or more together `HTMLURL`

packages/cogify/src/cogify/cli/cli.topo.ts
```diff
@@ -0,0 +1,207 @@
...
+
+      // for create-config: we need to tell create-config to create a bundled config for each epsg folder (latest only).
+      // workflow: will loop 'targets.json' and create a node for each path where each node's job is to create a bundled config.
+      await fsa.write(new URL('targets.json', targetURL), JSON.stringify(epsgDirectoryPaths, null, 2));
+
+      // tiles.json makes the tiff files
+      await fsa.write(new URL('tiles.json', targetURL), JSON.stringify(stacItemPaths, null, 2));
+      await fsa.write(new URL('brokenTiffs.json', targetURL), JSON.stringify(brokenTiffs, null, 2));
```
blacha, line 99: do any of the other outputs use `camelCasing.json` ?

packages/cogify/src/cogify/gdal/gdal.command.ts
```diff
@@ -2,18 +2,22 @@ import { Rgba } from '@basemaps/config';
...
 import { urlToString } from '@basemaps/shared';
 
-import { Presets } from '../preset.js';
+import { Presets } from '../../preset.js';
+import { CogifyCreationOptions } from '../stac.js';
 import { GdalCommand } from './gdal.runner.js';
-import { CogifyCreationOptions } from './stac.js';
 
 const isPowerOfTwo = (x: number): boolean => (x & (x - 1)) === 0;
+const DEFAULT_TRIM_PIXEL_RIGHT = 1.7; // 1.7 pixels to trim from the right side of the topo raster imagery
```
blacha, line 10: we generally use `TitleCase` here, I wonder why the linter is not picking this up?

packages/cogify/src/cogify/gdal/gdal.command.ts
```diff
@@ -2,18 +2,22 @@ import { Rgba } from '@basemaps/config';
...
 
-export function gdalBuildVrt(targetVrt: URL, source: URL[]): GdalCommand {
+export function gdalBuildVrt(targetVrt: URL, source: URL[], addalpha?: boolean): GdalCommand {
   if (source.length === 0) throw new Error('No source files given for :' + targetVrt.href);
   return {
     output: targetVrt,
     command: 'gdalbuildvrt',
-    args: [urlToString(targetVrt), ...source.map(urlToString)],
+    args: [addalpha ? ['-addalpha'] : undefined, urlToString(targetVrt), ...source.map(urlToString)]
+      .filter((f) => f != null)
```
blacha, line 18: I wonder if `flatMap` would work here

```
.flatMap(f => f != null ? String(f): [])


[['foo'], undefined, ['bar'], [undefined, 'baz']].flatMap(f => f != null ? String(f): [])
[ 'foo', 'bar', ',baz' ]
```

packages/cogify/src/cogify/stac.ts
```diff
@@ -62,37 +63,88 @@ export interface CogifyCreationOptions {
...
+     * An EpsgCode Enum representing a map sheet's projection.
+     *
+     * @example EpsgCode.Nztm2000 = 2193
+     */
+    'proj:epsg': EpsgCode;
+
+    /**
+     * The width of a map sheet in pixels.
+     */
+    'source.width': number;
```
blacha, line 130: `.` inside json objects property names can cause some tools to have lots of problems, some tools use it to expect object nesting `{ source: { width: 100 } }` could these be moved or renamed?

We should also be prefixing keys that we are adding eg `linz_basemaps:map_code`

packages/cogify/src/cogify/topo/extract.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // try to extract the epsg directly from the tiff
+  const epsg = img.epsg;
+
+  if (epsg != null) {
+    const code = Epsg.tryGet(epsg);
+
+    if (code != null) {
+      logger?.info({ found: true, method: 'direct' }, 'extractEpsgFromTiff()');
```
blacha, line 54: would be good to log which tiff, `tiff.source.url.href` in these log messages.

packages/cogify/src/cogify/topo/extract.ts
```diff
@@ -0,0 +1,121 @@
...
+ * @returns a Bounds object, on success. Otherwise, null.
+ */
+export function extractBoundsFromTiff(tiff: Tiff): Bounds | null {
+  const img = tiff.images[0];
+  if (img == null) {
+    throw new Error(`No images found in Tiff file: ${tiff.source.url.href}`);
+  }
+
+  if (img.valueGeo(TiffTagGeo.GTRasterTypeGeoKey) === RasterTypeKey.PixelIsPoint) {
+    throw new Error("'Pixel is Point' raster grid spacing is not supported");
```
blacha, line 22: it is a good idea to throw with the tiff location so its easier to track broken tiffs.

packages/cogify/src/cogify/topo/extract.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param tiff: the Tiff object from which to extract the size.
+ *
+ * @returns a Size object, on success. Otherwise, null.
+ */
+export function extractSizeFromTiff(tiff: Tiff, logger?: LogType): Size | null {
+  try {
+    const size = tiff.images[0]?.size ?? null;
+
+    logger?.info({ found: size }, 'extractSizeFromTiff()');
```
blacha, line 115: mixing booleans and numbers will cause issues with logging.

packages/cogify/src/cogify/topo/mapper.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import {
+  extractBoundsFromTiff,
+  extractEpsgFromTiff,
+  extractMapCodeAndVersion,
+  extractSizeFromTiff,
+} from './extract.js';
+import { brokenTiffs, ByDirectory, TiffItem } from './types.js';
+
+const slugs: { [key in EpsgCode]?: string } = {
```
blacha, line 13:
```suggestion
const slugs: Partial<Record<EpsgCode, string>> = {
```

would this work?

packages/cogify/src/cogify/topo/mapper.ts
```diff
@@ -0,0 +1,101 @@
...
+export function groupTiffsByDirectory(tiffs: Tiff[], logger?: LogType): ByDirectory<TiffItem> {
+  // group the tiffs by directory, epsg, and map code
+  const byDirectory = new ByDirectory<TiffItem>();
+
+  // create items for each tiff and store them into 'all' by {epsg} and {map code}
+  for (const tiff of tiffs) {
+    const source = tiff.source.url;
+    const { mapCode, version } = extractMapCodeAndVersion(source.href, logger);
+
+    const bounds = extractBoundsFromTiff(tiff);
```
blacha, line 51: if one of these throws its going to break the loop early, we should process all items and collect all errors before exiting.

packages/cogify/src/cogify/topo/stac.creation.ts
```diff
@@ -0,0 +1,263 @@
...
+  logger?: LogType,
+): { all: TopoStacItem[]; latest: TopoStacItem } {
+  const allStacItems = all.map((item) =>
+    createBaseStacItem(`${item.mapCode}_${item.version}`, item, tileMatrix, logger),
+  );
+
+  const latestURL = new URL(`${latest.mapCode}_${latest.version}.json`, allTargetURL);
+
+  // add link to all items pointing to the latest version
+  allStacItems.forEach((stacItem) => {
```
blacha, line 101: NIT: we generally perffer `for const (item of allStacItems)` over `.forEach`

packages/cogify/src/cogify/topo/stac.creation.ts
```diff
@@ -0,0 +1,263 @@
...
+ *
+ * @param tiffs: the list of Tiff objects to group by epsg, and map code, and version
+ * @returns a `ByDirectory<TiffItem>` promise
+ */
+export function groupTiffsByDirectory(tiffs: Tiff[], logger?: LogType): ByDirectory<TiffItem> {
+  // group the tiffs by directory, epsg, and map code
+  const byDirectory = new ByDirectory<TiffItem>();
+
+  // create items for each tiff and store them into 'all' by {epsg} and {map code}
+  for (const tiff of tiffs) {
```
blacha, line 184: this appears to be duplicated above and should be simplified.

packages/cogify/src/cogify/gdal/gdal.command.ts
Wentao-Kuang: This only works with one value in the nested array, and not working with others with multiple values like `[ ['-of', 'vrt'], ]`

PR #3391, opened 2025-01-21
feat(tiler-sharp): support uint32 and uint8 source datasets for color-ramp

blacha:
### Motivation

We are getting datasets that have other datatypes than float32 that need to be expanded into 4 band RGBA so basemaps can render them.

A common type is a one band `uint8` which is generally a grey scale dataset which can be directly band expanded from `0` to `r:0, g:0, b:0, alpha: 255` and `255` to `r:255, g:255, b:255, alpha: 255`

### Modifications

Added support for `uint8` and `uint32` with the tiler piplines

TODO: it would be nicer to handle the datatypes more generically, to hopefully avoid giant switch case statements, we should in theory be able to trust that lerc gives us the correct typed array for instance.

### Verification

Added unit tests and rendered it locally 

image

### DIFF
--- a/packages/lambda-tiler/src/cli/render.tile.ts
+++ b/packages/lambda-tiler/src/cli/render.tile.ts
@@ -9,8 +9,8 @@ import { extname } from 'path';
 import { TileXyzRaster } from '../routes/tile.xyz.raster.js';
 
 // Render configuration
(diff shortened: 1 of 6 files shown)


blacha:
Screenshots are "broken" due to the new released screenshot version https://github.com/linz/basemaps-screenshot/commit/eaf8713be6de55c8aad407d3a5506a366f023636

PR #3393, opened 2025-01-22
[DO NOT MERGE] feat(cogify): migrate argo-task work to basemaps-cogify BM-1127

tawera-manaena:
### Motivation

As a Basemaps user, I want to consume the NZTopo 50 & 250 Maps [as a tile service].

### Modifications

This work defines a new CLI command. The command's purpose is to generate [STAC] files for an NZTopo Map Series imagery collection. We have designed the command to process the following collections stored in the AWS S3 [TopoReleaseArchive] directory:

```
s3://topographic-upload/TopoReleaseArchive/NZTopo50_GeoTif_Gridless/
s3://topographic-upload/TopoReleaseArchive/NZTopo250_GeoTif_Gridless/
```

There is also a [new Argo workflow in development][pr-argo-workflow] that depends on this work. The workflow's purpose is to automate the standardisation of an NZTopo Map Series imagery collection. The workflow executes this command as a step in its process flow.

### Arguments

The arguments that can be passed to the command are as follows:

- **Title**

  The name/title of the Map Series imagery collection.

  | Argument | title |
  | :--- | :--- |
  | Type | string |
  | Required | yes |
  | Example | `Raster Topographic Maps 50k` |

- **Source**

  The source directory URL of the Map Series imagery collection.

  | Argument | source |
  | :--- | :--- |
  | Type | string |
  | Required | yes |
  | Example | `s3://topographic-upload/TopoReleaseArchive/NZTopo50_GeoTif_Gridless/` |

- **Target**

  The target directory URL into which to save the generated directory structure of StacItem and StacCollection files.

  | Argument | target |
  | :--- | :--- |
  | Type | string |
  | Required | yes |
  | Example | `s3://linz-workflows-scratch/<date>/<hash>/` |

- **Scale**

  The scale of the Map Series imagery collection's map sheets.

  | Argument | scale |
  | :--- | :--- |
  | Type | string |
  | Required | yes |
  | Options | `topo25`, `topo50`, or `topo250` |

- **Resolution**

  The resolution of the Map Series imagery collection's map sheets.

  | Argument | resolution |
  | :--- | :--- |
  | Type | string |
  | Required | yes |
  | Options | `gridded_600dpi` or `gridless_600dpi` |

- **Latest Only**

  A flag used to indicate whether all of the generated files should be saved to the target location, or only that of each map sheet's latest version.

  | Argument | latestOnly |
  | :--- | :--- |
  | Type | boolean |
  | Required | no |
  | Options | `true` or `false` |

### Process

The command processes a collection as follows:

1. Loads the collection of GeoTiff images into memory
2. Extracts the map code, version, and metadata (e.g. EPSG) from each image
3. Identifies the latest version of each map sheet by code
4. Groups the images by EPSG
5. Generates a StacItem file for each image, and a StacCollection file for each EPSG grouping
6. Structures the groups of StacItem and StacCollection files into a directory tree
7. Saves the files as they are structured to a target location

### Outputs

The command groups the images by EPSG and then structures the generated StacItem and StacCollection files, as illustrated:

| Schema | Example |
| - | - |
| ![][schema] | ![][example] |

The command then saves the generated tree of folders and files into the target location directory.

### Limitations

- **Tile Matrices**

  For each image in a collection, the command extracts the EPSG of the image and converts it to a runtime `Epsg` Enum value. This value is then mapped to the EPSG's corresponding Tile Matrix definition. At this time, there is no such Tile Matrix definition for the Chatham Islands EPSG code, 3793. The task will need to be updated once Basemaps supports this definition.

[as a tile service]: https://github.com/linz/basemaps/issues/3365#issuecomment-2445444420
[STAC]: https://stacspec.org/en
[TopoReleaseArchive]: https://linz-topographic-upload.s3.ap-southeast-2.amazonaws.com/topographic/TopoReleaseArchive/

[pr-argo-workflow]: https://github.com/linz/topo-workflows/pull/932

[schema]: https://github.com/user-attachments/assets/7e4feec8-b00f-45c9-83c8-fc79040ca01d
[example]: https://github.com/user-attachments/assets/f0c5e33e-7230-477f-95f6-60b5debef424


packages/cogify/src/topo-raster/stac/create-base-stac-item.ts
```diff
@@ -0,0 +1,83 @@
...
+ * This function creates a base StacItem object based on the provided parameters.
+ *
+ * @param fileName: The map sheet's filename
+ * @example "CJ10" or "CJ10_v1-00"
+ *
+ * @param tiffItem TODO
+ *
+ * @returns a StacItem object
+ */
+export function createBaseStacItem(
```
Wentao-Kuang, line 21: I believe, the `cover` command have all stac creation logic which are very similar, wonder we can extract them to reuse together.

PR #3396, opened 2025-01-28
docs: Update empty-tiles.md with maplibregl support for empty tile content

andrewharvey:
Since Maplibre GL v4.0.0 an empty response with 200 or 204 will be treated as a transparent tile

### Motivation

update to documentation reflecting current support by maplibregl

### Modifications

update to documentation

### Verification

I tested this locally with 

```bash
nc -c 'echo -ne "HTTP/1.0 204 No Content\r\nAccess-Control-Allow-Origin: *\r\n\r\n";' -l -p 8080 -k

# and

nc -c 'echo -ne "HTTP/1.0 200 OK\r\nContent-Length: 0\r\nAccess-Control-Allow-Origin: *\r\n\r\n";' -l -p 8080 -k
```

with the following

```js
    const map = new maplibregl.Map({
        container: 'map',                                                                                                                                                                                                                      
        style: {
            'version': 8,
            'sources': {
                'osm': {
                    'type': 'raster',
                    'tiles': [ 'https://tile.openstreetmap.org/{z}/{x}/{y}.png' ],
                    'tileSize': 256
                },
                '204': {
                    'type': 'raster',
                    'tiles': [ 'http://localhost:8080/{z}/{x}/{y}' ],
                    'tileSize': 256
                }
            },
            'layers': [
                // bottom, drawn first
                {
                    'id': 'osm',
                    'type': 'raster',
                    'source': 'osm',
                    'minzoom': 0,
                    'maxzoom': 22
                },
                {
                    'id': '204',
                    'type': 'raster',
                    'source': '204',
                    'minzoom': 0,
                    'maxzoom': 22
                }
                // top, drawn last
            ]
        },
        center: [0, 0],
        zoom: 1
    });
```

no errors reported in the console and tiles from the 204 layer are drawn transparently.

I also reviewed the maplibre-gl-js source code, there are two code paths for loading tiles depending on how much the cache should be respected and both code paths seem to support a 200/204 response with no content.


blacha:
Hi @andrewharvey, thanks for this I did not know that maplibre had introduced 204 no content support! 

Looks like it will soon be supported for terrain-rgb too! https://github.com/maplibre/maplibre-gl-js/pull/5392

andrewharvey:
Ah nice, I didn't think to check the terrain sources too. In that case maplibre doesn't have complete support yet until that is addressed.

blacha:
to be fair when this documented was created, it was only looking at RGBA tiles :)

blacha (approved):

PR #3397, opened 2025-02-05
feat: CITM2000 debug tile matrix

blacha:
### Motivation

We are starting to want to view imagery in other projections that do not have a standard Web Quad tile matrixes to work with. It would be helpful to have a way to generate a debug tile matrix that aligns with the WebMercatorQuad standard scales, while also being a quad starting at 1x1 -> 2x2 -> 4x4 grids that increase with powers of two.

### Modifications

Added a CITM Debug tile matrix based off the WebMercatorQuad scales.

Added a script to generate tile matrices from a supplied EPSG projection, 

The script will download the projection extent from a projjson, then calculate an appropriate WebMercatorQuad scale that will cover the entire projection extent, 

This script was only tested with CITM and NZTM it will likely fail if used with other projections.

### Verification

Tested basemaps server with QGIS and some CI topo50 mapsheets.

image


packages/geo/src/tms/citm2000.ts
```diff
@@ -0,0 +1,196 @@
...
+import { TileMatrixSet } from '../tile.matrix.set.js';
+
+/**
+ * **Warning** this is not a standard tile matrix it is used for debugging purposes only
+ *
+ * Created using ./scripts/debug.tile.matrix.ts
+ */
+const Citm2000Tmst: TileMatrixSetType & { $generated: unknown; $options: unknown } = {
+  type: 'TileMatrixSetType',
+  title: 'Debug tile matrix for EPSG:3793',
```
Wentao-Kuang, line 12:
```suggestion
  title: 'LINZ CITM2000Quad Map Tile Grid',
```

blacha: I am on the fence, this tile matirx is designed for debugging it really hasnt had any thought about center or appropriate coverage, so leaving this showing that it is debugging only may be helpful?

blacha:
> @blacha - I'm hitting the inequality check [here](https://github.com/linz/basemaps/blob/769bea5eec661ab5122ebc2aaf9ad59ea718ae13/packages/shared/src/scripts/debug.tile.matrix.ts#L102). Looks like the `xyOrder` and `axisOrder` variables hold values for different coordinate spaces. `xyOrder` being cartesian and `axisOrder` being geographic. I don't imagine the inequality check intends for this?
> 
> ```
> # TargetProjection = 3793;
> 
> # xyOrder: 'yx'
> # axisOrder: 'n,e'
> ```
> 
> ```
> Error: getXyOrder: yx is not the same as expected ordering from proj json
> ```

This should now be fixed, it only handled when the axis was labelled `x`  or `y` not `n` or `e`

Wentao-Kuang (approved):

PR #3403, opened 2025-02-19
build: refactor container build process

blacha:
### Motivation

There is a bug building `arm64` containers under `amd64` which is causing segfaults, we currently use QEMU to cross compile the containers which can be reasonably slow.

There is also another bug where external contributors cannot trigger builds on pull requests to basemaps which leaves pull requests unable to be merged. eg: https://github.com/linz/basemaps/pull/3396

The build process for containers was also split across two workflows this merges the workflows into a single container building workflow.

### Modifications

Disable build pipelines for external pull requests as they do not have access to our secrets, which causes them to fail.

Split container builds into a matrix build using arm64 runners for arm containers and amd64 runners for x86 containers.

### Verification

Created a number of containers with `container` tag on pull requests and pushed commits to the master branch on `blacha/basemaps`


blacha:
containers are unable to be pushed as this is from a basemaps fork :sob: I wonder if there is a better way of handling building containers when building off a public fork.

PR #3411, opened 2025-03-19
[WIP] feat(cogify): support band expansion of greyscale imagery

blacha:
### Motivation

LINZ is now publishing one band grey scale `[uint8]` lerc datasets such as 

```
s3://nz-elevation/new-zealand/new-zealand/dem-hillshade/2193/ 
s3://nz-elevation/new-zealand/new-zealand/dem-hillshade-igor/2193/
```

Basemaps prefers grey scale imagery to be stored as webp, webp requires 3 or 4 bands of source imagery.

### Modifications

if the `webp` preset is used on imagery that is one band `uint8` expand the source imagery into 3 bands of the same `uint8` 

if a source pixel's value is `240` then it will be converted into RGB: `240,240,240` 

### Verification

Imported the dem-hillshade locally

### DIFF
--- a/packages/cogify/src/cogify/cli/cli.cog.ts
+++ b/packages/cogify/src/cogify/cli/cli.cog.ts
@@ -21,6 +21,7 @@ import {
   gdalBuildVrt,
   gdalBuildVrtWarp,
   gdalCreate,
+  VrtOptions,
 } from '../gdal/gdal.command.js';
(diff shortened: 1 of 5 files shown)


blacha:
No longer needed, greyscale imagery is automatically expanded via the color-ramp pipeline.

PR #3364, opened 2024-10-29
feat(landing): enable labels upon first visit BM-1101

tawera-manaena:
### Motivation

As a Basemaps user, I would like to see labels when I arrive at the site so I can see the context of the imagery displayed to me.

### Modifications

- Updated the **landing** package so that labels are enabled upon first visit to the site via the default Basemaps URL (https://basemaps.linz.govt.nz).

   Situations where labels are **not enabled** upon first visit:
   - The `labels` URL query string is `false`, e.g. https://basemaps.linz.govt.nz/?labels=false
   - The `debug` URL query string is present, e.g. https://basemaps.linz.govt.nz/?debug

### Verification

- Added additional tests to the **landing** package's  `map.config.test` test suite.

#### Labels should be enabled

| https://basemaps.linz.govt.nz |
| - |
| ![][default] |
| Viewing the **aerial** layer and not in debug mode. Enable labels upon first visit. |

#### Labels should be disabled

| https://basemaps.linz.govt.nz/?labels=false |
| - |
| ![][labels=false] |
| The `labels` query string is present and equals `false`. Don't enable labels upon first visit. |

| https://basemaps.linz.govt.nz/?debug |
| - |
| ![][debug] |
| In debug mode. Don't enable labels upon first visit. |

| https://basemaps.linz.govt.nz/?i=ashburton-2023-0.1m |
| - |
| ![][individual_layer] |
| _Adjusted viewport to show imagery_ |
| Not viewing the `aerial` layer. Don't enable labels upon first visit. |

---

Resolves bug: [BM-1114](https://toitutewhenua.atlassian.net/browse/BM-1114)

[default]: https://github.com/user-attachments/assets/1b450c01-3409-400e-b794-6eccc6582021
[labels=false]: https://github.com/user-attachments/assets/2ca74653-d868-4420-8fd4-4e22273efa7d
[debug]: https://github.com/user-attachments/assets/462ac80d-a603-47a7-9b9b-683b59988521
[individual_layer]: https://github.com/user-attachments/assets/bac0f8a7-9a67-4683-880b-5a5fe484ae8b

[BM-1114]: https://toitutewhenua.atlassian.net/browse/BM-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3413, opened 2025-03-24
refactor(landing): refine the labels param parsing logic

tawera-manaena:
### Motivation

While looking through the `basemaps/landing` package, I noticed that the code for handling the `labels` URL query parameter is hard to understand. The default behavior is also unclear. This work aims to fix that.

- Refactors code from https://github.com/linz/basemaps/pull/3364

### Modifications

1. Modified the [code] responsible for parsing the `labels` URL query parameter. The code lives within the [updateFromUrl] function. A 'before' and 'after' is as follows:

   #### Before

   The current logic, although concise, is a little difficult to interpret:

   ```typescript
   if (labels == null) {
     this.labels = layerId === "aerial" && this.isDebug === false;
   } else {
     this.labels = labels !== "false";
   }
   ```

   #### After

   The updated logic, although verbose, seeks to clarify the process flow and explain each statement's context:

   ```typescript
   this.labels = false;

   if (typeof labels === "string") {
     // enable labels for any string value other than "false"
     if (labels !== "false") this.labels = true;
   } else {
     // if not in debug mode, show labels for the aerial layer by default
     if (layerId === "aerial" && !this.isDebug) this.labels = true;
   }
   ```

2. Expanded upon some of the [existing test cases][adjusted] that validate the parsing logic for the `labels` URL query parameter.

### Verification

All of the [original] and [adjusted] test cases pass, indicating that the parsing logic remains unchanged.

[code]: https://github.com/linz/basemaps/pull/3413/files#diff-f03767e7b3c93746f0651b4ce3f2d304bdf32185d88bb0d3374d03781ed22348

[updateFromUrl]: https://github.com/linz/basemaps/blob/e874e2b678ae074f344f4ce0fef6c6460fb42495/packages/landing/src/config.map.ts#L125

[original]: https://github.com/linz/basemaps/blob/e874e2b678ae074f344f4ce0fef6c6460fb42495/packages/landing/src/__tests__/map.config.test.ts#L204
[adjusted]: https://github.com/linz/basemaps/pull/3413/files#diff-31cb3968d74549f956e1a33e2913e003c78ea11bf84c02474b1d7241eb39ead7

### DIFF
--- a/packages/landing/src/config.map.ts
+++ b/packages/landing/src/config.map.ts
@@ -145,10 +145,14 @@ export class MapConfig extends Emitter<MapConfigEvents> {
     this.style = style ?? null;
     this.layerId = layerId.startsWith('im_') ? layerId.slice(3) : layerId;
     this.tileMatrix = tileMatrix;
-    if (labels == null) {
-      this.labels = layerId === 'aerial' && this.isDebug === false;
+    this.labels = false;
+
+    if (typeof labels === 'string') {
+      // enable labels for any string value other than "false"
+      if (labels !== 'false') this.labels = true;
     } else {
-      this.labels = labels !== 'false';
+      // if not in debug mode, show labels for the aerial layer by default
+      if (layerId === 'aerial' && !this.isDebug) this.labels = true;
     }
 
     if (this.layerId === 'topographic' && this.style == null) this.style = 'topographic';
--- a/packages/landing/src/__tests__/map.config.test.ts
+++ b/packages/landing/src/__tests__/map.config.test.ts
@@ -209,10 +209,20 @@ describe('WindowUrl', () => {
     assert.equal(mc.labels, true);
 
     // aerial layer, labels enabled & debug disabled
-    mc.updateFromUrl('?labels=true');
-    assert.equal(mc.layerId, 'aerial');
-    assert.equal(mc.isDebug, false);
-    assert.equal(mc.labels, true);
+    for (const params of ['?labels', '?labels=true']) {
+      mc.updateFromUrl(params);
+      assert.equal(mc.layerId, 'aerial');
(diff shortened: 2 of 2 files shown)


packages/landing/src/config.map.ts
```diff
@@ -145,10 +145,14 @@ export class MapConfig extends Emitter<MapConfigEvents> {
     this.style = style ?? null;
     this.layerId = layerId.startsWith('im_') ? layerId.slice(3) : layerId;
     this.tileMatrix = tileMatrix;
-    if (labels == null) {
-      this.labels = layerId === 'aerial' && this.isDebug === false;
+    this.labels = false;
+
+    if (typeof labels === 'string') {
```
blacha, line 150: the typing for `.get` is `URLSearchParam.get(name: string): string | null;`  so it can only ever be a `string` or `null`,

so `labels == null` is clearer to me over `typeof`

For example:

```typescript
new URLSearchParameters('x=1').get('x') // '1'
```

packages/landing/src/config.map.ts
```diff
@@ -145,10 +145,14 @@ export class MapConfig extends Emitter<MapConfigEvents> {
...
+      // enable labels for any string value other than "false"
+      if (labels !== 'false') this.labels = true;
     } else {
-      this.labels = labels !== 'false';
+      // if not in debug mode, show labels for the aerial layer by default
+      if (layerId === 'aerial' && !this.isDebug) this.labels = true;
```
blacha, line 155: We generally try and avoid `!foo` as that triggers type cohersion causing you to think of all the horrible values that javascript would consider to be truthy.

We generally try to be as explicit as possible, so `this.isDebug === false` would be my preference

PR #3421, opened 2025-04-03
feat(landing): add dsm to layer dropdown BM-1248

ccbblin:
### Motivation

We've recently published an [Elevation DSM Basemap][dsm-config]. This work add a new "Elevation DSM" entry to the `Layers` dropdown, so that customers can view the new Basemap in our application.

### Modifications

1. Added a new "Elevation DSM" entry to the `Layers` dropdown.
2. Renamed the existing "Elevation" entry to "Elevation DEM".

### Verification

| Before | After |
| - | - |
| ![][img_before] | ![][img_after] |

[dsm-config]: https://github.com/linz/basemaps-config/blob/master/config/tileset/elevation.dsm.json

[img_before]: https://github.com/user-attachments/assets/1cf65ae8-fd0d-4f95-9590-709a57723aba
[img_after]: https://github.com/user-attachments/assets/e68dbba3-0d09-43ec-8d30-5ebaf1ec9cad

### DIFF
--- a/packages/landing/src/config.map.ts
+++ b/packages/landing/src/config.map.ts
@@ -413,8 +413,17 @@ function addDefaultLayers(output: Map<string, LayerInfo>): void {
 
     {
       id: 'elevation',
-      title: 'Elevation',
+      title: 'Elevation DEM',
       projections: new Set([EpsgCode.Google]),
+      category: 'Basemaps',
+      pipeline: 'terrain-rgb',
+      imageFormat: 'png',
+    },
+
+    {
+      id: 'elevation-dsm',
+      title: 'Elevation DSM',
+      projections: new Set([EpsgCode.Nztm2000, EpsgCode.Google]),
       category: 'Basemaps',
       pipeline: 'terrain-rgb',
       imageFormat: 'png',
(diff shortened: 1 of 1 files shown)


packages/landing/src/config.map.ts
```diff
@@ -413,13 +413,22 @@ function addDefaultLayers(output: Map<string, LayerInfo>): void {
 
     {
       id: 'elevation',
-      title: 'Elevation',
+      title: 'Elevation DEM',
       projections: new Set([EpsgCode.Google]),
```
Wentao-Kuang, line 417: I think we [support DEM in 2193 now already](https://github.com/linz/basemaps-config/pull/922). Might be good to add that in together.
```suggestion
      projections: new Set([EpsgCode.Nztm2000, EpsgCode.Google]),
```

dwsilk: I don't think it's possible for this to be used by anything except Basemaps??

packages/landing/src/config.map.ts
```diff
@@ -413,13 +413,22 @@ function addDefaultLayers(output: Map<string, LayerInfo>): void {
...
       category: 'Basemaps',
       pipeline: 'terrain-rgb',
       imageFormat: 'png',
     },
 
+    {
+      id: 'elevation-dsm',
+      title: 'Elevation DSM',
+      projections: new Set([EpsgCode.Nztm2000, EpsgCode.Google]),
```
Wentao-Kuang, line 426: As discussed this morning. I think we want to replicate this with Elevation, and only add to webmercator projection.

PR #3432, opened 2025-04-23
feat(cli): move cogify create-config into cli-config package BM-1261

ccbblin:
### Motivation

The `create-config` command of the `cogify` package should live alongside the other `basemaps-config`-related commands in the new `cli-config` package.

### Modifications

1. Relocate the `cogify-config` command from the `cogify` package to the `cli-config` package.
2. Import the command into the `cli` package as a `config create` sub-command.
3. We've moved some utility functions to the `shared` package. So, this work also updates various functions to use the new _shared_ equivalents instead, removing the singular copies in the process.

2b82d71c-b820-410f-b05b-4b323f90745f

### Verification

check `bin/bmc.js config create-config` runs correctly


packages/cli-config/src/cli/action.create.config.ts
```diff
@@ -1,3 +1,5 @@
+import { isArgo } from '@basemaps/cogify/build/argo.js';
+import { Url, UrlFolder } from '@basemaps/cogify/build/cogify/parsers.js';
```
ccbblin, line 2: we can move `isArgo, Url, UrlFolder` to `@basemap/shared` later

Wentao-Kuang (dismissed):

Wentao-Kuang:
> Member

`const logger = LogConfig.get();` is the default way of getting logger for other packages. 
The `const logger = getLogger(this, args, 'cli-config');` should be used for cli packages that include the cli versions/packages infor in the logger.

Wentao-Kuang (approved):

PR #3435, opened 2025-04-28
feat(cli-vector): New cli to create and join mbtiles for vector map. BM-1268

Wentao-Kuang:
### Motivation
We want to move the Vector ETL work from [basemaps-team](https://github.com/linz/basemaps-team) into basemaps/packges/cli-vector.

The new argo workflow for the Vector shortbread ETL, Need at least 3 cli to make it work.
image

- extract cli that load the schemas and source data, then compare with cache mbtiles in the s3 cache bucket to output updated layers(layer version update, new configurations updates in schema, or new cli version in packages) that need to process.
- create cli that load stac item from extract cli to create individual mbtiles, then push into s3 cache bucket.
- join cli will load all the cached mbtiles to join into the output mbtiles

### Modifications

- This PR tidy up all the 3 clis and put into a test container to run tests in argo.
- update the cli container to include tippecanoe.
- update the cli container to node 24, in order to use node:sqlite
- update the `@types/node` to 22, in order to use node:sqlite

### Verification

- Testing workflow: https://argo.linzaccess.com/workflows/argo/test-basemaps-vector-etl-shortbread-r5hk9?tab=workflow&uid=a334ca91-79e6-4d87-a39f-dc380c3e41fd
- Takes 23mins for a full run, could be much faster after we got cached individual mbtile in s3.
- Preview of the output https://basemaps.linz.govt.nz/@-41.6956717,173.3842998,z5.51?style=topographic-v2&i=topographic&config=TmVmbYRQjL9T2JWgyaSie193b4D1qZnBD8hjaSqPFYSsEvEYYhaGYKrckSaNg716tfRgq2Qs2Ur3cx2bzBXUMsmuHYy7xhhtyuSo8rB6QLvA8PebFhVp8YNNr7jSKgM&debug=true


packages/cli-vector/src/generalization/generalization.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  // Simplify geometry
+  if (simplify != null) {
+    // Update the simplified feature zoom level
+    feature['tippecanoe'] = {
+      layer: options.name,
+      minzoom: simplify.style.minZoom,
+      maxzoom: simplify.style.maxZoom,
+    };
+    if (simplify.tolerance != null) {
```
ccbblin, line 92: needed ` feature.geometry != null`

packages/cli-vector/src/generalization/generalization.ts
```diff
@@ -0,0 +1,168 @@
...
+  for await (const line of rl) {
+    if (line === '') continue;
+    inputCount++;
+    // For simplify Duplicate feature for each zoom level with different tolerance
+    if (simplify != null) {
+      for (const s of simplify) {
+        const feature = tag(options, line, s);
+        if (feature == null) continue;
+        outputCount++;
+        features.push(JSON.parse(feature) as VectorGeoFeature);
```
ccbblin, line 46: can we `writeStream.write(feature + '\n');` instead of storing it in memory, hitting some oom errors when running locally

packages/cli-vector/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM node:20-slim as builder
```
Wentao-Kuang, line 1: This is just temporary file that I copy here as reference, we will move all the installs and ENV into package/cli Dockerfile later. Then test the vector cli in the cli container

packages/cli-vector/package.json
```diff
@@ -37,10 +37,14 @@
...
     "@basemaps/config": "^8.1.0",
+    "@basemaps/geo": "^8.0.0",
     "@basemaps/shared": "^8.1.0",
-    "@types/node-fetch": "^2.6.12",
+    "@types/geojson": "^7946.0.16",
+    "@types/mapbox__geojson-area": "^0.2.6",
+    "@types/p-limit": "^2.1.0",
+    "@types/polylabel": "^1.1.3",
+    "@types/tar-stream": "^2.2.2",
     "cmd-ts": "^0.12.1",
```
ccbblin, line 47: apart from the https://github.com/types, I think the rest should be in dependencies

packages/cli-vector/package.json
```diff
@@ -49,5 +53,16 @@
...
     "dist/"
-  ]
+  ],
+  "dependencies": {
+    "@linzjs/docker-command": "^7.5.0",
+    "@mapbox/geojson-area": "^0.2.2",
+    "geojson": "^0.5.0",
+    "p-limit": "^6.2.0",
+    "polylabel": "^2.0.1",
+    "sqlite": "^5.1.1",
```
blacha, line 63: we should not be using `sqlite` use `node:sqlite`

packages/cli-vector/package.json
```diff
@@ -49,5 +53,16 @@
   "files": [
     "build/",
     "dist/"
-  ]
+  ],
+  "dependencies": {
+    "@linzjs/docker-command": "^7.5.0",
+    "@mapbox/geojson-area": "^0.2.2",
```
blacha, line 59: We have `Area` in @basemaps/geo does this need to exist?

Wentao-Kuang: I think this is used to calculate the size of polygon for large lakes, do we have the same function in `Aera`.

packages/cli-vector/package.json
```diff
@@ -49,5 +53,16 @@
...
+    "geojson": "^0.5.0",
```
blacha, line 60: what does this do that @linzjs/geojson doesnt do?

packages/cli-vector/package.json
blacha: > I think this is used to calculate the size of polygon for large lakes, do we have the same function in Aera.

Area can give it to you in sqkm, if you reproject it to 2193, otherwise you can run it on the latlngs and it will give you it in square degrees which can be used to approximate the size of the lakes

ccbblin (dismissed):

packages/cli-vector/src/cli/cli.create.ts
```diff
@@ -0,0 +1,254 @@
...
+    long: 'from-file',
+    description: 'Path to JSON file containing array of paths to mbtiles stac json.',
+  }),
+  concurrency: option({
+    type: number,
+    long: 'concurrency',
+    defaultValue: () => 1,
+    defaultValueIsSerializable: true,
+  }),
+  join: flag({
```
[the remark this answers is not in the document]
Wentao-Kuang: We can keep this for local running and testing maybe? So we don't need run join cli when test this locally.

PR #3440, opened 2025-05-06
feat(cli-vector): introduce logic for overriding feature metadata and properties BM-1268

tawera-manaena:
1. This pull request extends the [feat/etl-create] branch, not the [master] branch.
2. The changed files contain short-circuiting code to enable faster debugging that needs to be purged before approval.

---

### Motivation

We are currently migrating our Vector ETL system into basemaps as a new `cli-vector` package. We are also re-building the system in the process. As a part of the re-build, we have streamlined the system's architecture in the following four steps:

1. `extract`
2. `create`
3. `join`
4. `analyse`

This work implements an extension to the base branch's implementation of the `create` CLI command. This work introduces a series of functions for handling dataset features of specific Shortbread layers for which we require special tagging.

### Modifications

| Diagram |
| - |
| ![][diagram1] |

1. Implements a switch-style pattern for identifying the Shortbread layer to which a feature is assigned.
2. Implements bespoke functions for overriding a feature's metadata and properties for the following Shortbread layers:
   - `contours`
   - `place_labels` - This one is weird. We will need to discuss the approach for this one.
   - `pois` - This one doesn't actually do anything as it never receives any features that will meet its conditions for special tagging. We can discuss this also.
   - `public_transport`
   - `street_labels` - This one's logic is very similar to the `streets` logic.
   - `streets` - This one's logic is very similar to the `street_labels` logic.
   - `water_polygons`

#### Processing a Vector Stac Item

The system processes a Stac Item file created via the **cli-vector** package's new `extract` command like so:

1. Parse the file as a runtime `VectorStacItem` type.
2. Download the source file (i.e. a GeoPackage file).
3. Convert the source file into an NDJSON file.
4. Parse each line (i.e. feature) of the NDJSON file, generalise it (i.e. simplify, add/remove attributes, or discard).
5. Process the collection of generalised features into an mbtiles file.
6. Upload a copy of the mbtiles file to the same directory as the Stac Item file.
7. Update the Stac Item file's contents and overwrite it.

| Diagram |
| - |
| ![][diagram2] |

### Verification

1. Established test cases for the special tagging functions for the following Shortbread layers:
   - `contours`

More to come if desired.

[feat/etl-create]: https://github.com/linz/basemaps/tree/feat/etl-create
[master]: https://github.com/linz/basemaps/tree/master

[diagram1]: https://github.com/user-attachments/assets/6a405417-59fb-4127-8630-cb0901c12618
[diagram2]: https://github.com/user-attachments/assets/2994bd5d-c668-4b99-957e-57623da7c6a1

### DIFF
--- a/packages/cli-vector/src/modify/layers/contours.ts
+++ b/packages/cli-vector/src/modify/layers/contours.ts
@@ -8,7 +8,7 @@ import { VectorGeoFeature } from '../../types/VectorGeoFeature.js';
  *
  * @param feature - the feature to process
  * @param options - the layer's options
- * @param logger
+ * @param logger - a logger instance
  * @returns the processed feature
  */
 export function handleLayerContours(
@@ -36,7 +36,7 @@ export function handleLayerContours(
  * Processes a 'contours' layer feature with a 'kind' value of 'contours'.
  *
  * @param feature - the feature to process
- * @param logger
+ * @param logger - a logger instance
  * @returns the processed feature
  */
 export function handleKindContours(feature: VectorGeoFeature, logger: LogType): VectorGeoFeature {
@@ -68,7 +68,7 @@ export function handleKindContours(feature: VectorGeoFeature, logger: LogType): 
  * Processes a 'contours' layer feature with a 'kind' value of 'peak'.
  *
  * @param feature - the feature to process
- * @param logger
+ * @param logger - a logger instance
  * @returns the processed feature
  */
 export function handleKindPeak(feature: VectorGeoFeature, logger: LogType): VectorGeoFeature {
--- a/packages/cli-vector/schema/streets.json
+++ b/packages/cli-vector/schema/streets.json
@@ -17,7 +17,8 @@
       "tunnel",
       "hway_num",
       "use_2",
-      "ref"
+      "ref",
+      "subclass"
     ]
   },
   "layers": [
--- a/packages/cli-vector/src/cli/cli.create.ts
+++ b/packages/cli-vector/src/cli/cli.create.ts
@@ -6,19 +6,12 @@ import PLimit from 'p-limit';
 import { createGunzip } from 'zlib';
 
 import { generalize } from '../generalization/generalization.js';
-import { Metrics } from '../schema-loader/schema.js';
 import { VectorStacItem } from '../stac.js';
 import { ogr2ogrNDJson } from '../transform/ogr2ogr.js';
 import { tileJoin, tippecanoe } from '../transform/tippecanoe.js';
-import { prepareTmpPaths } from '../util.js';
+import { ContentType, prepareTmpPaths, TmpPaths } from '../util.js';
 
-const tmpPath = fsa.toUrl('tmp/create/');
-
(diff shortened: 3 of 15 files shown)


ccbblin (approved):

packages/cli-vector/src/generalization/generalization.ts
```diff
@@ -32,87 +30,118 @@ export async function generalize(input: URL, output: URL, options: VectorCreatio
...
-    writeStream.close();
-    return output;
   }
-  return undefined;
+
+  await new Promise((resolve) => {
+    writeStream.close(resolve);
+  });
+
+  // special handling for the gazatteer (place_labels) dataset
```
Wentao-Kuang, line 62: Shall we delete the comment out code if not needed?

packages/cli-vector/src/generalization/generalization.ts
```diff
@@ -32,87 +30,118 @@ export async function generalize(input: URL, output: URL, options: VectorCreatio
...
+  });
 
+  // copy the stac json's tags to the feature (i.e. 'kind')
   Object.entries(options.layer.tags).forEach(([key, value]) => (feature.properties[key] = value));
 
-  // TODO: Add support of special tagging logics.
+  // adjust the feature's metadata and properties
+  const modifiedFeature = modifyFeature(feature, options, logger);
+  if (modifiedFeature == null) {
+    logger.debug({}, 'Tag:End');
```
Wentao-Kuang, line 104: I feel like these logs are not needed with not much information in it. And might dumping terminal a lot if we goes with debug mode.

packages/cli-vector/src/generalization/generalization.ts
```diff
@@ -32,87 +30,118 @@ export async function generalize(input: URL, output: URL, options: VectorCreatio
...
     }
   }
 
   // Remove unused properties
-  removeAttributes(feature, options);
+  // REVIEW: this function just removes the special tags. something isn't right here
+  const cleanedFeature = removeAttributes(modifiedFeature, options);
 
-  return JSON.stringify(feature);
+  logger.debug({}, 'Tag:End');
```
Wentao-Kuang, line 132: Same dumping logs too

Wentao-Kuang: I believe we want the logs like `ndjson start/end`, `generaize start/end` and `mbtiles start/end` for each layer.

packages/cli-vector/src/modify/layers/contours.ts
```diff
@@ -0,0 +1,104 @@
...
+ * @param options - the layer's options
+ * @param logger
+ * @returns the processed feature
+ */
+export function handleLayerContours(
+  feature: VectorGeoFeature,
+  options: VectorCreationOptions,
+  logger: LogType,
+): VectorGeoFeature {
+  logger.debug({}, 'HandleContours:Start');
```
Wentao-Kuang, line 19: Too much similar logs, I wonder we want to set --verbose to false to disable debug logs as default. Otherwise, we got one log for each feature.

PR #3441, opened 2025-05-08
feat(lambda-tiler)!: allow API keys to be disabled BM-1281

blacha:
### Motivation

We have seen a increase of consumer api keys using scripted methods to pull large amounts of data from basemaps.linz.govt.nz these users should be using bulk data access methods like linz/imagery or limz/elevation, as the keys are only consumer we have no easy way of contacting them.

Our only option is to disable their keys and hope that they reach out.

### Modifications

Add a environment variable to allow us to block API keys 

### Verification

unit tests and further tests in nonprod.

### DIFF
--- a/packages/lambda-tiler/src/util/validate.ts
+++ b/packages/lambda-tiler/src/util/validate.ts
@@ -1,6 +1,6 @@
 import { ConfigTileSetRaster, ConfigTileSetRasterOutput } from '@basemaps/config';
 import { ImageFormat, LatLon, Projection, TileMatrixSet, TileMatrixSets } from '@basemaps/geo';
-import { Const, isValidApiKey, truncateApiKey } from '@basemaps/shared';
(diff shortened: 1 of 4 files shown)


packages/lambda-tiler/src/util/validate.ts
```diff
@@ -36,6 +39,11 @@ export const Validate = {
     if (!valid.valid) throw new LambdaHttpResponse(400, 'API Key Invalid: ' + valid.message);
     // Truncate the API Key so we are not logging the full key
     req.set('api', truncateApiKey(apiKey));
+
+    if (this.blockedApiKeys.has(apiKey as string)) {
+      throw new LambdaHttpResponse(429, 'Too many requests! Please contact basemaps@linz.govt.nz');
```
ccbblin, line 44: Too many requests! Please contact basemaps@linz.govt.nz for a developer key

blacha: I have updated to your wording.

PR #3427, opened 2025-04-10
feat(cli): update cli-package commands to cmd-ts BM-1259

tawera-manaena:
### Motivation

We are in the process of migrating our Vector ETL system into Basemaps as a new package. To prepare for this migration, we are undertaking some housekeeping work (BM-1214). This pull request focuses on commands of the `cli` package.

### Modifications

1. Upgrades the following commands of the `cli` package from [rushstack], to [cmd-ts]:
   
   - `packages/cli/src/cli/config/action.bundle.assets.ts`
   - `packages/cli/src/cli/config/action.bundle.ts`
   - `packages/cli/src/cli/config/action.import.ts`

2. Deprecates the following commands of the `cli` package:

   - `packages/cli/src/cli/server/action.serve.ts`

3. Updates the following GitHub workflows to no longer use the deprecated `action.serve.ts` command:

   - `.github/workflows/screenshot.yml`
   - `.github/workflows/smoke-test.yml`

   These workflows now use the `basemaps-server` CLI command of the `server` package.

4. Updates the `cli` package's dependencies, installing those required, and uninstalled those no longer needed.

5. Updates the `cli` package's Dockerfile to no longer install the `basemaps-landing` and `basemaps-server` assets.

### Intention

We intend to migrate these commands to a new package called `cli-config` in another piece of work (BM-1260). The idea for this pull request is to update the commands without introducing any breaking changes.

### Verification

1. This is what the CLI for the commands looks like now:

| Before ([rushstack]) | After ([cmd-ts]) |
| - | - |
| ![][img_before] | ![][img_after] |

2. All three of the upgraded commands work locally without issue.

3. The updated GitHub workflows execute and succeed without issue.

[BM-1214]: https://toitutewhenua.atlassian.net/browse/BM-1214
[BM-1260]: https://toitutewhenua.atlassian.net/browse/BM-1260

[rushstack]: https://github.com/microsoft/rushstack
[cmd-ts]: https://github.com/Schniz/cmd-ts

[img_before]: https://github.com/user-attachments/assets/008a7cb5-6b7f-4b34-b8e1-314ff8d050da
[img_after]: https://github.com/user-attachments/assets/3a3755e1-acbc-4d1c-aa44-c55b0c669b0c

PR #3445, opened 2025-05-13
fix: correct cli-config 'import' parameters

tawera-manaena:
### Motivation

In an [earlier piece of work], we converted the `import` command from [rushstack] to [cmd-ts]. In doing so, we mistakenly converted the `output` parameter from an **optional** string, to a **non-optional** string.

This causes the basemaps-config's [Deploy Nonprod] job to fail. The [failing command] and error are as follows:

```yml
- name: Import config
  run: |
    ./scripts/bmc.sh ./index.cjs config import --config ${CONFIG_PATH} --commit
```

```
error: found 1 error

  1. No value provided for --output
```

### Modifications

1. Updated the `cli-config` package's `import`command's `output` parameter to an **optional** string.
2. Added missing references to the `cli` package's `tsconfig.json` file to resolve type declaration errors.
3. Re-ordered the root `tsconfig.json` file's references for easier maintenance.

### Verification

I've captured a screenshot showing that the `import` command's `output` parameter is now optional:

![][img-output-param]

I've also tested the `import` command locally with the `--commit` flag omitted. The command does require a valid `config`, so I used a random one from the S3 bucket:

```
node packages/cli/bin/bmc.js config import --config s3://linz-basemaps/config/config-28Fp43sn1XCmsh9QX6hwsyKESrm6e67PkG3agcVWmLEv.json.gz
```

We can see that the command executes successfully without requiring the `output` parameter:

```
...
[2025-05-14T02:08:06.137Z] INFO Change:Update type=ts record=ts_all description=all commit=false
[2025-05-14T02:08:06.137Z] INFO Import:Compare:Done objects=820 types={ ts=492 st=5 pv=1 im=322 }
[2025-05-14T02:08:06.138Z] INFO DryRun:Done
```

[earlier piece of work]: https://github.com/linz/basemaps/pull/3427/commits/01a90a2b98bab02632a9637bfed2067c9c047c61#diff-48a5e932b724482f9a280931b7bb1188d26ef004b87c331aadfc7a7948ec7f70

[rushstack]: https://github.com/microsoft/rushstack
[cmd-ts]: https://github.com/Schniz/cmd-ts

[Deploy Nonprod]: https://github.com/linz/basemaps-config/actions/runs/15007294128/job/42168979816

[failing command]: https://github.com/linz/basemaps-config/blob/a5b8b9baaa9941314a13058e53945bc5db5f5357/.github/workflows/build.yml#L131

[img-output-param]: https://github.com/user-attachments/assets/ddffcbbc-950d-44d8-af67-1b2d0149289d


tsconfig.json
```diff
@@ -5,25 +5,25 @@
...
+    { "path": "./packages/lambda-analytics/tsconfig.json" },
+    { "path": "./packages/lambda-tiler/tsconfig.json" },
     { "path": "./packages/landing/tsconfig.json" },
     { "path": "./packages/server/tsconfig.json" },
-    { "path": "./packages/smoke/tsconfig.json" }
+    { "path": "./packages/shared/tsconfig.json" },
+    { "path": "./packages/smoke/tsconfig.json" },
+    { "path": "./packages/sprites/tsconfig.json" },
+    { "path": "./packages/tiler/tsconfig.json" },
+    { "path": "./packages/tiler-sharp/tsconfig.json" }
```
blacha, line 27: this ordering is a little weird, `lambda-tiler` needs `tiler` and `tiler-sharp` so the order doesnt matter here?

blacha: It would sort of imply that something is maybe missing a project reference? what error do you get when you move it to the bottom?

PR #3446, opened 2025-05-14
feat(lambda-tiler): expose configuration id and hash if present

blacha:
### Motivation

I would like to know what configuration is in use on each instance of basemaps, we have a version endpoint `/v1/version` that gives the build information (commit hash, github actions run id) but does not say what configuration is being used.

### Modifications

If the current configuration has a `id` and `hash` return those as `{ config: { id, hash} }`

```json
{ 
  "id": "cb_01JTQ7ZK49F8EY4N5DRJ3XFT73", 
  "hash": "HcByZ8WS2zpaTxFJp6wSKg2eUpwahLqAGEQdcDxKxqp6" 
}
```

### Verification

Unit tests + running locally

### DIFF
--- a/packages/lambda-tiler/src/routes/version.ts
+++ b/packages/lambda-tiler/src/routes/version.ts
@@ -1,6 +1,16 @@
-import { HttpHeader, LambdaHttpResponse } from '@linzjs/lambda';
+import { BasemapsConfigProvider, ConfigProviderMemory } from '@basemaps/config';
+import { HttpHeader, LambdaHttpRequest, LambdaHttpResponse } from '@linzjs/lambda';
 
-export function versionGet(): Promise<LambdaHttpResponse> {
+import { ConfigLoader } from '../util/config.loader.js';
+
+function getConfigHash(cfg: BasemapsConfigProvider): { id: string; hash?: string } | undefined {
+  if (!ConfigProviderMemory.is(cfg)) return undefined;
+  if (cfg.id == null) return undefined;
+  return { id: cfg.id, hash: cfg.hash };
+}
+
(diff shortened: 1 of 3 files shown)


packages/lambda-tiler/src/routes/version.ts
```diff
@@ -1,6 +1,16 @@
-import { HttpHeader, LambdaHttpResponse } from '@linzjs/lambda';
+import { BasemapsConfigProvider, ConfigProviderMemory } from '@basemaps/config';
+import { HttpHeader, LambdaHttpRequest, LambdaHttpResponse } from '@linzjs/lambda';
 
-export function versionGet(): Promise<LambdaHttpResponse> {
+import { ConfigLoader } from '../util/config.loader.js';
+
+function getConfigHash(cfg: BasemapsConfigProvider): { id: string; hash?: string } | undefined {
+  if (!ConfigProviderMemory.is(cfg)) return undefined;
+  if (cfg.id == null) return undefined;
```
Wentao-Kuang, line 8: We still require dynamoDB to points to the config hash here, just double check our final goal is to use environment variables for the config hash? If so we might not need the id here?

blacha: the `id` for JSON files is the config bundle id eg `cb_01123abc123` 

generally this should be a config from JSON file by default.

Wentao-Kuang: Oh, I forgot that's the json id as well. I only thought that's the dynameDB id as it starts with `cb_`.

Wentao-Kuang (approved):

PR #3457, opened 2025-06-13
fix(cli-vector): collapse large analysis report in pr comment

ccbblin:
### Motivation

Tile size report is large, we will have two reports for Webmecator and NZTM.

### Modifications

collapse report


Wentao-Kuang:
I got this [error](https://github.com/linz/basemaps-config/actions/runs/15668559841/job/44135920612?pr=1102) from my previous changes. Could you please fix it together in here. I think we need check exists before `fsa.read` for the report.mb.

ccbblin:
> I got this [error](https://github.com/linz/basemaps-config/actions/runs/15668559841/job/44135920612?pr=1102) from my previous changes. Could you please fix it together in here. I think we need check exists before `fsa.read` for the report.mb.

ah you fixed this already :) https://github.com/linz/basemaps/blob/53134c0782cdea7fb683170eda30fb675d60923a/packages/cli-config/src/cli/action.import.ts#L294 We need to do a new release

PR #3482, opened 2025-07-31
feat: add aliases for tileSets

ccbblin:
### Motivation

Aerial Imagery PRs with the 'Aerial Imagery Deletes' messages are breaking existing WMTS links, 'urban' or 'rural' has been removed from the layer name as it's taken out of the aerial config, which in turn, changes the URL of the individual layer WMTS

### Modifications

Add an `aliases` property so that tilesets can be linked by different names:

### Verification

added test

### DIFF
--- a/packages/config-loader/src/json/json.config.ts
+++ b/packages/config-loader/src/json/json.config.ts
@@ -255,7 +255,7 @@ export class ConfigJson {
       tileSet.format = ts.type === TileSetType.Vector ? 'pbf' : 'webp';
     }
 
-    tileSet.alias = ts.alias;
+    tileSet.aliases = ts.aliases;
 
     return tileSet as ConfigTileSet;
   }
--- a/packages/config-loader/src/json/json.config.ts
+++ b/packages/config-loader/src/json/json.config.ts
@@ -255,6 +255,8 @@ export class ConfigJson {
       tileSet.format = ts.type === TileSetType.Vector ? 'pbf' : 'webp';
     }
 
+    tileSet.alias = ts.alias;
+
     return tileSet as ConfigTileSet;
   }
 
--- a/packages/config/src/config/base.ts
+++ b/packages/config/src/config/base.ts
@@ -41,6 +41,11 @@ export const ConfigBase = z.object({
   /**
    * Was this configuration object generated from another object
    *
+   * tileset-all: All tileset contains all raster layers
+   * tileset-alias: Alias that points to another tileset
+   * imagery-name: Tileset that is identified by its imagery name
+   * imagery-id: Tileset that is identified by its imagery id
+   *
    * @default undefined
    */
   virtual: z.enum(['tileset-all', 'tileset-alias', 'imagery-name', 'imagery-id']).optional(),
--- a/packages/config/src/config/base.ts
+++ b/packages/config/src/config/base.ts
@@ -41,9 +41,9 @@ export const ConfigBase = z.object({
   /**
    * Was this configuration object generated from another object
    *
-   * @default undefined / false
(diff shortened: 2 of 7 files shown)


packages/config/src/memory/memory.config.ts
```diff
@@ -143,6 +143,16 @@ export class ConfigProviderMemory extends BasemapsConfigProvider {
         const tileSet = this.imageryToTileSetByName(obj);
         allLayers.push(tileSet.layers[0]);
       }
+
+      // add any tileset aliases
+      if (ConfigId.getPrefix(obj.id) === ConfigPrefix.TileSet) {
+        const ts = obj as ConfigTileSet;
+        if (ts.alias) {
+          for (const newId of ts.alias) {
```
blacha, line 151: I think this needs to be a "virtual" tile set, I am wondering about changing virtual to be a word so this would be `virtual: "tileset-alias"` compared to `virtual: "imagery-id"` or `virtual: "imagery-name"`

We are also starting to share a lot of objects between configs, if someone did `ts.layers.pop()` it would affect all the virtual tilesets too?

packages/config/src/config/tile.set.ts
```diff
@@ -60,6 +60,9 @@ export interface ConfigTileSetBase extends ConfigBase {
 
   /** Maximum zoom level for this tileSet @default 30 */
   maxZoom?: number;
+
+  /** Array containing any tileSet aliases */
```
blacha, line 64: Where do we document why we want aliases and what a alias is used for?

packages/config-loader/src/json/json.config.ts
```diff
@@ -255,6 +255,8 @@ export class ConfigJson {
       tileSet.format = ts.type === TileSetType.Vector ? 'pbf' : 'webp';
     }
 
+    tileSet.alias = ts.alias;
```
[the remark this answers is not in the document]
ccbblin: happy to changes to `aliases` , just thought it looked slightly less nice in the config

packages/config/src/config/tile.set.ts
ccbblin: Nowhere yet :laughing: . I will add some docs and tests

packages/config/src/memory/memory.config.ts
blacha: we could maybe make the `virtual` one more complete, it is meant to be a tileset that was created from another object, eg A tile set generated from `im_:imageryId` so we dont need to create another tile set for every imagery layer created.

blacha: also `structuredClone(obj)` could be a better option for creating a new version of the tile set, the reason the other one creates a new object is because it was changing all the zoom levels, it should possibly also use `structuredClone`

packages/config/src/config/base.ts
```diff
@@ -41,9 +41,9 @@ export const ConfigBase = z.object({
   /**
    * Was this configuration object generated from another object
    *
-   * @default undefined / false
+   * @default undefined
    */
-  virtual: z.boolean().optional(),
+  virtual: z.string().optional(),
```
blacha, line 46: this would be better served as a literal type `tileset-all` `imagery-id` etc. then we can document what the virtual types are?

packages/config/src/memory/memory.config.ts
ccbblin: @blacha used `structuredClone`, not sure what "the other one" refers to though

blacha: the other instance of cloning a tile set was using `{...oldObj, newPropA:123}`

PR #3484, opened 2025-08-06
feat(geo): support tile matrixes that are not square

blacha:
### Motivation

The standard tile grids for Topo50 are not square, they are 36,000 high x 24,000 wide

### Modifications

Support rectangles tiles in tile matrixes

### Verification

unit tests.


packages/geo/src/tms/topo50.ts
```diff
@@ -0,0 +1,183 @@
...
+};
+
+export const Topo50Tms = new TileMatrixSet(Topo50tmsT);
+
+const SheetCodeRegex = /^[A-Z][A-Z][0-9][0-9]$/;
+const charA = 'A'.charCodeAt(0);
+
+function getTopLevelTile(tile: Tile): Tile {
+  if (tile.z === 0) return tile;
+  const pt = Topo50Tms.tileToSource({ x: tile.x + 0.5, y: tile.y + 0.5, z: tile.z });
```
Wentao-Kuang, line 81: Why do we need to plus 0.5 here?

blacha: using the center allows us to bypass some floating point issues eg x: 13.999999 vs 14.000001

PR #3500, opened 2025-08-27
fix: release please update to v4

ccbblin:
### Motivation

Get release please v4 working along with edited commit messages


.github/workflows/push.yml
```diff
@@ -25,11 +25,6 @@ jobs:
     environment:
       name: prod
 
-    permissions:
```
ccbblin, line 28: added this when release-please was failing, turns out needed linz-li-bot to accept basemaps request

.github/workflows/push.yml
```diff
@@ -73,15 +68,6 @@ jobs:
       - name: Check Dependencies
         run: node scripts/detect.unlinked.dep.mjs
 
-      - name: (Prod) Create tag
```
blacha, line 76: does release-please still create the `v8.6.0` tag as many things rely on that tag existing, I thought in monorepos it defaults to making specific tags per packages eg `basemaps-tiler_v8.6.0`

ccbblin: it will tag all the packages that have changed, so for https://github.com/linz/basemaps/pull/3499 I expect we will get:
`@basemaps/cli-raster-v8.7.0`
`@basemaps/cli-vector-v8.7.1`
etc

ccbblin: ah that top level tag gets used for `basemaps/cli` and `basemaps/server` container packages

release-please-config.json
```diff
@@ -25,7 +25,10 @@
     "packages/sprites": {},
     "packages/__tests__": {},
     "packages/tiler": {},
-    "packages/tiler-sharp": {}
+    "packages/tiler-sharp": {},
+    ".": {
```
ccbblin, line 29: @blacha I've added the root package so we will carry on getting v8.8.0 tags

.github/workflows/containers.yml
```diff
@@ -76,7 +76,7 @@ jobs:
               tags.push('latest');
 
               // If on a release commit add `v7`, `v7.1` and `v7.1`
-              if (${{ toJson(github.event.head_commit.message) }}.startsWith('release:')) {
+              if (${{ toJson(github.event.head_commit.message) }}.startsWith('chore: release')) {
```
blacha, line 79: would be best to keep the `release:` commit message so we have a consistent history, I think you can change it in release-please settings.

.github/workflows/push.yml
```diff
@@ -76,21 +69,6 @@ jobs:
...
-        if: github.repository == 'linz/basemaps' && github.ref == 'refs/heads/master' && startsWith(github.event.head_commit.message, 'release:')
-        run: |
-          git config user.name "github-actions[bot]"
-          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-          CURRENT_VERSION=$(node -p "require('./lerna.json').version")
-          git tag v${CURRENT_VERSION} -m v${CURRENT_VERSION}
-          git push --tags
-
-      - name: (Prod) Create github release
```
ccbblin, line 88: I've remove this as release please does this

blacha (approved):

PR #3501, opened 2025-08-31
build: track basemap's requests to LINZ's public imagery & elevation buckets

blacha:
### Motivation

LINZ tracks requests to its public buckets and if requests are not signed it can be difficult to know who had requested the data.

It would be nice to exclude basemaps from any usage stats so we need a easy way to identify basemaps from other data requests

### Modifications

sign requests to linz's open data registry buckets so basemaps usage can be excluded from reports.

### Verification

Once deployed into dev it should just "work" ?


packages/shared/src/file.system.ts
```diff
@@ -96,9 +96,6 @@ s3Fs.credentials = credentials;
 
 fsa.register('https://', new FsHttp());
 fsa.register('s3://', s3Fs);
-fsa.register('s3://nz-imagery', s3FsPublic);
```
Wentao-Kuang, line 99: Do we still need `s3FsPublic` as it is not been used anywhere.

blacha: its exported so wasnt sure if any dependent packages were using it

Wentao-Kuang: Cool, happy to keep it.

PR #3511, opened 2025-09-16
feat: add ZSTD decompression and rgbi pipelines

blacha:
### Motivation

LINZ has quite a large collection of RGBi imagery, we need a way for basemaps to render  simple visulisations for the imagery. 

### Modifications

Adds a pipeline to convert from 16 bit uint rgbi imagery into NDVI, RGBA and false color outputs

### Verification

Run locally against 16bit rgbi imagery
image

image


packages/config/src/memory/imagery.outputs.ts
```diff
@@ -12,12 +13,96 @@ export function addDefaultOutputPipelines(
   // Unable to determine band information
   if (img.bands == null || img.bands.length === 0) return undefined;
 
-  const colors = img.bands.map((m) => m.color).join(',');
+  // const colors = img.bands.map((m) => m.color).join(',');
```
Wentao-Kuang, line 16:
```suggestion
```

Wentao-Kuang (approved):

PR #3541, opened 2025-10-07
build: replace all action secrets with environments

blacha:
### Motivation

To protect secrets put them behind environments and branch protections

### Modifications

moves all secrets to environments


packages/lambda-analytic-cloudfront/src/index.ts
```diff
@@ -3,4 +3,4 @@ import { lf } from '@linzjs/lambda';
 
 import { main } from './handler.js';
 
-export const handler = lf.handler(main, { tracePercent: 0, rejectOnError: true }, LogConfig.get());
+export const handler = lf.handler(main, { tracePercent: 1, rejectOnError: true }, LogConfig.get());
```
ccbblin, line 6: assuming tracing everything will be okay

PR #3614, opened 2026-04-15
feat: swap to new addressing layer BM-1623

blacha:
### Motivation

NZ addressing pilot is deprecated and not getting updates

### Modifications

Swap to new addressing layer

### DIFF
--- a/packages/cli-vector/schema/addresses.json
+++ b/packages/cli-vector/schema/addresses.json
@@ -3,9 +3,9 @@
   "metadata": { "attributes": ["housenumber", "name", "id"] },
   "layers": [
     {
-      "id": "105689",
-      "name": "105689-nz-addresses-pilot",
-      "source": "s3://linz-lds-cache/105689/",
+      "id": "123113",
+      "name": "123113-nz-addresses",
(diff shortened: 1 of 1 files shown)


packages/cli-vector/schema/addresses.json
```diff
@@ -3,9 +3,9 @@
...
   "layers": [
     {
-      "id": "105689",
-      "name": "105689-nz-addresses-pilot",
-      "source": "s3://linz-lds-cache/105689/",
+      "id": "123113",
+      "name": "123113-nz-addresses",
+      "source": "s3://linz-lds-cache/123113/",
       "tags": {},
       "attributes": { "address_id": "id", "full_address_number": "housenumber" },
```
Wentao-Kuang, line 10: Just want to check if the new address data have the same property of `address_id` and `full_address_number`?

blacha: As far as I know the schema hasnt really changed at all https://data.linz.govt.nz/layer/123113-nz-addresses/ both those fields exist

image

Wentao-Kuang (approved):

